PR #6, opened 2017-11-07
Add support to show current user name in title.

pmisik:
This can be useful if you run FarManager as different user.


alabuzhev:
Thanks.
However, introducing a new variable might be not the best possible approach.
We can simply expand environment variables there, e.g. `strTitleAddons += os::env::expand(Global->Opt->strTitleAddons);`
After that you will be able to put `%username%` there or anything else you want.

pmisik:
@alabuzhev 
You are right. Your solution is clean and more generic. 
So I'm closing this PR in in favor of #7

PR #24, opened 2018-02-05
Translated two help pages to English in FarEng.hlf.m4.

MKadaner:
1. Translated @FindFileAdvanced and @PanelSettings help pages to English in FarEng.hlf.m4.
2. Added missing option descriptions to Russian and English help.

Also, some help and string resources cleanup.


far/FarEng.hlf.m4
```diff
@@ -2506,56 +2505,54 @@ $ #Settings dialog: panel#
...
 
-  #Удваивать глобальные#    Вместо одинарной будет рисоваться 
-  #разделители колонок#     двойная внутренняя рамка между колонками.
+  #Highlight column#        If this option is on, internal column separators
+  #separators#              are displayed in the list item's color.
+                          Otherwise, the color of the outer frames is used.
 
+  #Double global#           If this option is on, the double internal frame
+  #column separators#       between columns is displayed instead of the single
+                          frame.
```
MKadaner: What does this option do? I did not see any changes when I toggled this option in the build 5000.

far/FarEng.hlf.m4
```diff
@@ -2506,56 +2505,54 @@ $ #Settings dialog: panel#
...
 
 @TreeSettings
 $ #Tree settings#
```
MKadaner: Do "Tree settings" options menu item and the corresponding options dialog exist? I do not see them in the build 5000.

far/FarEng.hlf.m4
alabuzhev: It applies to "primary" columns only (i.e. N).
If you set your panel mode to something like `N,S,N,S,N,S` you will see the difference:

diff

It was introduced to make it easier to recognise global primary (`N`) and non-primary columns (e.g. `S`) when `Highlight column separators` is not enabled.

The image above also demostrates what does `Highlight column separators` do.

far/FarEng.hlf.m4
w17: far:config
select Panel.Tree.TurnOffCompletely
Shift+F1

far/FarEng.hlf.m4
MKadaner: Thank you. I'll update description of `Double global column separators` in the next PR.

far/FarEng.hlf.m4
MKadaner: Ah-a! `По умолчанию значение = True`. This explains why I've never seen it. And I like it this way. Thank you.

I will need to translate this help topic and probably add topics for other `Panel.Tree` options.

PR #25, opened 2018-02-12
Refactored calculation of stripes (former global columns) on file panels.

MKadaner:
I wanted to write better description of #Double global column separators# in @PanelSettings dialog (both Russian and English). While I tried to make sense of "global columns" calculations in FileList::PrepareColumnWidths, I slightly refactored the algorithm (and made it marginally faster).

If you like the change, please merge. I am not sure I properly followed the style of the existing code; I'll be glad to address the comments.

If you do not like the change, or feel it is gratuitous, just kill the pull request; I'll be fine with that too,


MKadaner:
Thank you!

PR #26, opened 2018-02-12
Description of columns and stripes on @PanelViewModes and @PanelSettings help pages.

MKadaner:
Do you like the word "stripes" ("полосы") instead of "global columns"? I am not happy with this but could not think of anything better. Suggestions are welcome.

By the way, I tried "Far.exe -set:DefaultLanguage=Russian", and it did not work, meaning Far started with English resources. Language setting does work in Far.exe.ini. Is it a bug, or did I do something wrong?


far/farlang.templ.m4
```diff
@@ -2251,12 +2251,12 @@ upd:"Highlight column separators"
 "Підсвічувати роздільники колонок"
 
 MConfigDoubleGlobalColumnSeparator
```
MKadaner: Should I rename MConfigDoubleGlobalColumnSeparator to MConfigDoubleStripeSeparator? There is also DoubleGlobalColumnSeparator in config.cpp. Renaming of the latter is probably not advisable, because it will be a breaking change to config DB, right?

far/farlang.templ.m4
```diff
@@ -2251,12 +2251,12 @@ upd:"Highlight column separators"
...
-"Удваивать глобальные разделители колонок"
-"Double global column separators"
+"Удваивать разделители полос"
+"Double stripe separators"
 "Dvojité oddělovače hlavních sloupců"
```
MKadaner: Should I reset other languages to "upd:Double stripe separators"? More general, does anybody translate to other languages?

alabuzhev: It depends.
If the meaning is changed completely, it is better to reset to English and prepend with "upd:".
If it is more about style or wording, i.e. the outdated string still makes sense, you can leave as is and just prepend with "upd:".

Translations are provided by volunteers. Some are updated more or less regularly, some are not.

far/farlang.templ.m4
alabuzhev: It is ok to rename language ids accordingly.
Renaming db strings would be a breaking change, so yes, it is better to leave them as is.

alabuzhev:
> I tried "Far.exe -set:DefaultLanguage=Russian", and it did not work, meaning Far started with English resources.

`Far.exe -set:whatever` overrides `far:config` values.
In far:config there is no `DefaultLanguage` key, but there are `Language.Main` and `Language.Help`.

far/FarEng.hlf.m4
```diff
@@ -3920,8 +3921,20 @@ the list. In this list "Brief mode" item corresponds to brief panel mode
 on. The last item, "Alternative full", corresponds to view mode called with
 #LeftCtrl+0#. After selecting the mode, you can change the following settings:
 
-  - #Column types# - column types are encoded as one or several
-characters, delimited with commas. Allowed column types are:
+  - #Column types# - column types are encoded as comma-delimited list
+of column specifiers. Each column specifier starts with
+a single-character denoting a file attribute. Some file attribute
```
alabuzhev: `file attribute` is not the best term here (and below) - it can be confused with "real" file attributes (Read Only, Hidden etc.).

far/FarRus.hlf.m4
```diff
@@ -4043,9 +4044,19 @@ $ #Настройка режимов просмотра панели файло
 (#ЛевыйCtrl+2#) и так далее. После выбора режима появляется диалог, в котором
 можно изменить следующие параметры:
 
-  - #Типы колонок# - типы колонок кодируются с помощью одной
-или нескольких букв, разделённых запятыми. Допускаются следующие
-типы колонок:
+  - #Типы колонок# - типы колонок кодируются списком спецификаторов
+колонок, разделённых запятыми. Каждый спецификатор начинается с буквы,
+обозначающей атрибут файла. Некоторые атрибуты могут иметь
```
alabuzhev: Same as in the English help - `атрибут файла` already has a different meaning.

far/FarRus.hlf.m4
```diff
@@ -4043,9 +4044,19 @@ $ #Настройка режимов просмотра панели файло
...
+квалификаторы. Допустимые атрибуты и квалификаторы перечислены ниже.
+
+    Если список типов колонок состоит из нескольких повторяющихся групп,
+файлы на панели организованы в «полосы». Атрибуты каждого файла показаны
+в колонках одной полосы, а список файлов продолжается с полосы
+на полосу, как текст в газетной статье. Если список типов колонок
+невозможно разбить на одинаковые группы, список файлов показан в одной
```
alabuzhev: `список файлов будет показан` maybe?

far/FarEng.hlf.m4
```diff
@@ -3920,8 +3921,20 @@ the list. In this list "Brief mode" item corresponds to brief panel mode
...
+characters can be followed by an attribute qualifier characters.
+Supported column type characters and their modifiers are listed below.
+
+    If the list of column types consists of two or more repeated groups,
+the files on the panel are listed in “stripes.” Attributes of each file
+are displayed in columns of a stripe, and the list of files continues
+from one stripe to the next like the text laid out in a multi-column
+paper article. If column type list cannot be properly split into the
+equal groups, the files are listed in a single stripe.
```
alabuzhev: `will be listed` maybe?

MKadaner:
Thank you for the comments. I reworked the descriptions of columns and stripes.

Got it about `Far.exe -set:whatever`. I mixed up two adjacent lines of command-line help. :(

PR #29, opened 2018-02-22
Fixed VC15 64-bit compilation error C3487.

MKadaner:
In 64-bit mode, VC15 (2017) compiler complained about a lambda where it could not deduce return type:
```
string_utils.cpp(319,1): error C3487: 'int': all return expressions must deduce to the same type: previously it was '__int64'
                        return EndOrNonDigit1? -1 : 1;
string_utils.cpp(322,1): error C3487: 'int': all return expressions must deduce to the same type: previously it was '__int64'
                return Comparer({ It1++, 1 }, { It2++, 1 });
```
This happened because `ZerosResult` returned above was defined `auto` and deduced from `ptrdiff_t`.

Fixed by `static_cast<int>`ing the initializer of `ZerosResult`.

### DIFF
--- a/far/changelog
+++ b/far/changelog
@@ -1,3 +1,7 @@
+MZK 21.02.2018 22:37:57 -0800 - build 5149
+
+1. Пофиксил ошибку компиляции в VC15 64-бит: C3487: 'int': all return expressions must deduce to the same type: previously it was '__int64'.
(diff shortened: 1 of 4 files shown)


alabuzhev:
Thanks!

PR #34, opened 2018-03-19
Rewrote @PreserveStyle help page (Both English and Russian).

MKadaner:
Finally! This one was tough, probably because the feature is frankly obscure and I do not like much.


alabuzhev:
Thanks!

MKadaner:
Thank you!

PR #36, opened 2018-04-02
Fixed buffer overflow in SearchString.

MKadaner:
While trying to understand what CtrlEnter does in Find-All menu, I hit buffer overflow (`assert` in `range::operator[]`). This change fixes the issue; plus some refactoring in `SearchString` around regex search. Also, I noted another issue with regex search.

The search algorithm is suboptimal, especially since it is used to "find all" and "find in files" (if I am not mistaken). From long ago, I have an implementation of [Bitap algorithm](https://en.wikipedia.org/wiki/Bitap_algorithm), which should be marginally more efficient than that published on Wikipedia. I have an idea of how to adapt it to `wstring`'s without building of the half-megabyte table. It also should nicely cover the case-insensitive option. I'll try to brush it up and apply here, but it's unlikely to happen very soon.


far/editor.cpp
```diff
@@ -3800,6 +3800,10 @@ bool Editor::Search(bool Next)
 					break;
 				case KEY_CTRLENTER:
 				case KEY_RCTRLENTER:
+					// MZK 2018-04-01:	After the Find-All dialog is hidden and the cursor is positioned to the found string,
+					//					the next pressed key is still handled by the Find-All dialog, which becomes visible again.
+					// Improvement:		This behavior is probably expected, but some kind of visual indication would be very welcome.
+					// BUG?				Also, navigation keys (arrows, PgUp/PgDn, etc.) show the dialog without the shadow.
```
MKadaner: I am going to describe current behavior (modulo minor glitches) on the `@FindAllMenu` help page. Main question is whether Find-All menu should disappear on CtrlEnter? After all, it's a modal dialog. Currently it disappears.

far/strmix.cpp
```diff
@@ -911,71 +993,33 @@ bool SearchString(string_view const Source, const string& Str, const string& Upp
 	if (!Regexp && PreserveStyle && PreserveStyleReplaceString(Source, Str, ReplaceStr, CurPos, Case, WholeWords, WordDiv, Reverse, *SearchLength))
 		return true;
 
+	const int SourceSize = static_cast<int>(Source.size());
+
 	if (Reverse)
 	{
+		// MZK 2018-04-01 BUGBUG: regex reverse search: "^$" does not match empty string
```
MKadaner: I will take care of this in a follow up PR.

far/strmix.cpp
```diff
@@ -911,71 +993,33 @@ bool SearchString(string_view const Source, const string& Str, const string& Upp
...
 
-		if (Position == static_cast<int>(Source.size()))
+		if (Position == SourceSize)
 			return false;
 
-		const int Length = *SearchLength = (int)Str.size();
+		const int Length = *SearchLength = static_cast<int>(Str.size());
 
-		for (int I=Position; (Reverse && I>=0) || (!Reverse && I < static_cast<int>(Source.size())); Reverse ? I--:I++)
+		for (int I = std::min(Position, SourceSize - Length); 0 <= I && I <= SourceSize - Length; Reverse ? I-- : I++)
```
MKadaner: This was the root cause of the buffer overflow. If `I` was close to the end of the `Source`, the original code could continue matching characters of `Str` beyond the `Source`'s end.

far/strmix.cpp
```diff
@@ -911,71 +993,33 @@ bool SearchString(string_view const Source, const string& Str, const string& Upp
...
-			}
-			if (found)
-			{
-				*SearchLength = pm[half].end - pm[half].start;
-				CurPos = pm[half].start;
-				ReplaceStr = ReplaceBrackets(Source, ReplaceStr, pm + half, n, hm);
-			}
-
-			return found;
+			return SearchStringRegex(Source, re, pm, hm, Position, Reverse, ReplaceStr, CurPos, SearchLength);
```
MKadaner: The refactoring was mostly extracting of the `SearchStringRegex` helper function (plus some shuffling of things around). Unfortunately, the diff is not great.

far/editor.cpp
alabuzhev: It should not.
Ctrl-Enter changes the position without closing the dialog.
Please do not describe this in help.

The current behaviour is probably caused by some of the recent changes in window manager.
@zg0 , could you have a look please?

far/changelog_eng
```diff
@@ -3,6 +3,10 @@ This file is a translation of the main russian changelog and is provided by volu
 It might not always be as up to date as the main changelog.
 =======================================================================================
 
+MZK 01.04.2018 21:23:21 -0700 - build 5178
+
+1. Починил переполнение буфера в SearchString + рефакторинг.
```
alabuzhev: Looks like changelog entries are mixed up.

MKadaner: Thank you. Fixed.

far/editor.cpp
alabuzhev: This issue has been fixed, please drop the comment.

alabuzhev:
> The search algorithm is suboptimal, especially since it is used to "find all" and "find in files"

"Find in files" uses Boyer-Moore-Horspool algorithm, perhaps it can be generalised and reused in Editor.

far/editor.cpp
MKadaner: Thank you. And thank you, @zg0 for the very quick fix. I reverted the comment. The behavior is simple and logical now. I'll describe it with my pleasure.

alabuzhev:
Thanks, merged.

far/strmix.cpp
alabuzhev: @MKadaner, could you please explain how exactly the initial issue can be reproduced?
It looks like this change [breaks](https://forum.farmanager.com/viewtopic.php?p=149038#p149038) the search logic, we need to fix it in a different way.

MKadaner: Thank you for reverting. Sorry for not checking repeated search.

Repro: *debug* build. File contains
```
ABCD
```
Searching for `DEF` from the beginning will hit `assert` in `range::operator[]`.

MKadaner: @alabuzhev, I looked at your change. The early return you added
```
	if (Haystack.empty())
		return false;
```
breaks the regex search. Now regex /^$/ does not match empty string .

I was going to fix this same problem (with reverse search only) caused by another earlier return, which was there for a long time, so you can leave this one to me as well. However, I won't be able to do this earlier than the coming weekend, and even then not sure I'll have enough time.

alabuzhev: Thanks, fixed. I'll leave reverse search case to you.

PR #38, opened 2018-04-08
Fixed LeftCtrl+0..9 in the Find-all menu. The menu used to disapper.

MKadaner:
Original repro:

1. Editor -> Set a bookmark (e.g., RightCtrl+0).
2. Search -> Find all.
3. Go to the bookmark (LeftCtrl+0).
4. Find-all dialog disappears.
5. Press any key (not Esc or Enter).
6. Find-all dialog reappears.

Please look at the change. I am not sure I did not break something, especially since I touched `ProcessKeyInternal` function.


alabuzhev:
Thanks

PR #39, opened 2018-04-09
Reworked @FindAllMenu and @FileOpenCreate help pages.

MKadaner:
Also, added handling of Ctrl+Left Mouse Click to the Find-all menu. It feels like a natural counterpart of Ctrl+Enter.


far/FarEng.hlf.m4
```diff
@@ -3225,58 +3225,52 @@ string is changed to that of the found string. For example:
...
+    #Ctrl+Up#         - Scroll the text in the editor.
+    #Ctrl+Down#
 
-      #Ctrl+Enter#           - перейти к позиции найденного текста
+    #Ctrl+Enter#      - Go to the position of the found text.
+    #Ctrl+Left#
+      #mouse click#
 
-      #Add#                  - добавить стековую закладку с текущей позицией
+    #Gray +#          - ^<wrap>Add stack bookmark with the current position.
```
MKadaner: What are "stack bookmarks"? There seems to be full-fledged backend functionality in the editor, but I could not find, either in the code or in the help, any UI to access it except for this place in the Find-all menu.

far/changelog
```diff
@@ -1,3 +1,9 @@
+﻿MZK 08.04.2018 19:03:09 -0700 - build 5187
```
MKadaner: I added BOM here. Is it OK?

far/FarEng.hlf.m4
alabuzhev: 'stack bookmarks' is an outdated name. See [session bookmarks](https://api.farmanager.com/ru/service_functions/editorcontrol.html#ECTL_ADDSESSIONBOOKMARK).

far/FarEng.hlf.m4
```diff
@@ -3225,58 +3225,52 @@ string is changed to that of the found string. For example:
...
+    The #Shift+F4# key combination opens an existing file or creates
+a new file.
 
-    1) Если файл открывался ранее, то возможно для него уже сохранена информация о
-кодовой странице (зависит от состояния параметра "[x] Сохранять позицию файла"
-в ~настройках редактора~@EditorSettings@).
-В этом случае используется запомненная ранее кодовая страница.
+    If the specified file does not exist, a new file will be created.
+The code page for the new file is specified in the #Code page# list.
+By #default#, the ANSI code page is used. The default code page can be
```
alabuzhev: Maybe skip this sentence, given that the default code page can be something else than ANSI?

far/editor.cpp
```diff
@@ -3806,6 +3806,8 @@ bool Editor::Search(bool Next)
 
 				case KEY_CTRLENTER:
 				case KEY_RCTRLENTER:
+				case KEY_CTRL|KEY_MSLCLICK:
+				case KEY_RCTRL|KEY_MSLCLICK:
```
alabuzhev: Would it always work as expected - say, if you click the border or outside the menu?

far/FarEng.hlf.m4
MKadaner: It's all about the API. However, there seems to be no built-in UI in FAR other than this single place in the Find-all dialog. Feels strange and asymmetrical. It seems without plugins or macros, one can set a session bookmark, but cannot use it in any way. I would humbly suggest removing this hotkey here as well.

far/changelog
MKadaner: Makes sense. Will remove in the next iteration.

far/editor.cpp
MKadaner: Good question. Somewhat surprisingly, the behavior is reasonable.

Gesture | Baseline | My Change
--- | --- | ---
Ctrl+Click outside | Closes dialog | Closes dialog
Ctrl+Click on border frame | Does nothing | Jumps to the currently selected item without changing the item *)
Ctrl+Click on border margin (outside of the frame) | Moves dialog | Jumps to the currently selected item without changing the item *)
Ctrl+Drag from border outside | Closes dialog | Closes dialog

*) Admittedly, this behavior is somewhat unexpected. Will it do, or should I try to filter out Ctrl+Click on the border?

far/FarEng.hlf.m4
MKadaner: Good point. I'll reword. To be sure: default code page is defined in the global editor setting dialog; no additional logic is applied, right?

@zg0 Why? Could you please be more specific?

far/FarEng.hlf.m4
alabuzhev: Yes, session bookmarks initially were intentionally introduced without any UI.
This feature was added [later](https://bugs.farmanager.com/view.php?id=2096), not sure why.

far/editor.cpp
alabuzhev: Thanks for checking.
I'm asking because mouse events are usually handled via ProcessMouse(), using KEY_MSLCLICK is quite rare.
Perhaps it's ok in this case.

far/FarEng.hlf.m4
MKadaner: Thank you. Strange.

far/editor.cpp
MKadaner: Thank you for the explanations. I'll think about it. I do not quite like the behavior on the border.

alabuzhev: Let's proceed with KEY_MSLCLICK, it's good enough for now.

MKadaner:
I did not squashed/rebased to expedite reviewing the changes. If you are OK with this version, I'll prepare the final commit.

MKadaner:
Done! Thank you.

MKadaner:
Merge please.

MKadaner:
@alabuzhev, could you please merge this PR?

alabuzhev:
Merged, thanks.

MKadaner:
Thank you.

PR #49, opened 2018-06-19
Delete changelog

techtonik:
git log gives more complete history
https://github.com/FarGroup/FarManager/commits/master/enc


alabuzhev:
There's no harm in having history in two places.

nightroman:
log != git history
Log ~ "release notes" is for potential users with some notes and instructions.
Git history is just some dev notes on commits.

techtonik:
changelog format is too similar to Git. and the info there is outdated, so I decided to clean it up

PR #52, opened 2018-06-20
nightly/enc.sh: Build enc in repository root _output/enc

techtonik:
nighly build files for Encyclopedia will be available at:

    _output/outfinalnew32/Encyclopedia
    _output/outfinalnew64/Encyclopedia

And the build itself is taking place in:

    _output/enc  - which is a 1:1 copy from the repo

### DIFF
--- a/misc/nightly/enc.sh
+++ b/misc/nightly/enc.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# absolute path to this script
+DIR=$(dirname $(readlink -f $0))
+# absolute path to source root
+ROOT=$(cd $DIR/../.. && pwd)
+BUILDDIR=$ROOT/_output
+
+
 function benc2 {
 LNG=$1
 L=$2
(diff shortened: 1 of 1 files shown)


techtonik:
If this layout is okay I will update the rest of `nightly` scripts to follow the same scheme.

trexinc:
What is the purpose of those changes?

trexinc:
And in any case this breaks Far build.

techtonik:
1. This is needed to understand where files produced by nightly build go. It also makes it possible to run build scripts from any place.
2. Where are logs from nightly builds? It could be tested by Travis.

techtonik:
So this is a way to document this certain dir structure in code. I need to get .hlf files from source on Linux and upload them to https://fardocs.netlify.com/ for experiments. Started with `enc.sh`, because I know its layout.

techtonik:
@trexinc what is the dir structure that scripts expect? Is it possible to figure out what changes are made on disk without going through the source code?

PR #56, opened 2018-06-30
at 30.06.2018 08:09:00 -0800 - build 5235

andrastantos:
Hi!

First of all, I'm new to this, so please tell me if I'm doing something wrong...

I'm requesting a pull to incorporate a few new key-mappings to FAR to make it more usable on small laptops without numeric keypads:

1. Added new key-mappings to support laptops with no numeric keypad
   Added new 'Don't use numeric keypad' option (default off) to toggle some conflicting or changed behavior
   Changed (English) help to describe changes

Thanks,
Andras Tantos

### DIFF
--- a/far/FarEng.hlf.m4
+++ b/far/FarEng.hlf.m4
@@ -339,25 +339,36 @@ $ #Panel control commands#
   Restore default panels width                          #Ctrl+Numpad5#
   Restore default panels height                     #Ctrl+Alt+Numpad5#
   Show/Hide functional key bar at the bottom line.            #Ctrl+B#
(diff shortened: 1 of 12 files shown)


randolf:
This seems to be a great idea.  It should also be fairly easy to test even on laptops with numeric keypads by simply not using the numeric keypad.

alabuzhev:
The conventional way to redefine keyboard shortcuts is to use [macros](https://github.com/FarGroup/FarManager/tree/master/addons/Macros).

trexinc:
So maybe create a macro script that can be included with Far to assist all laptop users?

PR #67, opened 2018-09-03
Translated help page @ChangeDriveMode into English and edited Russian page

MKadaner:


far/FarRus.hlf.m4
```diff
@@ -3575,62 +3577,64 @@ CD-RW, CD-RW/DVD, DVD-ROM, DVD-RW и DVD-RAM. Функция определен
...
-    Показать общее и свободное место на диске.
-    Комбинация клавиш в меню: Ctrl+5
+    Показать размер диска и количество свободного места на нём.
+    Комбинация клавиш в меню: #Ctrl+5#.
 
     #Показывать размер в виде десятичной дроби#
-    Показать общее и свободное место на диске в виде десятичной дроби.
-    Комбинация клавиш в меню: Ctrl+5 (повторное нажатие).
+    Показать размер диска и количество свободного места в виде
+десятичной дроби.
```
MKadaner: I am not sure this describes the actual behavior. It looks like Far switches between two measurement units (M vs. G or G vs. T, etc.) Should I describe the actual behavior (and adjust `Change Drive Menu Options` dialog accordingly) or would you suggest modifying how `Change drive` menu shows sizes?

alabuzhev: If checked it's rational (1.04 G), otherwise integer (1,061 M).
The behaviour is the same as in file panels and probably already described there.

alabuzhev:
Thanks.

far/FarRus.hlf.m4
MKadaner: Do you mean the `C` modifier to file size column type (`S`)? Then, the description is not very helpful:
```
C - format file size;
```

And the behavior seems to be somewhat different.

alabuzhev: Not helpful indeed. Would you like to improve it? :)

MKadaner: Yeah! I'll add some explanations with the next PR. I'll also try to come up with a better wording for `Ctrl+5` in `Change drive` menu.

PR #69, opened 2018-09-17
Help: clarified description of file size formatting options.

MKadaner:


far/FarEng.hlf.m4
```diff
@@ -4024,31 +4027,35 @@ files will be listed on a single stripe.
...
-                            (i.e. 0.97K);
+                 where: C - ^<wrap>group digits using the symbol from
+Windows settings;
+                        T - ^<wrap>use decimal units instead of binary,
+i.e., to calculate kilobytes, the size will be divided by 1000 instead
+of by 1024; in this mode unit symbol is shown in lower case, e.g., #k#,
+#m#, #g# instead of #K#, #M#, #G#;
+                        F - ^<wrap>show size like Windows Explorer, as
+a decimal fraction with no more than three digits before decimal point,
+e.g., 999 bytes will be shown as #999#, while 1000 bytes as #1.00 k#;
```
MKadaner: This is actual behavior; probably not what was intended. I could not get `0.97 K` from the formatting algorithm though. I can fix the implementation if you like, but I would throw away all the math there and use a table of hard-coded thresholds to choose units.

alabuzhev: - Since "T" is not on by default, let's change this to "while 1024 bytes as 1.00 K"
- "show size like Windows Explorer" is useless in this particular example- these days Explorer shows 0 bytes as 0 K and 1 byte as 1 K.

PR #70, opened 2018-09-24
Translated help page @SelectFiles into English and reworked the Russian page.

MKadaner:


far/FarEng.hlf.m4
```diff
@@ -4225,54 +4225,53 @@ and the '|' character separates include masks from exclude masks.
...
-    Если щелкнуть правой кнопкой мыши над списком файлов (строка
-с заголовками колонок) или под списком файлов (статусная область) то при
-удерживании правой кнопкой мыши список файлов будет прокручиваться вверх
-или вниз помечая или снимая пометку с файлов, по которым пробегает курсор.
-Действие (пометка или снятие пометки) соответствует состоянию файла под
-курсором до начала прокрутки.
+    #Right click# toggles selection on the clicked item and moves the
+cursor to it.
 
+    #Right click and hold# on the #status line# or #column titles# moves
```
MKadaner: 1. It looks like this feature heavily glitches. It seems to take into account the state of the next item, not the item under cursor. Also, the behavior somehow depends on the current state of the selection, if not on the history of the cursor movement. I could not formulate the exact rules.
2. The cursor moves way too fast, even with debug build. It is difficult or impossible to use this feature meaningfully.

far/FarEng.hlf.m4
```diff
@@ -4225,54 +4225,53 @@ and the '|' character separates include masks from exclude masks.
...
+  #Mouse Selection#
 
-    Если щелкнуть правой кнопкой мыши над списком файлов (строка
-с заголовками колонок) или под списком файлов (статусная область) то при
-удерживании правой кнопкой мыши список файлов будет прокручиваться вверх
-или вниз помечая или снимая пометку с файлов, по которым пробегает курсор.
-Действие (пометка или снятие пометки) соответствует состоянию файла под
-курсором до начала прокрутки.
+    #Right click# toggles selection on the clicked item and moves the
+cursor to it.
```
MKadaner: `Right click selects files` option in the `Panel settings` does not affect the `Right click` behavior. As per `Settings dialog: panel` help page, if this option is off, right click should open Windows Explorer Context menu. Should I mention the setting here and fix the behavior, or update the `Settings dialog: panel` help page?

alabuzhev: What exactly do you mean?
It opens a context menu if the option is off and EMenu plugin is loaded.

far/FarEng.hlf.m4
alabuzhev: 1 - what do you mean by "take into account"? It just scrolls.
2 - yes, we should probably add some sleep() there to emulate keyboard speed (same for editor/viewer).

MKadaner: 2. Is `sleep()` appropriate? It is blocking. The "right" solution would be to schedule a timer event and exit the mouse handler. Is such technique used elsewhere in Far? I fully understand that it's much more involved. So, if `sleep()` is OK, especially if it is used in other places, I'll try to apply it here.

alabuzhev: It is ok, used in other places and used to be here too I think.
Probably somewhere around or in `IsMouseButtonPressed()` function.

alabuzhev:
Thanks, merged

far/FarEng.hlf.m4
```diff
@@ -4225,54 +4225,54 @@ and the '|' character separates include masks from exclude masks.
...
+items along the way. The action (selection or deselection) depends
+on the state of the item under cursor before pressing the key
+combination.
 
-    #Ctrl+<Gray +># и #Ctrl+<Gray -># выбирают и снимают пометку со всех
-файлов с тем же расширением, что и у файла под курсором. Если параметр
-#Пометка папок# в диалоге ~Настроек панели~@PanelSettings@ включен, то
-пометка меняется и для папок.
+    #Gray +# selects, and #Gray -# deselects the items using one or more
+~File masks~@FileMasks@. #†#
```
alabuzhev: @MKadaner , what exactly these `†` for? (just noticed)

MKadaner: I meant a footnote which applies to several paragraphs. The footnote text is right below the `Shift+<Gray +>` paragraph. I tried star, like this `*)`, but it did not look right. The dagger symbol looks better. I did not like the way it was done in the original version and wanted to avoid the repetition.

If you can suggest better approach, I'll be happy to fix it.

alabuzhev: Ah, footnote, right.
It's just a little confusing because five sequential paragraphs refer to the same footnote and the footnote is not actually in the page footer :)

PR #75, opened 2018-10-29
Made fuction `while_mouse_button_pressed` a template.

MKadaner:
See if it makes sense to merge. The optimization is certainly marginal. The parameter to `while_mouse_button_pressed` is always a lambda. Passing it as a template parameter instead of `std::function` should slightly improve performance by avoiding virtual function call.


alabuzhev:
Templates are definitely faster.
The question is [how much faster](http://coliru.stacked-crooked.com/a/006ab7dc103c540a) and is it worth polluting the header with datetime and winapi stuff.

MKadaner:
Well, I was probably carried away with perf optimization. I currently work on a project where we do count microseconds because they are added up when repeated 10^12 times. On the other hand, here, when the main loop actively waits, this petty improvement is simply ridiculous.

As for polluting the headers, it may not be so bad, but would require a major refactoring in the direction of having virtually all the code in the headers. This approach has its own benefits, but again likely not worth it in Far project.

I am abandoning the pull request and retiring to my main occupation: help translation. :)

PR #79, opened 2018-12-09
add catalog_user.xml in msi

ctapmex:
В новой версии схем colorer появился обязательный файл catalog_user.xml.  Добавляем его в инсталятор.

### DIFF
--- a/misc/msi-installer/features.wxs
+++ b/misc/msi-installer/features.wxs
@@ -424,6 +424,7 @@
               <ComponentRef Id="catalog_console.xml" />
               <ComponentRef Id="catalog_rgb.xml" />
               <ComponentRef Id="catalog_text.xml" />
+              <ComponentRef Id="catalog_user.xml" />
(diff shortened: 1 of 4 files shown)


yegor-mialyk:
Hi guys,

I do not think that adding **catalog_user.xml** in to the installer was the right approach. This config was created to separate user settings from global settings in order not to be overwritten during installation.

Now it is not the case and you lose your settings after the update.

Thanks,
Yegor

ctapmex:
https://github.com/colorer/Colorer-schemes/issues/70

AnrDaemon:
Then supply a very specific instructions on how to setup the `User's file of color styles…`.
I've had to, basically, use a trial and error approach to get it working.

ctapmex:
I started describing the work with the catalog [here](https://github.com/colorer/Colorer-library/tree/master/docs), I will add it to the plugin.

PR #81, opened 2018-12-18
Added Polish translation into installer.

BestiaPL:
New file - polish translation.


alabuzhev:
I'll merge this PR because having that file there won't break anything.
However, someone will need to update makefile to make it work.

PR #103, opened 2019-05-20
Partially translated into Polish

BestiaPL:


alabuzhev:
This won't work.
Arclite generates lng files from msg files.
Please have a look at [plugins/arclite/en.msg](https://github.com/FarGroup/FarManager/blob/master/plugins/arclite/en.msg).

pl.msg should be in the same format.

alabuzhev:
Ok, I've created pl.msg myself.
Please translate that file in future.

PR #104, opened 2019-05-20
Changed to UTF

BestiaPL:


alabuzhev:
Merged, but could you please put multiple file onto one PR in future please? Thanks.

PR #108, opened 2019-05-20
Started translation help file into Polish

BestiaPL:


alabuzhev:
I see you've started translating FarPol.hlf.**m4** in #119 (which is the proper way to do that), shall we close this one?

PR #157, opened 2019-07-05
Allow reusing the original title via %Default in -title

lunakid:
Hello there,

As discussed at https://forum.farmanager.com/viewtopic.php?f=54&t=11685, this change allows 
reusing the default "smart" FAR title in user-customized titles (e.g. when I launch FAR as a shell for
a dev. env., now I can do `far "-title:VC2019 - %default"` to both have a clear indication of which FAR
window that is, and also enjoy the usual dynamic title updates of FAR).

*Notes:*

* I've reused the technique seen in `FileEditor::ShowConsoleTitle()`, that's where the %... syntax comes from. This way I could avoid inventing or introducing anything new for this small feature.

* I originally thought of using `%Title` for the placeholder, but that was not as clear as I'd like it. Another advantage of `%Default` is that it could then be also reused for future cmd-line args (or anywhere else, for that matter) the same way, if needed.

Cheers,
Sz.


alabuzhev:
Thanks, merged.
Would you like to be mentioned in [CONTRIBUTORS](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTORS.md) file? If so, how?

lunakid:
Great, thank you!
You could add me as Szabolcs Szász.
(I'd like to contribute more in the future, and adding me now might help/prod me a bit to actually do so. :) )
Thx, cheers!

PR #162, opened 2019-08-12
Translated several far:config help pages into English.

MKadaner:


far/FarEng.hlf.m4
```diff
@@ -5485,67 +5487,69 @@ list.
...
 $ #far:config Help.ActivateURL#
- Параметр позволяет управлять активацией URL ссылок в HLF-файлах:
+ This parameter controls activation of URL links in HLF-files:
 
- 0 - отключить активацию.
- 1 - активация включена.
- 2 - активация включена, но выдавать предупреждающее сообщение.
+ 0 – Activation is disabled.
+ 1 - Activation is enabled.
+ 2 - ^<wrap>Activation is enabled; warning dialog is displayed.
```
alabuzhev: - 'Activation' sounds weird in this context. Maybe 'opening' (and 'активация' - 'открытие')?

- It's better to put <^wrap> in the 1st line, even if it's needed only in the 3rd (same for the other hlf).
- The first line uses an en dash (`–`), while other use a hyphen (`-`). Can it be consistent please?

far/FarRus.hlf.m4
```diff
@@ -5573,65 +5575,69 @@ $ #far:config Codepages.NoAutoDetectCP#
...
 
  Номера битов:
- 0 - Панели.
- 1 - Редактор.
- 2 - Внутренняя программа просмотра.
- 3 - Окно подсказки.
- 4 - Диалоги.
+ 0 – Панели;
+ 1 – Редактор;
+ 2 - Программа просмотра;
```
alabuzhev: Ditto, can it be consistent please?

MKadaner: Argh! Replaced all en dashes with hyphens. I did not mean it. MS Word autocorrect. :( I'll be more careful in the future.

far/FarEng.hlf.m4
MKadaner: I think "activation" came from some formal specs around Windows. To me it also sounds strange, always has. On the other hand, given that parameter is called `Help.ActivateURL`, I think we should keep this word somewhere. I reworded both languages. Please see if you like them better.

Addressed the other two bullets.

MKadaner:
> Could you please copy-paste updated parts of FarEng.hlf.m4 into FarHun.hlf.m4 and FarPol.hlf.m4?
Done.

Should I do it always?

alabuzhev:
> Done. Should I do it always?

Thanks. Yes please, if it's not too difficult.

PR #213, opened 2019-11-21
Hdropstr

kyky:
CTRLC - список фалов HDROP дублируется в CF_UNICODE

### DIFF
--- a/far/clipboard.cpp
+++ b/far/clipboard.cpp
@@ -371,6 +371,19 @@ bool clipboard::SetHDROP(const string_view NamesData, const bool bMoved)
 	if (!Clear() || !SetData(CF_HDROP, std::move(Memory)))
 		return false;
 
+	 string TextData;
+	 const auto Eol = eol::str(eol::system());
+	 for (const wchar_t & c : NamesData)
+		{
+		 if (c!=L'\0') TextData+=c; else TextData+=Eol;
+                 //TextData+=((c!=L'\0')?c:Eol);  // hate c++
+		}
+         auto hData = os::memory::global::copy(string_view(TextData));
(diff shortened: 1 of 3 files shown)


far/clipboard.cpp
```diff
@@ -371,6 +371,19 @@ bool clipboard::SetHDROP(const string_view NamesData, const bool bMoved)
 	if (!Clear() || !SetData(CF_HDROP, std::move(Memory)))
 		return false;
 
+	 string TextData;
```
alabuzhev: This is how (more or less) it was implemented initially in 1609.
Later it turned out that copying both format does more harm than good, so in 1751 these actions were separated.

Overall, I'm not convinced that we need this change.
To copy cmdline selection you can turn panels off and use Ctrl+C, or use Ctrl+Ins anywhere, or write a macro for any other key.

kyky: Why you think that hdrop so important. I have no ideas who use it. The copying files in FAR is easily that clipboarding them to Explorer. Why you make copying so hard to understand for this unuseful feature? If it so important to you please specify any other key for this. Using different keys for copying is alogically and give bad user experiense.

Yes it possible to turn panels off to use Ctrl+C, keep in mind the Ctrl+Ins alias, trying to write macros to newly installed FAR at client office but why don't make FAR a little better and add 2 strings of code?????????????????????????????????????????????????????

alabuzhev: I don't think it's "so important". I don't mind rebinding it to another hotkey (say, Ctrl+Shift+C) or removing it altogether, but your PR does neither of that. Instead it just adds more bizarre logic around it.

PR #216, opened 2019-12-12
Added method for read-only edit fields to dialog builder.

lazyhamster:
New method for FarDialogBuilder class - AddReadonlyEditField.
Currently to add read-only edit filed one has to create temporary writable buffer, copy content to it and then use AddEditField and set DIF_READONLY. New function allows to it in one call and avoid creating redundant buffer making code more cleaner

### DIFF
--- a/far/DlgBuilder.hpp
+++ b/far/DlgBuilder.hpp
@@ -1072,6 +1072,15 @@ public:
 			return Item;
 		}
 
+		FarDialogItem *AddReadonlyEditField(const wchar_t *Value, int Width)
+		{
+			FarDialogItem *Item = AddDialogItem(DI_EDIT, Value);
+			SetNextY(Item);
+			Item->X2 = Item->X1 + (Width > 0 ? Width : TextWidth(*Item)) - 1;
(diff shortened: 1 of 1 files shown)


far/DlgBuilder.hpp
```diff
@@ -1,4 +1,4 @@
-﻿#ifndef DLGBUILDER_HPP_E8B6F5CA_37A9_403A_A3F0_9ED7271B2BA7
+#ifndef DLGBUILDER_HPP_E8B6F5CA_37A9_403A_A3F0_9ED7271B2BA7
```
alabuzhev: What's been changed here?

lazyhamster: Sorry about that. Accidentally lost BOM. Fixed.

alabuzhev: Ok, squash please.

lazyhamster: Done

alabuzhev: Thanks

PR #222, opened 2020-01-15
Buffer overflow fix

Ajaja:
Details:
https://forum.farmanager.com/viewtopic.php?p=158555#p158555

### DIFF
--- a/plugins/multiarc/custom.cpp
+++ b/plugins/multiarc/custom.cpp
@@ -989,9 +989,9 @@ void ParseListingItemRegExp(Match match,
 {
 
     if(const char *p = match["name"])
-        lstrcat(Item->FindData.cFileName, p);
+        lstrcpyn(Item->FindData.cFileName, p, sizeof(Item->FindData.cFileName));
     if(const char *p = match["description"])
-        lstrcat(Info->Description, p);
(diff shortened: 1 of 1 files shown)


johnd0e:
The commit message is bad.

plugins/multiarc/custom.cpp
```diff
@@ -989,9 +989,9 @@ void ParseListingItemRegExp(Match match,
 {
 
     if(const char *p = match["name"])
-        lstrcat(Item->FindData.cFileName, p);
+        lstrcpyn(Item->FindData.cFileName, p, sizeof(Item->FindData.cFileName));
```
alabuzhev: Why the original code was appending the strings rather than copying them?

Ajaja: Item - this structure must be filled:
https://github.com/FarGroup/FarManager/blob/86c26a80a21e24e3c2a7b4d259a76d3ef34d44fc/plugins/multiarc/README.TXT#L65

Ajaja: I think it always should be empty
https://github.com/FarGroup/FarManager/blob/86c26a80a21e24e3c2a7b4d259a76d3ef34d44fc/plugins/multiarc/arcread.cpp#L97

PR #229, opened 2020-03-09
Translated more far:config help pages into English

MKadaner:
Also, cleaned up eng/TechInfo.txt.


far/FarEng.hlf.m4
```diff
@@ -6179,144 +6179,164 @@ copied.
...
- System.MsHWheelDeltaView - в программе просмотра
- System.MsHWheelDeltaEdit - во встроенном редакторе
- System.MsHWheelDeltaHelp - в системе помощи
- System.MsHWheelDelta     - в прочих областях
+ System.MsHWheelDeltaView - ^<wrap>in the internal Viewer
+ System.MsHWheelDeltaEdit - in the internal Editor
+ System.MsHWheelDelta     - in other areas
 
- По умолчанию значение = 1
+ Default value: 1 (for all parameters).
```
MKadaner: Need to add note that `Alt+Wheel` always scrolls one line / character at at time.

MKadaner:
Ready for review. I'll do all formalities when the change is approved.

far/FarEng.hlf.m4
```diff
@@ -6179,144 +6179,167 @@ copied.
...
 
- Изменение этого параметра возможно через ~far:config~@FarConfig@
+ This parameter can be changed via ~far:config~@FarConfig@ only.
 
 
 @System.WordDiv
 $ #far:config System.WordDiv#
- Параметр "System.WordDiv" позволяет задавать дополнительные (кроме пробела и табуляции) cимволы-разделители слов.
+ This string parameter defines additional word delimiters besides
+#Space# (U+0020) and #Tab# (U+0009).
```
alabuzhev: Do we really need to mention the corresponding unicode code points for space and tab?

far/FarEng.hlf.m4
```diff
@@ -6179,144 +6179,167 @@ copied.
...
+settings dialog.
 
 
 @Editor.FileSizeLimit
 $ #far:config Editor.FileSizeLimit#
- Параметр "Editor.FileSizeLimit" задаёт максимальный размер редактируемого файла в байтах.
-Если размер редактируемого файла превышает максимально допустимый, то будет показано предупреждающее сообщение перед
-открытием такого файла.
+ This numeric parameter defines the maximum size of a file in bytes
+which can be safely loaded into Editor.
```
alabuzhev: "Safely" sounds a little alarming, like "something bad is going to happen if I proceed".
Maybe something like "reasonable size"?

It's probably worth explaining why this can be useful in the first place, e.g. "since the whole content of a file is read into memory, accidental opening of very large files could be undesirable".

far/FarEng.hlf.m4
MKadaner: Probably not. Removed everywhere. Thanks.

far/FarEng.hlf.m4
MKadaner: Yeah, probably. I was not happy with this article as well. Thank you for pointing out. I reworked it entirely, both English and Russian. See if you like it this way.

alabuzhev:
Thanks!

PR #254, opened 2020-07-17
Changelog cleansing

MKadaner:
I am playing with combining changelogs. It will be helpful, if the headings match exactly between the files.


misc/changelog/changelog.linq
```diff
@@ -0,0 +1,176 @@
+<Query Kind="Program">
```
MKadaner, line 1: I am not going to merge this script, but it'll be safer if you review it.

far/changelog
```diff
@@ -19616,7 +19580,7 @@ t-rex 19.04.2011 03:40:18 +0200 - build 1973
 
 1. folder shortcuts в sqlite.
 
-t-rex 19.04.2011 01:09:36 +0200 - build 1972 - хаг самэах!
+t-rex 19.04.2011 01:09:36 +0200 - build 1972 - Hag Sameach!
```
alabuzhev, line 19583: I don't know the language, but google [suggests](https://www.google.com/search?q=hag+sameach) "chag".

johnd0e: Why not just use Hebrew original

shmuz: > Why not just use Hebrew original

I can not use Hebrew in Far because of font issues. Probably ConEmu can but I did not try.

johnd0e: Ok. With https://alittlehebrew.com/transliterate/ I've got `chg shmch`.

MKadaner: Thank you for checking, @alabuzhev. I simply copied the words from the English version trusting t-rex; should have researched myself ;(

https://alittlehebrew.com/transliterate/ depending on style gives:

Style                | Transliteration
---------------------|-----------------
Ashkenazi            | `chag someiach`
Sefardi (advanced)   | `chag saméach`
Sefardi (simplified) | `Chag sameach`
technical            | `ḥaḡ śå̄mēaḥ`

[Wikipedia](https://en.wikipedia.org/wiki/Jewish_greetings) suggests `Chag sameach`, which seems to be consistent with [Bing](https://www.bing.com/search?q=hag+sameach) and [Google](https://www.google.com/search?q=hag+sameach) suggestions. Let me settle with this variant.

alabuzhev: ..._and here, gentlemen, you can see The Bike Shed Effect at its finest._

PR #268, opened 2020-09-28
Enable saving state of unmounted volumes display

bitraid:
This is a very small change, to make possible for the state of the recent option of displaying the unmounted volumes in Disk menu (4c26eecea888faf4c1c9c7b182ac749f003e80c5), to be remembered with the `Save setup` command. I (and i guess many more users) find the volumes display very useful and i would like to have it *on* by default, so i hope this (or something similar) would be added. Thanks.


alabuzhev:
> I (and i guess many more users) find the volumes display very useful and i would like to have it on by default

I thought it'd be an extremely niche feature :)

The changes look good. Could you please update vbuild.m4 and changelogs, e.g. something like this in both: 
```
bitraid 28.09.2020 13:00:01 +0100 - build 5673

1. gh-268: Enable saving state of unmounted volumes display
```

bitraid:
Files updated!

alabuzhev:
And squash please

bitraid:
Sorry for not squashing right away - some repos don't like that.

Issue #253, opened 2020-07-13
Combine changelog and changelog_eng


MKadaner:
Starting at 64c221090e27a3adf8e9ed82a5f3fa7d6f785d22, file `changelog_eng` is considered the first-class official Far Manager change list equal to the Russian version at `changelog`. Yay!

Now, every developer have to update two files with the summary of their changes. It's not that hard, but still taxing. How about combining the files together? There is obvious advantage of lower maintenance. Not only one file need to be updated instead of two, but some (many?) records, like `Refactoring` or `Correction of` can be made in English only; hopefully, such simple comments can easily be understood by any reader.

Here is a proposal of how the combined file might look like: [far/changelog.md](https://github.com/MKadaner/FarManager/blob/mzk-changelog/far/changelog.md). And here is the comment by @alabuzhev to start the discussion:

> Some points to consider:
> - The files are huge: github already can't render changelog and changelog_eng is close to that. Merging them into even huger one won't improve the situation, so any formatting doesn't really make sense.
> - It might be easier to update the merged version, but I'm not sure about reading it - a reader will have to jump over several lines over and over again.
>
> _Originally posted by @alabuzhev in https://github.com/FarGroup/FarManager/pull/252_

File size is indeed an issue. However, it is not strictly tied to combing the files together. They are already large; something need to be done, combining or not. A possible solution is to split the file(s) at strategic places, like it was done with famous [Ralf Brown's Interrupt List](https://www.cs.cmu.edu/~ralf/files.html). Then, the pieces could be linked together with ordinary hyperlinks. Global search over the entire contents would suffer, especially online. How often global search is required, especially online?

Readability is a real problem. Do you have any notion of how often people read change log? Long ago, to decide whether I wanted to update Far at certain point, I tried to read change log and glean what new nice features were implemented or annoying bug were fixed. It was rather difficult anyway, mostly for three reasons:

1. No highlighting. The headings and comments are not easily distinguishable, at least for my poor eyesight. This could be neatly fixed with markdown.
2. Lots of "uninteresting" records like `Refactoring` or `Correction of`.
3. Sometimes, it's difficult to understand whether a new feature considered completed or still in experimental phase; there is a record saying "added such and such feature," then several records about corrections and continuations, but usually there is no final record saying it's production ready.

(The latter two issues are traditionally solved by maintaining release notes, but writing them is a task on its own. I do not think this overhead is manageable for such a small team as Far developers'.)

[...]

PR #323, opened 2020-12-26
Combining changelogs. Please review. DO NOT MERGE!

MKadaner:
The purpose of this PR is to facilitate the review of the proposed combining of Russian and English changelog versions. After the automation and folding of records are reviewed, only the combined changelog should be merged. The goal of the change is to replace two changelog files with the single combined changelog. For the background discussion see #253.

Terminology: Changelog consists of _records_ describing the change, one record per commit. Each record starts with a well-formatted _header_ line and contains one or more often numbered _items_. The records in Russian and English changelogs can be correlated by the headers. If the number of items in the correlated records are the same, the items within the records can be correlated by their order or, if the items are numbered, by the numbers. In the following discussion such correlation is called _structural_.

Structural correlation is used to combine English and Russian changelogs into a single bilingual file. Each record in this file starts with the common header, followed by the English version, separator, and the Russian version. In many cases, English and Russian parts of a combined record are trivial translations of each other, for example:
```
--------------------------------------------------------------------------------
drkns 25.10.2020 01:11:40 +0000 - build 5688

1. Refactoring.

· · · · · · · · · · · · · · · · · · · · · · · · ·

1. Рефакторинг.

```
Obviously, it would be nice to _fold_ such bilingual records and keep only English variant. The challenge in deciding which records are foldable is in the definition of “trivial translation.” To make this determination, let’s imagine an advanced computer user (all Far users are presumably advanced) who is a native Russian speaker but has only limited command of English. From the perspective of such a user, let’s consider whether such and such Russian translation adds any value to the changelog record. The value may be either in better understanding of the nature of the associated change, or simply in enjoinment of idiomatic Russian as opposed to neutral English. The boundary is of course fuzzy. What is worse, personally I have had hard time imagining an advanced computer user who nevertheless is only a beginner English speaker. Hence, my decisions of which items and records are foldable were very subjective. As such, I ask reviewers to challenge my decisions, whether to fold certain records or to refrain from folding.

This PR contains four artifacts:
1. `changelog_00`: the combined changelog proposed to replace existing Russian and English changelog files. Out of 4889 bilingual records, 1686 were folded.
2. `changelog.linq`: the script which combines Russian and English changelogs and folds the records whenever possible. Combining is based on structural correlation of records and items, and has several built-in checks ensuring correctness. Folding is based on the collection of hard coded regexes stored in the `translations` array (line 25). Please check if any of these regexes feel too aggressive. Grouping of the regexes is for performance optimization only and unimportant for the review.
3. `folded_records`: transcript of all records folded in `changelog_00` with Russian version preserved for verification.
4. `rejected_candidates.csv`: transcript of all structurally correlated and short (single line, 50 or less characters) items which were _not_ folded because I considered Russian translations valuable. Please check if any of these items should be folded. Again, grouping is an artifact of performance optimization and does not carry any semantical meaning.


MKadaner:
Combined changelog is on master, closing this supporting PR.

PR #324, opened 2020-12-27
`Alt+<Gray *>` inverts selection on files only, excluding folders.

MKadaner:
Generally, I prefer dealing with folders pretty much like with files. Thus, `Select folders` option of `Panel settings` is turned on. However, occasionally I find myself inverting selection (usually after some panel comparison) and immediately _manually deselecting folders_. The proposed `Alt+<Gray *>` key combination is intended to do this in one go.

The new key combination can be considered a counterpart of the existing `Ctrl+<Gray *>` one:
- `Ctrl+<Gray *>` inverts selection _including_ folders even if `Select folders` option is turned _off_.
- `Alt+<Gray *>` inverts selection _excluding_ folders even if `Select folders` option is turned _on_.

If you think this feature may be useful, please merge this PR.


far/FarEng.hlf.m4
```diff
@@ -353,9 +353,9 @@ $ #Panel control commands#
  Deselect files with the same extension as the        #Ctrl+<Gray ->#
  current file
  Invert selection including folders                   #Ctrl+<Gray *>#
- (ignore command line state)
```
MKadaner, line 356: I did not notice any interactions of any hotkeys in this group with command line state. Even `<Gray *>` apparently is not affected by the state of the command line. There is no reason to make this note.

alabuzhev:
Looks good, thanks

MKadaner:
Thank you!

alabuzhev:
There are some doubts about the wording: https://forum.farmanager.com/viewtopic.php?p=163520#p163520

MKadaner:
Thank you for pointing out. I replied on the forum.

PR #339, opened 2021-01-15
Customization of default history key in F2 user menu input field (UserVar)

yogdg:
The idea is to be able to replace the default history key used while inputting text in the user menu with one provided by the end user. This PR will allow Far to save the input text by a different key, provided the title ends with a *|*. So, for example, an input field (o serveral ones) can have its own private history without altering the rest of the input fields defined in the user menu.

### DIFF
--- a/far/FarEng.hlf.m4
+++ b/far/FarEng.hlf.m4
@@ -2261,6 +2261,9 @@ executing command. <title> and <init> - title and initial text of edit control.
 
  Leave the name empty to disable history.
 
+ If the title field ends with a #|# the title itself is used internally as the history key for the input text (instead of #UserVarN#)
+allowing different entries in the menu to have different input text histories.
+
  The entered string can also be accessed later as #%<history># (or as #%UserVarN#, where N is the index of the corresponding input).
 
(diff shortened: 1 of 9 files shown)


far/FarEng.hlf.m4
```diff
@@ -2261,6 +2261,9 @@ executing command. <title> and <init> - title and initial text of edit control.
 
  Leave the name empty to disable history.
 
+ If the title field ends with a #|# the title itself is used internally as the history key for the input text (instead of #UserVarN#)
```
alabuzhev, line 2264: It's a neat idea. However, have you seen the previous paragraph?

yogdg: Yes, but I couldn't figure out what `name` is referring to.
I 've tried `notepad.exe !??!`, but it keeps both, history and last input text.

alabuzhev: [`!?$<history>$<title>?<init>!`](https://github.com/FarGroup/FarManager/blob/13d47a573f87c8a2a3d5bf4d1e9ae07e98ed80b2/far/FarEng.hlf.m4#L2258-L2260)

yogdg: Nice, closing the PR

PR #349, opened 2021-02-11
Updated Russian translation

snxx-lppxx:
Fixed incorrect translation. Removed incorrect abbreviations

### DIFF
--- a/far/farlang.templ.m4
+++ b/far/farlang.templ.m4
@@ -442,7 +442,7 @@ l:// Panels: functional keys
 "Pagalb"
 
 MF2
-"Меню пользователя"
+"ПользМ"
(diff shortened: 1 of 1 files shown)


far/farlang.templ.m4
```diff
@@ -442,7 +442,7 @@ l:// Panels: functional keys
 "Pagalb"
 
 MF2
-"ПользМ"
+"Меню пользователя"
```
alabuzhev, line 445: These strings are short for a reason: they should fit and make some sense when the window isn't maximised.

alabuzhev: This, obviously, applies to all the changed F-key labels, not just this one.

PR #350, opened 2021-02-12
Polish help file - small corrections

BestiaPL:
Visual formatting.

### DIFF
--- a/plugins/compare/CmpPol.hlf
+++ b/plugins/compare/CmpPol.hlf
@@ -3,41 +3,43 @@
 
 @Contents
 $ #Zaawansowane porównywanie#
- Wtyczka jest podobna do polecenia Far "Porównaj foldery", ale dodatkowo można ustawić dodatkowe opcje porównywania:
+    Wtyczka jest podobna do polecenia FAR "Porównaj foldery",
+ale dodatkowo można ustawić dodatkowe opcje porównywania:
 
- #Przetwarzaj:#
+#Przetwarzaj:#
 
- #Podfoldery#
- Porównuje pliki w podfolderach. Nie może być użyte dla paneli wtyczek.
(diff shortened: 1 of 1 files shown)


plugins/compare/CmpPol.hlf
```diff
@@ -3,41 +3,43 @@
...
- Wtyczka jest podobna do polecenia Far "Porównaj foldery", ale dodatkowo można ustawić dodatkowe opcje porównywania:
+    Wtyczka jest podobna do polecenia FAR "Porównaj foldery",
+ale dodatkowo można ustawić dodatkowe opcje porównywania:
 
- #Przetwarzaj:#
+#Przetwarzaj:#
 
- #Podfoldery#
- Porównuje pliki w podfolderach. Nie może być użyte dla paneli wtyczek.
+  #Podfoldery#               Porównuje pliki w podfolderach. Nie może być użyte dla paneli wtyczek.
```
alabuzhev, line 11: Please keep this style:
```
 Title
 Description text

 Another title
 Another description text
```

Two column mode is a nice thing, but:
- it's hard to get it right here due to the hlf format limitations
- it will almost never look good when the window size is different from yours
- it's too time consuming to maintain, update and synchronize.

Also, please keep the number of paragraphs consistent with the English file, and, in general, please avoid _any_ changes in formatting.

If the formatting of this file diverges from the English one, we won't be able to keep them synchronized. This means that instead of untranslated English pieces there will be no updates at all, and anyone who would like to translate it would need to analyze and pick all the updates from the English file manually.

BestiaPL: OK. I undestand.
Please, remove my changes (cancel my commit).

alabuzhev: Ok

PR #364, opened 2021-02-24
Polish lang - corrections

BestiaPL:
Proofreading for a better understanding of the definition.

### DIFF
--- a/far/FarPol.hlf.m4
+++ b/far/FarPol.hlf.m4
@@ -57,7 +57,7 @@ $^#Copyright © 1996-2000 Eugene Roshal#
  ~O programie Far~@About@
  ~Licencja (po angielsku)~@License@
 
- ~Przełączniki linii poleceń~@CmdLine@
+ ~Przełączniki wiersza poleceń~@CmdLine@
  ~Skróty klawiszowe~@KeyRef@
  ~Obsługa wtyczek~@Plugins@
  ~Przegląd możliwości wtyczek~@PluginsReviews@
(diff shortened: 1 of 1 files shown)


far/FarPol.hlf.m4
```diff
@@ -1922,10 +1923,12 @@ i zawiera wszystkie uwzględnienia oraz żadnego z atrybutów wykluczenia:
...
  #[?]# - ^<wrap>atrybut jest ignorowany.
 
- Atrybuty #Skompresowany#, #Zaszyfrowany#, #Sparse#, #Tymczasowy# i #Offline# używane
-są tylko na dyskach z systemem NTFS.
-Atrybuty #Strumień integralności# i #Brak danych kontrolnych# są obsługiwane tylko
-na wolumenach ReFS począwszy od Windows Server 2012.
+ Atrybuty #Skompresowany#, #Zaszyfrowany#, #Nie zaindeksowany#, #Rzadki#, #Tymczasowy#, #Offline#
+i #Punkt dowiązania# są dostępne tylko na dyskach z systemem NTFS.
+Atrybuty #Skompresowany# i #Zaszyfrowany# wykluczają się wzajemnie, można użyć tylko jednego z nich.
+Nie można wyczyścić atrybutu #Rzadki# w Windows 2000/XP/2003.
```
alabuzhev, line 1929: I've removed these 2 lines in 396540436.

I understand the intent to sync this paragraph with FileAttrDlg, but the difference is deliberate:
- in FileAttrDlg the description of certain file system limitations is quite relevant because the user is actually about to interact with the file system and face those limitations
- this paragraph, however, describes the Filter configuration, which doesn't have such limitations: the user can set any combinations of attributes here, even if they make no sense on their file system.

BestiaPL: My mistake, I found two similar paragraphs and confused their order. Your version is correct.

PR #369, opened 2021-03-13
Update edit.cpp

snxx-lppxx:


alabuzhev:
@snxx-lppxx could you please describe what your changes do and why they're needed?
"Update edit.cpp" doesn't tell much.

snxx-lppxx:
I have shortened the check. Sorry my oversight.

far/edit.cpp
```diff
@@ -421,10 +421,9 @@ bool Edit::ProcessInsPath(unsigned int Key,int PrevSelStart,int PrevSelEnd)
 			return false;
 		Data.Folder = os::env::expand(Data.Folder);
 	}
-	else
+	else if (!MakePathForUI(Key, Data.Folder))
```
MKadaner, line 424: Please do not do this. While technically equivalent, the `else if` construct with a condition irrelevant to the condition in the main `if` clause is often confusing and more difficult to interpret.

alabuzhev:
@snxx-lppxx, while contributions that improve and shorten the code are definitely welcome, I tend to agree with Michael here: not every `else + if` is `else if`. It can be useful sometimes, but in this particular case the nested condition is error handling *within* the "else" branch, moving it upper won't really make the code clearer.

Nevertheless, thank you for trying to improve it.

PR #370, opened 2021-03-14
update build scripts for colorer

ctapmex:
change branch for build


misc/nightly/colorer.sh
```diff
@@ -47,6 +47,7 @@ cd ..
 #git clone must already exist
 cd FarColorer || exit 1
 rm -fR build
+git checkout origin/build-from-bar
```
alabuzhev, line 50: `|| exit 1`?

PR #373, opened 2021-03-22
HlfChecker compares references in paragraphs.

MKadaner:
Added paragraph comparison to HlfChecker script. So far, only collections of references (e.g., "@About@") in corresponding paragraphs are compared. There are two comparison modes, strict and lax. In lax mode, the collections are sorted and repeated references are removed before comparing. In strict mode, the collections are compared exactly as extracted.

Also, fixed some inconsistencies in English and Russian Far Hlf files. As of this commit, there are two lax (serious) mismatches between English and Russian help files. I need more time to investigate which help better describes the functionality. Three additional mismatches in strict mode are in reference order only and likely benign.


alabuzhev:
@MKadaner awesome, thank you. It captured so many little issues.

A couple of wishes:
- Is it possible to minimise the output of the script (reading... parsing... comparing... etc.) by default? Printing only detected issues would be ideal.
- Is it possible to include the line numbers into the output?

MKadaner:
Sure, will do as time permits.

PR #374, opened 2021-03-22
Reorder python file

snxx-lppxx:
Changed a little.

### DIFF
--- a/far/tools/translation.summary.merge.py
+++ b/far/tools/translation.summary.merge.py
@@ -2,37 +2,37 @@ import sys
 
 def patch_feed(feed_file_name, lng_id, new_lines):
 	lines = []
-	with open(feed_file_name, "r", encoding='utf-8-sig') as feed_file:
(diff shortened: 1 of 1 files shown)


alabuzhev:
@snxx-lppxx could you please explain what your PR does, why it's needed etc.?

snxx-lppxx:
@alabuzhev, changed the code style. The code will work the same, but looks better.

techtonik:
@snxx-lppxx usually people don't add such commit, because they add extra steps while navigating source history for potential problems. But adding https://github.com/psf/black to CI may by a good choice to avoid wasting time on code formatting matters.

alabuzhev:
@snxx-lppxx, what @techtonik said.

Adhering to a single style is a good thing in general, but replacing `"` with `'` in some supplementary script doesn't look that useful, it will only add noise.

Technically, PEP 8 [explicitly abstains from enforcing this](https://www.python.org/dev/peps/pep-0008/#string-quotes), so it's a matter of taste. Since our primary language is C++ here, double quotes just look more natural around strings.

PR #384, opened 2021-04-12
Update FarPol.hlf.m4

BestiaPL:
In hex browser (lines 2933-2938) i have used Polish text, which show all characters used in Polish language. Also changed CP1252 to CP1250 - this is default code page in all MS Windows.


BestiaPL (commented):
@InfoPanelSettings topics count - explanation.
In English version is 26 topics, because there is an unnecessary blank line there (compare lines 1198-1199 and 1227-1228 in FarEng.hlf.m4) .

PR #386, opened 2021-04-17
"Out of tree" nightly builds for Encyclopedia

techtonik:
This makes script more readable to understand if builds are deterministic and cleanup before the build is done right.

### DIFF
--- a/misc/nightly/enc.sh
+++ b/misc/nightly/enc.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# absolute path to this script
+DIR=$(dirname $(readlink -f $0))
+# absolute path to source root
+ROOT=$(cd $DIR/../.. && pwd)
+BUILDDIR=$ROOT/_output
+
+
 function benc2 {
 LNG=$1
 L=$2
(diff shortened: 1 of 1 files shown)


alabuzhev:
Unfortunately this broke the build. Reverted for now.

Note: on the build server these scripts are copied into the root directory (i.e. where the enc folder is) and invoked from there.

techtonik:
So that's the build server that needs to be fixed then?

alabuzhev:
> So that's the build server that needs to be fixed then?

Maybe.
As I said earlier, debugging this is a pita.
Personally I find bash too abominable to even look at, let alone touch, sorry. I'd rather rewrite these scripts in python.
You might want to talk to @trexinc directly, he knows how it works.

techtonik:
Can you point me to the build that failed?

Why nightly builds don't use CI scripts in this repo?

alexsatori:
Can you explain why you modify scripts you don't understand?

techtonik:
@alexsatori this one quite obvious - I wanted to patch encyclopedia, but the script failed to build it from plain checkout, so I started to modify the scripts until I could get the outpu. After all these years I don't remember what exactly I wanted to fix in the content, but at least other contributors will have less time figuring it out.

techtonik:
That was a hope, anyway.

alexsatori:
contributors do not use those scripts. if there are problems with enc build or something, report that. Don't touch those scripts.

techtonik:
I am contributor and I've used that script. Who are you then?

alexsatori:
That script is not for you, it's for a private build server

techtonik:
So why are you hosting proprietary build scripts in an open source repo?

alexsatori:
Because we can

techtonik:
If you can this, maybe you can as well make the public CI build services use the same scripts.

alexsatori:
Impossible

techtonik:
Then shut down your server. It is not needed with GitHub Actions and Releases.

alexsatori:
Thank you for your opinion

PR #387, opened 2021-04-21
Remove outdated changelogs

techtonik:
There is repository history now.


alabuzhev:
I believe we [discussed this already](https://github.com/FarGroup/FarManager/pull/49) some time ago.

techtonik:
The nail that still sticks in my head. Whatever..

johnd0e:
Well, those changelogs are actually outdated.
What'd be useful - track api changes in some document (and we have one, outdated too).

alabuzhev:
Even outdated, these logs still can be useful. I see no harm in having them there.
Updates are always welcome.

PR #397, opened 2021-05-10
wrong order of version elements

ctapmex:
Wrong order of version elements. Dispaled in F9 - Options - Plugins configuration


johnd0e:
>  while always staying the same feature-wise (and hence with the same "build").

Where I can read more about such versioning scheme?

PR #414, opened 2021-06-05
get latest version of FarColorer from colorer repo

ctapmex:
and update to 1.4.5

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -108,7 +108,7 @@ build_script:
   - copy /Y Far-NetBox\*.md ci\%configuration%.%platform%\Plugins\NetBox
   - copy /Y Far-NetBox\LICENSE.txt ci\%configuration%.%platform%\Plugins\NetBox
   #build FarColorer
-  - for /F "delims=xxx" %a in ('curl -s --ssl-no-revoke https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt') do set COLORER_VERSION=%a
+  - for /F "delims=xxx" %a in ('curl -fsSLJ --ssl-no-revoke https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt') do set COLORER_VERSION=%a
   - set COLORER_PLATFORM=x86
   - if "%platform%" == "x64" set COLORER_PLATFORM=x64
   - set COLORER_FILE_NAME=FarColorer.%COLORER_PLATFORM%.v%COLORER_VERSION%.7z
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -108,7 +108,7 @@ build_script:
   - copy /Y Far-NetBox\*.md ci\%configuration%.%platform%\Plugins\NetBox
   - copy /Y Far-NetBox\LICENSE.txt ci\%configuration%.%platform%\Plugins\NetBox
   #build FarColorer
-  - set COLORER_VERSION=1.4.4
+  - for /F "delims=xxx" %a in ('curl -fsSLJ --ssl-no-revoke "https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt"') do set COLORER_VERSION=%a
```
alabuzhev, line 111: - `%` should be doubled in scripts, e.g. `%%a`  in both places (probably that's why it fails now).
- `"delims=xxx"` should probably be `"delims="`.

misc/nightly/colorer.sh
```diff
@@ -3,7 +3,7 @@
 function bcolorer {
   BIT=$1
   PLUGIN=FarColorer
-  COLORER_VERSION=1.4.4
+  COLORER_VERSION=$(curl -s https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt)
```
alabuzhev, line 6: curl keys & quoting should probably be the same in both cases.

appveyor.yml
alexsatori: why do you need --ssl-no-revoke? would be better without it.

ctapmex: it wasn`t work without --ssl-no-revoke on local test in windows console

misc/nightly/colorer.sh
```diff
@@ -3,12 +3,12 @@
 function bcolorer {
   BIT=$1
   PLUGIN=FarColorer
-  COLORER_VERSION=$(curl -s https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt)
+  COLORER_VERSION=$(curl -fsLJ 'https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt')
```
alexsatori, line 6: COLORER_VERSION=$(curl -fsLJ https://raw.githubusercontent.com/colorer/FarColorer/master/version4far.txt)
if [ -z "$COLORER_VERSION" ]; then
  echo "Failed to get Colorer version"
  return 1
fi

alexsatori:
Looks good now. One last thing. For both build types - print the version so it will exist in the logs.

PR #417, opened 2021-06-16
luamacro: more introspection with mf.GetAllScriptsCopy

johnd0e:
by IgorZ: https://bugs.farmanager.com/view.php?id=3536#c17088


johnd0e:
Close in favor of #420

PR #420, opened 2021-06-20
LuaMacro: more introspection with mf.EnumScripts

johnd0e:
New function EnumMacros, ref: https://bugs.farmanager.com/view.php?id=3536

MacroBrowser is now able to list such entities as well:
- Items of Plugins menu
- Prefixes
- Panel modules
- Content columns
- Custom sort modes


johnd0e:
@shmuz

johnd0e:
@zg0 @w17

PR #435, opened 2021-08-14
Update FarPol.hlf.m4

BestiaPL:
Updated Polish translation.
Also in EditorSettings topic:
- changed order of help items, according to menu order
- reformatted as ViewerSettings topic.


alabuzhev:
> * changed order of help items, according to menu order
> * reformatted as ViewerSettings topic.

That's a good idea, thank you.
However, as you've probably noticed, we have automated checks that validate that all the *.hlf.m4 files have the same structure, so updating only one of them breaks the build.

It's usually better to separate translations from reordering, reformatting etc.
If you see any inconsistences, please feel free to submit the translation as is and raise an issue to address structure and formatting changes separately.

PR #463, opened 2021-10-29
Panel API: ignore PluginId in GetPluginDirList because hPanel is enough

johnd0e:
by @shmuz

Related LuaFar bug: https://forum.farmanager.com/viewtopic.php?t=12639

Some considerations:
- [x] [Related](https://api.farmanager.com/ru/service_functions/getplugindirlist.html) Encyclopedia article should be changed too


far/dirinfo.cpp
```diff
@@ -354,8 +354,14 @@ static bool GetPluginDirListImpl(Plugin* PluginNumber, HANDLE hPlugin, string_vi
...
 	else
 	{
-		DirListPlugin = std::make_unique<plugin_panel>(PluginNumber, hPlugin);
-		hDirListPlugin = DirListPlugin.get();
+		const auto aHandle = Global->CtrlObject->Cp()->ActivePanel()->GetPluginHandle();
+		const auto pHandle = Global->CtrlObject->Cp()->PassivePanel()->GetPluginHandle();
+		if (aHandle->panel() == hPlugin)
+			hDirListPlugin = aHandle;
+		else if (pHandle->panel() == hPlugin)
+			hDirListPlugin = pHandle;
```
johnd0e, line 362:
```suggestion
		if (const auto aHandle = Global->CtrlObject->Cp()->ActivePanel()->GetPluginHandle(); aHandle->panel() == hPlugin)
			hDirListPlugin = aHandle;
		else if (const auto pHandle = Global->CtrlObject->Cp()->PassivePanel()->GetPluginHandle(); pHandle->panel() == hPlugin)
			hDirListPlugin = pHandle;
```
(by @alabuzhev)

shmuz: If I understand it correctly it is the same code written differently (more concise) doesn't it?

alabuzhev: It's the same, except that pHandle is only queried when actually needed.

PR #464, opened 2021-10-30
Update Edit.OpenURL.lua

sidorsett:
## Summary
Make URLs to open on Ctrl+Left Mouse Click as common to many other Windows applications

 
## Checklist

## Details
~~I have deleted the "area" key by mistake. Will need to correct.~~

### DIFF
--- a/extra/Addons/Macros/Edit.OpenURL.lua
+++ b/extra/Addons/Macros/Edit.OpenURL.lua
@@ -19,8 +19,23 @@ local patt = regex.new( [=[
 
 Macro {
   description="Open URL or UNC under cursor";
-  area="Editor"; key="CtrlEnter CtrlNumEnter";
+  area="Editor"; key="CtrlEnter CtrlNumEnter CtrlMsLClick";
+  condition=function(trigger)
+    mouseClick = trigger:match("Ms")
+    if not mouseClick then return true end
+    local info = editor.GetInfo()
+    local maxY = info.WindowSizeY - 1
+    minY = 0
+    if 0 ~= bit64.band(info.Options, far.Flags.EOPT_SHOWTITLEBAR) then
+      minY, maxY = 1, maxY + 1
+    end
+    return Mouse.Y >= minY and Mouse.Y <= maxY
+  end;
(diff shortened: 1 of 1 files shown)


shmuz (commented):
It's OK except that Ctrl-LeftClick on (function-keys line) F10 ("Go to file") stops working.

sidorsett:
Hi,

Thanks for paying attention on that. Will have a look if I can manage that.

Cheers!

shmuz:
I suggest a somewhat simpler condition:
```lua
  condition=function(trigger)
    mouseClick = trigger:match("Ms")
    if not mouseClick then return end
    local info = editor.GetInfo()
    local maxY = info.WindowSizeY - 1
    if 0 ~= bit64.band(info.Options, far.Flags.EOPT_SHOWTITLEBAR) then
      maxY = maxY + 1
    end
    return Mouse.Y <= maxY
  end;
```

sidorsett:
Hi,

Wonderful! Thanks for your support. I had to update the code a bit so that to account presence of the status bar in the `action` too.

Cheers!

shmuz:
@sidorsett
I suggest not to cache `Mouse.X` and `Mouse.Y`. Each of these takes about 100 nanoseconds execution time, (i.e. practically zero time). So IMHO it'd be better to leave `condition()` as in my previous post, and use `Mouse.X` and `Mouse.Y` in `action()`.

sidorsett:
Done, even though I don't really understand the rationale. Does it save few bytes of memory? In return, the action can take place in a different from the click position, if the macros is delayed and the mouse moves away.

Cheers!

shmuz:
I think the probability of mouse position change between condition() and action() is extremely low. I didn't tried "to save few bytes" but rather to make the code simpler.
Nevertheless if you think that the variant with caching is better then go ahead and restore it. I'm OK with either of the 2 variants.

sidorsett:
Let's go with the current one. I trust your experience.

Thank you once again for your support.

Cheers!

alabuzhev:
Thanks for your contribution.

PR #470, opened 2021-11-05
Update Edit.OpenURL.lua

sidorsett:
## Summary

If the link contains spaces, the macros catches only part of the link. Generally, it is not possible to catch all types of links correctly in principle. This change allows users to select the desired fragment so that the macros interpret the selection as a link if the pattern-based approach doesn't work sometimes.

 
## Checklist

### DIFF
--- a/extra/Addons/Macros/Edit.OpenURL.lua
+++ b/extra/Addons/Macros/Edit.OpenURL.lua
@@ -4,6 +4,11 @@
 -- Confirm launching the URL under cursor. Change it to false if you want no confirmations.
 local Confirm = true
 
+-- Change this to true to use selection (if present)
+-- instead of a fragment matching the pattern as the target URL.
+-- WARNING: setting this to true may be unsafe if you select and open a random code instead of an URL.
+local UseSelection = false
+
 local patt = regex.new( [=[
 (?:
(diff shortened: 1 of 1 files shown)


shmuz:
@sidorsett, I made 2 changes, please compare with your version.
[Edit.OpenURL.lua.zip](https://github.com/FarGroup/FarManager/files/7490044/Edit.OpenURL.lua.zip)

sidorsett:
@shmuz, thank you for the corrections. I applied your changes. However, I just understood that this allows execution of a potentially malicious code. Shouldn't the selection be verified against some other pattern for safety reasons? Please note that even if it is possible, I am not experienced with complex regex expressions and will not be able to compose one myself.

shmuz:
@sidorsett
1. Being familiar with regular expressions is not enough because considering all possible kinds of malicious code may be very hard or impossible.
2. We don't execute the code immediately but only after the user's confirmation
3. We may introduce an option at the file top, i.e.
```lua
local EnableRunSelection = false -- WARNING: setting this to true may be unsafe
```
and enable selection handling only if the user themselves changed it to true.

sidorsett:
Modified as suggested.

shmuz:
It's OK for me.

PR #471, opened 2021-11-05
Update Editor.ColorWord.moon

sidorsett:
## Summary

The macros is updated to search either for a word or for a selection (when present). In the latter case, the macros supports vertical blocks as well as simple selections with one or multiple lines.

 
## Checklist

### DIFF
--- a/extra/Addons/Macros/Editor.ColorWord.moon
+++ b/extra/Addons/Macros/Editor.ColorWord.moon
@@ -6,7 +6,9 @@ F=far.Flags
 color = far.AdvControl(F.ACTL_GETCOLOR, far.Colors.COL_EDITORTEXT)
 color.ForegroundColor, color.BackgroundColor = color.BackgroundColor, color.ForegroundColor
 colorguid=win.Uuid "507CFA2A-3BA3-4f2b-8A80-318F5A831235"
-words={}
+lines={}
+selStarts={}
+selTypes={}
 
 Macro
   area:"Editor"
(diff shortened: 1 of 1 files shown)


shmuz:
@alabuzhev, currently I doesn't have time for reviewing that.

alabuzhev:
> The macros is updated to search either for a word or for a selection (when present)

@sidorsett, looks like the logic has changed in the case when there's no selection: we're no longer looking for a *word*, but just for *any* match with the word under cursor, as if in F7 search with "Whole words" off.
E.g. pressing F5 on `a` highlights all `a` letters in the whole text.
Can it be made compatible with the previous behavior please? (look for whole words only if there's no selection)

sidorsett:
@alabuzhev, indeed it has changed. Thanks for paying attention on that, I didn't notice it myself!

Isn't it more natural and consistent though?
- When there is no selection, the macros gets the word under cursor and looks for all matches (not necessarily whole word matches).
-  When there is a selection, the macros gets the selection and similarly looks for all matches, independent of whether it is a whole word or not.
Therefore, the current behavior is consistent across the two options.

Btw is there any way to integrate this macros to the F7 search so that the macros got automatically activated on execution of F7 search?

alabuzhev:
I expect the results to be consistent with the input.

Without a selection it gets *the word*, so I expect it to highlight words only as well.
With a selection it doesn't care whether it's a word or not, so I expect it to highlight all matches.

Currently it's often unusable. Here I wanted to highlight the `i` variable and check where it is used in the loop, so I pressed F5 on `i` and...

image

sidorsett:
Indeed, your use case makes a lot of sense! At the same time, it is disputable whether the different behavior between two modes can be considered consistent. For example, a user can make a manual selection of `size_t i` from your exemplary code and then argue that `size_t iNode` gets highlighted. Whole word and non-whole word searches are clearly two different options both with their own use cases. I will have a look on the code and will consider what are the possible options. Please don't expect an instant resolution though. I will try to submit a commit by the end of the week.

PR #475, opened 2021-11-18
Improve standard macros

johnd0e:
- More user-friendly formatting for all std macros
- Miscellaneous improvements (more info in commit msgs).

Ref: https://forum.farmanager.com/viewtopic.php?t=7812

_to be continued_


johnd0e:
`OpeningLastFile.lua` (_Opening last view/edit of a file_) is not-functional since 5773, and it's hard to restore it's former functionality, because now we have no info about order of previously closed files.

extra/Addons/Macros/Panel.SpaceToSelect.lua
```diff
@@ -1,6 +1,10 @@
+local APANEL,INVERT,MODE_IDX,CUR_ITEM = 0,2,1,0
 Macro {
   description="Panel: Use Space to select files";
-  area="Shell"; key="Space";
+  area="Shell";
+  key="Space";
   condition=function() return CmdLine.Empty and APanel.Visible end;
-  action=function() Keys('Ins') end;
+  action=function()
+    Panel.Select(APANEL,INVERT,MODE_IDX,CUR_ITEM)
```
alabuzhev, line 8: Reverted.
Please don't smuggle breaking changes.

johnd0e: It was not breaking change, because it is not part of code. It is just a sample macro. Sample config.
But whatever. I do not care anymore.

alabuzhev: "Breaking" as in "breaking users' habits".

Remember that people are lazy and don't want to learn anything.
For you it's an example, a building block/inspiration for something larger, for laypeople it's a black box of magic.

I'm not saying it's set in stone and we can't change any user-observable behavior, but it's usually better to discuss and at least do it separately from formatting / refactoring.

johnd0e: > "Breaking" as in "breaking users' habits".

I still insist that changes in Addons do not break habits.
If someone already has macro for Space - it does not break.


> Remember that people are lazy and don't want to learn anything.

Far is not for so lazy people, as I can observe.
Because every month everything is changing in Far that brokes someones habits. And that was never seen as such a big issue.

> [...] but it's usually better to discuss and at least do it separately from formatting / refactoring.

I can agree that discussing it first is always better. That's why I created the forum's topics (though there is usually no noticeable community feedback).

BTW my PR introduced changes in behavior of some other macros too. So one can say that there were mixed formatting, refactoring, and functional changes.
But every commit was separate, so I do not see real issue here.

alabuzhev: > If someone already has macro for Space - it does not break.

Indeed, but from time to time people set it up from scratch.
Also, if I need to work on a new machine and don't have access to my profile, I might just run the installer and copy a couple of macros like Esc=CtrlO, Del=F8 and Space=Ins to make it borderline comfortable because I have neither time nor inspiration for more.

> every month something is changing in Far that brokes someones habits. And that was never seen as such a big issue.

Probably not every month, but yes, sometimes it happens, sometimes unintentionally, sometimes inevitably. But if we can avoid breaking things, then why break them?

> I created the forum's topics (though there is usually no noticeable community feedback)

I believe in this particular case there was a feedback. And, since you didn't object back then, we supposedly reached a concensus.

> my PR introduced changes in behavior of some other macros too

Oh. Not too dramatic changes hopefully?

> one can say that there were mixed formatting, refactoring, and functional changes.

Maintainer's POV:

I see a new PR.
I see it's something about formatting & refactoring in macros.
I see it's from John Doe.
I know that John Doe is a valuable and respectable member of the community and a long-time contributor who knows way more about macros than I do or ever will.
Therefore, I do not need to spend hours reviewing every single line and read 40 commit messages - I can merge it straight away and attend to my evening affairs.

But changes like "Adjusted things to my personal preferences" do need more attention and thus belong to a separate PR.
[Obligatory XKCD](https://xkcd.com/1172/).

> I still insist that changes in Addons do not break habits.

Macros like Panel.Esc, Panel.Del, Panel.BS and Panel.SpaceToSelect are not exactly some random examples, they are primordial interface options moved to macros a quarter of a century ago:
image

They always worked the way they do.
People expect them to work the way they do.
They are probably mentioned in more than one FAQ.
Someone probably deploys them automatically to new workstations.
And someone probably asked about them in CS exam questions.
That's why I suggested a new macro back in 2015 and still think it's the right thing to do.

On top of that, the Ins key is heavily neglected on modern keyboards, so it's only natural for this macro to replicate its behavior exactly, rather than *almost, but not quite, entirely unlike*.

johnd0e:
> I might just run the installer and copy a couple of macros like Esc=CtrlO, Del=F8 and Space=Ins to make it borderline comfortable

From my point of view Addons/Macros is far from being macro repository. It is so much far from it that I still doubt that someone in the world really do this. But I can be wrong here, sure, who knows.

> But if we can avoid breaking things, then why break them?

Because I saw it another way: add more useful macro sample, and remove trivial reassigning. To keep macro set tidy and clean.

> Macros like Panel.Esc, Panel.Del, Panel.BS and Panel.SpaceToSelect are not exactly some random examples, they are primordial interface options moved to macros a quarter of a century ago:

As it was changed so long ago, I do not see any reason to treat them in any special way anymore.

> [...]
> Someone probably deploys them automatically to new workstations.

That is soo-oo marginal case that I refuse to take it into account.
For me the best solution would be to add explicit statement in FAQ: _Nobody should expect that the macros in Addons will be kept same forever. They can change their names, placement, and functions. As well as be added and removed. It is not macro repository. If you need some particular macro - just copy it to scripts (and don't forget to backup this and other part of profile/config. But better learn to make simple key reassignments yourself, it is as simple as [short code]_
But .. I see that Fargroup members have other opinion, so all that I can - try to discuss it.

> Oh. Not too dramatic changes hopefully?

Is it ever important? As I see - nobody cares, so whatever.
(As I can remember nothing important, but you see - I didn't expect objections to Ins/Space as well)

> I believe in this particular case there was a feedback. And, since you didn't object back then, we supposedly reached a concensus.

Well.. I worked on this macro set since 2013 (or even earlier), your objection is from 2015, and final PR dated 2021, so it is just too long time to really remember (and take into account) all minor details.
During 2013-2021 I changed macros many times, forth and back (sometimes). I can admit that my own vision have changed since then.

alabuzhev:
> From my point of view Addons/Macros is far from being macro repository.

Definitely. I see it as:
- Workarounds for removed functionality, e.g. the one being discussed.
- Bindings to behave more like other apps.
- A few actually useful examples.

> I still doubt that someone in the world really do this

I did exactly that more than once 🤷

> add more useful macro sample, and remove trivial reassigning. To keep macro set tidy and clean.

Fair enough. And including `panel.setpos(N+1)` probably would've make it an even more useful example (and also backwards-compatible, avoiding this lengthy discussion altogether :) )

> That is soo-oo marginal case that I refuse to take it into account.

I invented it of course. Having said that, I wouldn't bet my money that it isn't happening somewhere.

> add explicit statement in FAQ

Yes, and to the readme in Addons.

> As I see - nobody cares, so whatever.

This spawned a few discussions already, so apparently people do care.

And please don't take it personally: it's natural to have different preferences and opinions, we can't always agree on everything, keeping the status quo is often simply the lesser evil.

johnd0e:
> I did exactly that more than once

And was it really easier than record macro with Ctrl[Shift]-Dot?
If yes, then ok.

> And including panel.setpos(N+1)

The actual point not moving cursor is:
- more convenience when selecting in opposite direction (walking from bottom to top)
- acting like Space in Total Commander
- and it is not so trivial as using Ctrl[Shift]-Dot

alabuzhev:
> And was it really easier than record macro with Ctrl[Shift]-Dot?

Of course. There are multiple areas. There are exceptions like non-empty command line. A semi-working macro is annoying.

> more convenience when selecting in opposite direction (walking from bottom to top)

I just use shift-arrows.

> So the whole discussion is about "Idea".

The idea is that you're very welcome to add a new Panel.SpaceToSelectTCLike macro if you want, there will be no objections.

johnd0e:
> Of course. There are multiple areas. There are exceptions like non-empty command line. A semi-working macro is annoying.

Ok then, I can admit that old macro still have sense.

MKadaner:
> ...explicit statement in FAQ: _Nobody should expect that the macros in Addons will be kept same forever..._

My 2¢: I've always perceived this set of macros as yet another "Options" dialog where I can enable few key assignments which for whatever reasons were not implemented natively. I use only half dozen, but I will be very upset if some of them disappear or change. Look, things like `Edit.SaveFile.lua`, `Panel.Del.lua`, or `Panel.Esc.lua` are completely uninspiring. They are just part of my overall Far experience which has been there for decades. And yes, sometimes when I set up a new machine, I use this macro _repository_ to enable my habitual key assignments instead of restoring my full profile from somewhere.

PR #485, opened 2021-11-29
new option for disable network scan

ctapmex:
## Summary
New option for disable network scan.

 
## References
https://forum.farmanager.com/viewtopic.php?p=169115#p169115

## Checklist


plugins/network/NetEng.hlf
```diff
@@ -44,30 +44,34 @@ $ #Network browser configuration#
  Enables adding "Network" item to the disks menu.
 
  #Add to plugins menu#
- Enables adding "Network" item to the plugins menu.
+ Enables adding "Network browser" item to the plugins menu.
 
- #Выводить ".." в списке провайдеров#
- Показывать ".." на панели в списке провайдеров.
+ #Show ".." in the providers#
```
alabuzhev, line 49: `the providers list` or `the list of providers`

plugins/network/NetEng.hlf
```diff
@@ -44,30 +44,34 @@ $ #Network browser configuration#
...
 workgroup). The option works only for Microsoft Windows networks.
 
- #Показывать принтеры#
- Показывать предоставляемые хостом общие принтеры.
+ #Show printers#
+ Show shared printers.
 
- #Показывать полные пути к общим ресурсам#
- ...
+ #Show share full name#
```
alabuzhev, line 60: Probably `full share name`

plugins/network/NetEng.hlf
```diff
@@ -44,30 +44,34 @@ $ #Network browser configuration#
...
+ When the option is enabled, \\Server1 and \\Server1\Share1 are displayed, when disabled - Server1 and Share1
 
  #Show hidden shares#
- If this option is enabled, hidden (administrative) shares are shown in the plugin.
+ Опция управляет показом скрытых (административных) общих ресурсов компьютера.
```
alabuzhev, line 64: ?

plugins/network/NetEng.hlf
```diff
@@ -44,30 +44,34 @@ $ #Network browser configuration#
...
 
- #Показывать скрытые ресурсы атрибутом Hidden#
- Если включена эта опция, плагин для скрытых административных ресурсов ставит атрибут Hidden.
+ - #Never show#
+   There is no search for hidden shares.
```
alabuzhev, line 67: *"There is no spoon"* :)
`Don't show hidden shares` or `hidden shares are never shown` or something like that might sound more natural.

plugins/network/NetEng.hlf
```diff
@@ -44,30 +44,34 @@ $ #Network browser configuration#
  Enables adding "Network" item to the disks menu.
 
  #Add to plugins menu#
- Enables adding "Network" item to the plugins menu.
+ Enables adding "Network browser" item to the plugins menu.
```
alabuzhev, line 47: The English hlf is the baseline for all the other translations.
It's easier to spot the untranslated bits by comparing with it.
If you change anything in it, please update all the other hlfs as well, it's a pain when they go out of sync.

plugins/network/NetCfg.cpp
```diff
@@ -52,16 +53,17 @@ int Config()
...
+			Opt.HiddenSharesAsHidden = false;
 			break;
 		case 2: // always show
-			Opt.HiddenShares = TRUE;
-			Opt.HiddenSharesAsHidden = FALSE;
+			Opt.HiddenShares = true;
+			Opt.HiddenSharesAsHidden = false;
+			break;
+		case 1: // make hidden
+		default:
```
alabuzhev, line 64: Why do we need this change?
The value comes from a dialog, we won't suddenly get `42` here.

ctapmex: >we won't suddenly get 42 here.
sure ? ;) ok, even if this happens, then the cost of error here is small

plugins/network/NetEng.hlf
ctapmex: I took these lines (and below) from the NetEng.lng. But ok, once we started, let's put it in order.

plugins/network/NetEng.hlf
BestiaPL: Please add the "Scan network" topic to help file - now it is not described.

ctapmex: @BestiaPL  in the next patch

PR #491, opened 2021-12-04
Mantis#0003921 command 'cd' in network panel

ctapmex:
## Summary
Fix Mantis#0003921

 
## References
https://bugs.farmanager.com/view.php?id=3921

## Checklist


plugins/network/NetClass.cpp
```diff
@@ -476,10 +476,15 @@ int NetBrowser::ProcessEvent(intptr_t Event, void* Param)
 	}
 	if (Event == FE_COMMAND)
 	{
-		SetOpenFromCommandLine(static_cast<wchar_t*>(Param));
-		PsInfo.PanelControl(this, FCTL_SETCMDLINE, 0, (void*)"");
-		PsInfo.PanelControl(this, FCTL_UPDATEPANEL, 0, nullptr);
-		return TRUE;
+		auto command = static_cast<wchar_t*>(Param);
+		if (wcsncmp(command, L"net:", 4) == 0 || wcsncmp(command, L"netg:", 5) == 0)
```
alabuzhev, line 480: SetOpenFromCommandLine already checks this and returns false if there's no prefix.
Could we just return false here if SetOpenFromCommandLine returned false?

ctapmex: ok

PR #497, opened 2021-12-08
refactoring

ctapmex:
## Summary
Refactoring to avoid using raw pointers and old data types.

 
## Checklist


plugins/network/NetClass.cpp
```diff
@@ -51,110 +49,54 @@ void NetBrowser::LogData(const wchar_t * Data)
...
-	free(Dest.lpRemoteName);
-	free(Dest.lpLocalName);
-	free(Dest.lpComment);
-	free(Dest.lpProvider);
-	memcpy(&Dest, &Src, sizeof(NETRESOURCE));
-	Dest.lpRemoteName = CopyText(Src.lpRemoteName);
-	Dest.lpLocalName = CopyText(Src.lpLocalName);
-	Dest.lpComment = CopyText(Src.lpComment);
-	Dest.lpProvider = CopyText(Src.lpProvider);
+	ResList.emplace_back(Res);
```
alabuzhev, line 59: Res should probably be either `const NetResource&` or `NetResource&&` (and moved).

plugins/network/NetClass.cpp
```diff
@@ -167,21 +109,18 @@ BOOL NetResourceList::Enumerate(
...
 		}
 
 		if (NetCount > 0)
 		{
-			ResList = (NETRESOURCE*)realloc(ResList, (ResCount + NetCount) * sizeof(*ResList));
-			ZeroMemory(&ResList [ResCount], sizeof(*ResList)*NetCount);
-
-			for (unsigned I = 0; I < NetCount; I++)
-				CopyNetResource(ResList[ResCount + I], nr[I]);
+			ResList.reserve(ResList.size() + NetCount);
```
alabuzhev, line 120: Exact reserve in a loop is usually an antipattern - instead of growing exponentially we're reallocating on every iteration, which kills the amortized constant complexity.

plugins/network/NetClass.cpp
```diff
@@ -167,21 +109,18 @@ BOOL NetResourceList::Enumerate(
...
 
-			ResCount += NetCount;
+			for (size_t i = 0; i < NetCount; i++)
+				ResList.emplace_back(NetResource(nr[i]));
```
alabuzhev, line 123: Is NetResource required here? emplace_back often can accept just arguments.

plugins/network/NetClass.cpp
```diff
@@ -350,20 +287,6 @@ int NetBrowser::GetFindData(PluginPanelItem** pPanelItem, size_t* pItemsNumber,
 
 	if (ChangeDirSuccess)
 	{
-		if (CmdLinePath[0])
```
alabuzhev, line 353: Not needed anymore?

plugins/network/NetClass.cpp
```diff
@@ -476,13 +397,10 @@ int NetBrowser::ProcessEvent(intptr_t Event, void* Param)
...
 	if (Event == FE_COMMAND)
 	{
-		if (SetOpenFromCommandLine(static_cast<wchar_t*>(Param)))
-		{
-			PsInfo.PanelControl(this, FCTL_SETCMDLINE, 0, (void*)"");
-			PsInfo.PanelControl(this, FCTL_UPDATEPANEL, 0, nullptr);
-			return TRUE;
-		}
-		return FALSE;
+		SetOpenFromCommandLine(static_cast<wchar_t*>(Param));
```
alabuzhev, line 400: Do we now *always* handle the command line? Why?

plugins/network/NetClass.cpp
```diff
@@ -566,39 +484,39 @@ BOOL NetBrowser::CancelConnection(const wchar_t* RemoteName)
...
+bool NetBrowser::GetDriveToDisconnect(const wchar_t* RemoteName, wchar_t* LocalName)
 {
 	wchar_t LocalNames[MAX_PATH][10];
 	DWORD LocalNameCount = 0;
 	DWORD i;
 
 	for (i = 0; i < ConnectedList.Count(); i++)
 	{
-		NETRESOURCE& connRes = ConnectedList[i];
+		NetResource& connRes = ConnectedList[i];
```
alabuzhev, line 511: auto&?

plugins/network/NetClass.cpp
```diff
@@ -1015,34 +931,34 @@ BOOL NetBrowser::ChangeToDirectory(const wchar_t* Dir, OPERATION_MODES opmodes,
...
 				NetList[I].lpRemoteName)
 			{
 				if (IsFind)
-					return (FALSE);
+					return false;
 
-				wchar_t NewDir[MAX_PATH], LocalName[MAX_PATH];
-				GetLocalName(NetList[I].lpRemoteName, LocalName);
+				string NewDir;
+				string LocalName = GetLocalName(NetList[I].lpRemoteName.get());
```
alabuzhev, line 955: Do we modify LocalName anywhere? If no, can it be a const ref?

plugins/network/NetClass.cpp
```diff
@@ -2301,24 +2228,24 @@ int NetBrowser::GotoComputer(const wchar_t* Dir)
...
 		{
-			IsShare = TRUE;
+			IsShare = true;
 			*p = L'\0';
 		}
 	}
 
 	CharUpper(ComputerName);
-	NETRESOURCE res{};
+	NetResource res{};
```
alabuzhev, line 2237: {} is probably redundant, same below.

plugins/network/NetClass.hpp
```diff
@@ -2,37 +2,110 @@
 #define __NETCLASS_HPP__
 
 #include <plugin.hpp>
+#include <memory>
```
alabuzhev, line 5: plugin.hpp includes windows.h, it's probably better to include it after std stuff just in case (macro conflicts etc.)

plugins/network/NetClass.hpp
```diff
@@ -2,37 +2,110 @@
...
+using string = std::wstring;
+
+class NetResource
 {
+public:
+	DWORD dwScope{};
+	DWORD dwType{};
+	DWORD dwDisplayType{};
+	DWORD dwUsage{};
+	std::unique_ptr<string> lpLocalName{};
```
alabuzhev, line 22: Why pointers to strings and not just strings?
Especially if this class is copyable?

plugins/network/NetClass.hpp
```diff
@@ -2,37 +2,110 @@
...
+	NetResource() = default;
+
+	NetResource(const NetResource& other)
+	{
+		copy(other);
+	}
+
+	NetResource& operator=(const NetResource& other)
+	{
+		if (this == &other)
```
alabuzhev, line 50: This is usually redundant. Self-copy is a logic error anyway.

plugins/network/NetClass.hpp
```diff
@@ -2,37 +2,110 @@
...
+	}
 
 public:
-	NetResourceList();
-	~NetResourceList();
+	NetResource() = default;
+
+	NetResource(const NetResource& other)
+	{
+		copy(other);
```
alabuzhev, line 45: Instead of having a private copy function it can be delegated to the operator, e.g. `*this = other;`

plugins/network/NetClass.hpp
```diff
@@ -48,45 +121,44 @@ class NetBrowser
...
 	NetResourceList NetList; // list of resources in the current folder
-	wchar_t NetListRemoteName[MAX_PATH]; // remote name of the resource stored in NetList
+	string NetListRemoteName; // remote name of the resource stored in NetList
 	NetResourceList ConnectedList; // list of resources mapped to local drives
 	NetResourceList RootResources; // stack of resources above the current level
 	// (used in non-MS Windows networks only)
-	NETRESOURCE CurResource = {}; // NETRESOURCE describing the current location
-	NETRESOURCE* PCurResource; // points to CurResource or nullptr (if at root)
+	NetResource CurResource; // NETRESOURCE describing the current location
+	NetResource* PCurResource; // points to CurResource or nullptr (if at root)
```
alabuzhev, line 129: Initialise it maybe?

plugins/network/changelog
```diff
@@ -1,3 +1,8 @@
+ctapmex 08.12.2021 18:00:00 +0000 - build 106
+
+1. продолжение 105.1 - приактивной панели плагина в командной строке обрабатываются только команды формата net:* или netg:*
```
alabuzhev, line 3: Typo in `приактивной`

plugins/network/NetClass.cpp
ctapmex: I don't have a cycle here
```
if (NetCount > 0)
		{
			ResList.reserve(ResList.size() + NetCount);

			for (size_t i = 0; i < NetCount; i++)
				ResList.emplace_back(NetResource(nr[i]));
		}
```

plugins/network/NetClass.cpp
ctapmex: https://bugs.farmanager.com/view.php?id=3921
"при активной панели плагина в командной строке обрабатываются только команды формата net:* или netg:*"
т.е., мы всегда должны обработать командную строку, не отдавая её Far. Иначе он во многих случаях выдаст ошибку из-за невозможности выставить CurDir

plugins/network/NetClass.hpp
ctapmex: because the string may not be specified. on the other hand, the code contains checks for an empty string in the value, and we do not need to differ between an empty string and an unset string.

plugins/network/NetClass.cpp
alabuzhev: This whole thing is in one big `for(;;)`, isn't it?

plugins/network/NetClass.hpp
alabuzhev: `std::optional` is way more suitable for optional things than `unique_ptr` (as the name suggests), at least because it's copyable and has no heap overhead. And indeed, an empty string could do as well.

plugins/network/NetClass.cpp
alabuzhev: I'm afraid this would cause way more confusion and bug reports than occasional curdir errors with relative paths.

plugins/network/NetClass.cpp
ctapmex: chang vector to list

plugins/network/NetClass.cpp
ctapmex: rollback

plugins/network/NetClass.cpp
ctapmex: yes, don`t need

plugins/network/NetClass.cpp
```diff
@@ -384,46 +305,44 @@ int NetBrowser::GetFindData(PluginPanelItem** pPanelItem, size_t* pItemsNumber,
...
 	{
 		ReenterGetFindData--;
-		return (FALSE);
+		return FALSE;
 	}
 
 	int CurItemPos = 0;
 
-	for (unsigned I = 0; I < NetList.Count(); I++)
+	for (auto list: NetList.ResList)
```
alabuzhev, line 320: const ref maybe?

plugins/network/NetClass.cpp
```diff
@@ -1015,34 +929,34 @@ BOOL NetBrowser::ChangeToDirectory(const wchar_t* Dir, OPERATION_MODES opmodes,
 	if (AnsiDir[1] == L'/')
 		AnsiDir[1] = L'\\';
 
-	for (unsigned I = 0; I < NetList.Count(); I++)
+	for (auto list: NetList.ResList)
```
alabuzhev, line 932: const ref maybe?

plugins/network/NetClass.cpp
```diff
@@ -2137,67 +2053,77 @@ void NetBrowser::ManualConnect()
...
+			PsInfo.PanelControl(this, FCTL_GETPANELITEM, static_cast<intptr_t>(PInfo.CurrentItem), &gpi);
 			ChangeToDirectory(PPI->FileName, OPM_NONE, true);
 			free(PPI);
 		}
 	}
 }
 
 
-void NetBrowser::GetRemoteName(NETRESOURCE* NetRes, wchar_t* RemoteName)
+string NetBrowser::GetRemoteName(NetResource* NetRes)
```
alabuzhev, line 2070: If NetRes can't be null here, maybe accept by a const ref?

plugins/network/NetClass.hpp
```diff
@@ -1,38 +1,86 @@
...
 {
-private:
-	NETRESOURCE* ResList;
-	unsigned int ResCount;
+public:
+	DWORD dwScope{};
+	DWORD dwType{};
+	DWORD dwDisplayType{};
+	DWORD dwUsage{};
+	string lpLocalName{};
```
alabuzhev, line 22: No need to explicitly initialise strings.

plugins/network/NetClass.hpp
```diff
@@ -1,38 +1,86 @@
...
+	string lpRemoteName{};
+	string lpComment{};
+	string lpProvider{};
 
 public:
-	NetResourceList();
-	~NetResourceList();
+	NetResource() = default;
+
+	NetResource(const NetResource& other) = default;
```
alabuzhev, line 30: If all 4 are default, no need to declare them at all.

plugins/network/NetClass.hpp
```diff
@@ -1,38 +1,86 @@
...
+		dwType = other.dwType;
+		dwDisplayType = other.dwDisplayType;
+		dwUsage = other.dwUsage;
+		lpLocalName = other.lpLocalName? other.lpLocalName : L"";
+		lpRemoteName = other.lpRemoteName? other.lpRemoteName : L"";
+		lpComment = other.lpComment? other.lpComment : L"";
+		lpProvider = other.lpProvider? other.lpProvider : L"";
+	}
+
+	static NETRESOURCE getNETRESOURCE(const NetResource* resource)
```
alabuzhev, line 47: Why a static function? It needs a class instance anyway.

plugins/network/NetClass.cpp
ctapmex: no.  in line 326 not const function

plugins/network/NetClass.cpp
ctapmex: no. in line 934 not const function

plugins/network/NetClass.cpp
ctapmex: ref, but not const

plugins/network/NetClass.hpp
ctapmex: in NetResourceList::Enumerate  NetResource maybe is null . In this case getNETRESOURCE return empty NETRESOURCE

alabuzhev: It's kinda fishy that we then pass that we then pass that empty NETRESOURCE to [WNetOpenEnum](https://docs.microsoft.com/en-us/windows/win32/api/winnetwk/nf-winnetwk-wnetopenenumw), which expects either a null pointer or a properly filled structure.

In the current version we pass the ptr as is, without turning nullptr to an empty struct:
https://github.com/FarGroup/FarManager/blob/f20c908cfe7746c4f9d425c3548f18a3b1b82a42/plugins/network/NetClass.cpp#L154

Are you sure you want an empty struct there?

plugins/network/NetClass.cpp
alabuzhev: list is almost always way worse than vector, unless you have specific requirements like no iterators invalidation.

vector is totally fine here, just please remove the explicit `reserve` and let it grow automatically.

plugins/network/NetClass.hpp
ctapmex: fixed

alabuzhev:
Looks good, please rebase & squash redundant commits.

PR #523, opened 2022-04-25
Code refactor and C++ performance features

GermanAizek:
## Summary

Fixed most of the parameters they are now const references, using `std::move()` in explicit places.
`noexcept` simplifying code generation by compiler in constructors and operators.

More info in title commits.


far/config.cpp
```diff
@@ -1230,7 +1230,7 @@ void Options::SetFilePanelModes()
 
 		if (DeleteMode)
 		{
-			const auto SwitchToAnotherMode = [&](panel_ptr p)
+			const auto SwitchToAnotherMode = [&](const panel_ptr& p)
```
alabuzhev, line 1233: We pass shared pointers by value for a reason.
Window handling in Far is somewhat complicated and certain operations can remove individual windows from their container, thus destroying the corresponding pointers if they're passed by a const ref.
On the contrary, when we pass by value, it is guaranteed that the pointer will stay alive until the end of the function.

A simplified example:
https://godbolt.org/z/ej3rzda3P

Unless you're absolutely sure that the function and all the other functions it may call won't invalidate the passed pointer and *it will be the case forever*, it's better to leave it as is.

far/filelist.cpp
```diff
@@ -1953,7 +1953,7 @@ bool FileList::ProcessKey(const Manager::Key& Key)
 						if (PluginMode && UploadFile)
 						{
 							PluginPanelItemHolderHeap PanelItem;
-							const auto strSaveDir = os::fs::get_current_directory();
+							const auto& strSaveDir = os::fs::get_current_directory();
```
alabuzhev, line 1956: This [doesn't really change anything](https://godbolt.org/z/KhTr7qGr5).

far/copy.cpp
```diff
@@ -797,7 +797,7 @@ ShellCopy::ShellCopy(
 			msg(Link? lng::MCMLTargetIN : lng::MCMLTargetTO));
 	}
 
-	CopyDlg[ID_SC_TARGETTITLE].strData=strCopyStr;
+	CopyDlg[ID_SC_TARGETTITLE].strData = std::move(strCopyStr);
```
alabuzhev, line 800: It is a valid optimisation, but:
- It makes the code harder to support: now it relies on an *assumption* (not a statically proven fact) that `strCopyStr` is not used anywhere below this line. It could be the case right now, but tomorrow someone could use `strCopyStr` for something else and it might even work for some time due to SSO and then suddenly break.
- It's not a hot path, just showing the copy dialog *once*, so we won't see any effect of it whatsoever.

far/common/preprocessor.hpp
```diff
@@ -124,13 +124,13 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 	NOT_COPY_ASSIGNABLE(Type)
 
 #define MOVE_CONSTRUCTIBLE(Type) \
-	Type(Type&&) = default
+	Type(Type&&) noexcept = default
```
alabuzhev, line 127: [Please don't](https://stackoverflow.com/questions/18653726/is-the-default-move-constructor-defined-as-noexcept).

far/common/preprocessor.hpp
```diff
@@ -124,13 +124,13 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 
 #define NOT_MOVE_CONSTRUCTIBLE(Type) \
 	Type(Type&&) = delete
 
 #define MOVE_ASSIGNABLE(Type) \
-	Type& operator=(Type&&) = default
+	Type& operator=(Type&&) noexcept = default
```
alabuzhev, line 133: Ditto.

alabuzhev:
Static analysers are a good thing, but please don't follow their suggestions blindly: quite often they can be either pointless or even make things worse.

GermanAizek:
@alabuzhev, accepted for information.

far/filelist.cpp
MKadaner: Hmm... If I am not mistaken, `os::fs::get_current_directory` is [declared](https://github.com/FarGroup/FarManager/blob/33df9f5537981b4b3e33bbaf7d7c01772aec2475/far/platform.fs.hpp#L496) as
```
const string& get_current_directory();
```
If so, declaring reference here [makes whole lot of difference](https://godbolt.org/z/eqb4WbTed).

GermanAizek: @MKadaner,
I agree with you, my IDE when hovering over `const auto` returns that the type will be `const string` without reference.
Tested on Qt Creator 7.0.0 and VS 2019

far/copy.cpp
GermanAizek: @alabuzhev, clearly this may prevent further code modification, but the IDE returns a warning when reused after `std::move()`. Most developers disable compiler alerts, so I agree that this will make it difficult to change the code, but when the code is final, you will have to use `std::move()`.

alabuzhev: > IDE returns a warning when reused after `std::move()`

That would've been great, but, [unfortunately, no](https://godbolt.org/z/P4zzbPK1W).

far/filelist.cpp
alabuzhev: @MKadaner, yes, you're right.

However, then it's another argument against changing this: as the name of the variable suggests, we save the "current directory" value here, apparently because the code below sets it to something else, and afterwards we restore it from the saved copy. If we change it to a reference, it will no longer "save" anything and the logic will be broken.

MKadaner: @alabuzhev Oh! That's for sure. My bad; did not look further down. :(

PR #531, opened 2022-06-21
Allow run modal editor with nonfatal exit codes in apiEditor #530

EugeneManushkin:
## Summary
Allow run modal editor with nonfatal exit codes in apiEditor. This fix invalid EF_OPENMODE_NEWIFOPEN flag behaviour.

Since all fatal exit codes (XC_OPEN_ERROR, XC_LOADING_INTERRUPTED and XC_EXISTS) are handled in case statements above, there is no need to check on errors before Global->WindowManager->ExecuteModal(Editor) call.

 
## References
#530 

## Checklist


far/plugapi.cpp
```diff
@@ -1957,8 +1957,7 @@ intptr_t WINAPI apiEditor(const wchar_t* FileName, const wchar_t* Title, intptr_
 					/* $ 15.05.2002 SKV
 					  Зафиксируем вход и выход в/из модального редактора.
 					  */
-					if (-1 == editorExitCode)
-						Global->WindowManager->ExecuteModal(Editor);
+					Global->WindowManager->ExecuteModal(Editor);
```
alabuzhev, line 1960: According to 384e51ec4f6e8eaa038dc286fe7b388f398fee42, the intention was to fix a crash, so I'm not quite comfortable with just reverting it like this, and tbh have neither time nor inspiration to investigate whether it's safe to do so.

Looks like in this case the exit code is `XC_OPEN_NEWINSTANCE`, maybe we can do something like this instead?
```diff
- if (-1 == editorExitCode)
+ if (any_of(editorExitCode, -1, XC_OPEN_NEWINSTANCE))
```
It should address your case and preserve the fix from 384e51ec4f6e8eaa038dc286fe7b388f398fee42, whatever it was about.
If yes, could we do that in all the affected places (filelist.cpp, findfile.cpp, usermenu.cpp) for consistency?

EugeneManushkin: @alabuzhev, thank you for your advice. Done.

PR #539, opened 2022-08-07
Fixed two bugs in the string search algorithm.

MKadaner:
## Summary
Fixed two bugs in `SearchAndReplaceString` function when searching for whole words.

## Checklist

## Details
Fixing two bugs in the search algorithm when in the "whole words" mode:
1. Forward whole words search for a pattern containing repeated substrings separated by a word delimiter character does not find the occurrence overlapping with the sequence disqualified because it is a part of a larger word.
   - For example, the whole word search for `abc.abc` in the line `$abc.abc.abc` starting from the beginning of the line does not find the pattern at the end of the line.
2. Reverse whole words search gets into an infinite loop if the sequence disqualified because it is a part of a larger word happens to be closer to the beginning of the line than the size of the search pattern.
   - For example, reverse whole words search for `abc` in the line `$abc` starting from the end of the line ends up in infinite loop.


far/strmix.cpp
```diff
@@ -923,8 +923,8 @@ bool SearchAndReplaceString(
 	const auto Next = [&](size_t const Offset, size_t const Size)
```
alabuzhev, line 923: - Looks like we do not need `Size` here anymore, maybe remove it?
- There is [a very similar piece](https://github.com/FarGroup/FarManager/blob/a6c59d3c6af7ce340013c2449e047a487ca467c9/far/findfile.cpp#L1225-L1228) in findfile.cpp, it probably needs the same fix.

MKadaner: - Doh!
- Fixed findfile.cpp

alabuzhev:
Thank you.

PR #554, opened 2022-10-13
Borland paste fix

JouriM66:
## Summary

Workaround for clipboard data pasted by old Borland IDE`s.

## Checklist

## Details

If clipboard data in UNICODE and was created by old Borland products they all have same bug: create unicode-like data in clipboard but actually its simple text in ACP codepage "expanded" to unicode by zeroes like this:
```
"text"
```
will be:
```
"t\x00e\x00x\x00t\x00\x00\x00"
```
This works as UNICODE for ansi chars but failed with any international multibyte codepages like win1251.

Solution:
 - compress original text back to MB
 - convert it to correct unicode replacing data read from clipboard

!!NOTE:

This code assumes what ANY application which put data to clipboard using borland-clipboard-flag has this bug.

All versions of old Borland products I know (Builder, Delphi) has it but may be some exceptions exists,

### DIFF
--- a/far/clipboard.cpp
+++ b/far/clipboard.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 clipboard.cpp
 
 Работа с буфером обмена.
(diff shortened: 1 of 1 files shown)


alabuzhev:
This has nothing to do with Borland products in particular.
All apps that puts CF_TEXT and not CF_UNICODETEXT to clipboard (basically all those ancient non-Unicode apps from Win 9x times) are affected:

https://learn.microsoft.com/en-us/windows/win32/dataxchg/standard-clipboard-formats#CF_LOCALE
> When you close the clipboard, if it contains CF_TEXT data but no CF_LOCALE data, the system automatically sets the CF_LOCALE format to the current input language.
> The system uses the code page associated with CF_LOCALE to implicitly convert from CF_TEXT to CF_UNICODETEXT.

The issue is known for over a decade and the workaround is trivial - switch to the corresponding input language before copying.

The proper way to address this is described [on the same page](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enumclipboardformats#remarks):
> If you are pasting information from the clipboard, retrieve the first clipboard format that you can handle. That will be the most descriptive clipboard format that you can handle.
> The system provides automatic type conversions for certain clipboard formats. In the case of such a format, this function enumerates the specified format, then enumerates the formats to which it can be converted.

I.e. we should not trust CF_UNICODETEXT and take it blindly, but enumerate the available formats, pick CF_TEXT *if it comes before* and convert it to Unicode ourselves, using CP_ACP.

Personally I'm fine with this approach, but some disagree. Last time we discussed it (circa 2016) the argument was that with the current implementation it is possible to copy text as *non-ACP* from a legacy app by switching to a non-ACP input language and sometimes it is desirable.

JouriM66:
Thanks for the info.

Now decode block looks like this:
```C++
		if (IsOldBorlandText())
		{
		  const auto CheckAnsiOnly = [&](auto src, int maxS) -> int
		  {
			int sz = 0;
			for (sz = 0; sz < maxS && *src; src++, sz++)
			  if ((*src & 0xFF00) != 0)
				return 0;
			return sz;
		  };
		  int sz = CheckAnsiOnly( Data.c_str(), Data.length() );
		  if (sz)
		  {
			auto buff = std::make_unique<char[]>(Data.length());
			auto src = Data.c_str();
			auto b = buff.get();
			for ( int n = 0; n < sz; b++, src++, n++ )
			  *b = *src & 0xFF;
			MultiByteToWideChar(CP_ACP, 0, buff.get(), sz, (LPWSTR)Data.c_str(), sz);
		  }
```

Works in both input languages now,
For me its very annoying to remember to switch languages before copy, so I will use this fix.

alabuzhev:
I agree that it can be annoying enough and a workaround is needed, and not only for Borland products.
6041 / 0c990fa should hopefully cover all such cases.

However, those ANSI apps are too rare these days to enable it by default and make everyone pay the price for extra conversions.
Set ClipboardUnicodeWorkaround to true in far:config to enable it.

JouriM66:
I checked your commit, its fully working, thank you.

WitMakishori:
@JouriM66 hey, please, can you aswer me in mail? i sent one to you, please, if its possible, from witmakishori@gmail.com
(sorry for offtopic, its really important for me and i didnt found a way to DM on github, only way to get email by github api events)

PR #561, opened 2022-11-09
Updated Polish language

BestiaPL:
## Summary
Updated translation up to latest vesrion.

 
## Checklist


far/farlang.templ.m4
```diff
@@ -13027,7 +13027,7 @@ MInfoMemoryCommittable
 "Committable"
 "Committable"
 "Committable"
-"Committable"
+"Zarezerwowana"
```
alabuzhev, line 13030: Just for the context:

"Committable" here is the amount of memory that is, in theory, available for a native process, e.g. a sum of the physical memory and page files (if any). 
Another possible word is "allocable".

My Polish is extremely basic, but "Zarezerwowana" sounds like something that is "*already* reserved" rather than "*can be* reserved".

Maybe "committable" is not the best term here, I'm not even sure how to properly translate it to the languages I know.
Maybe we should change it to "available" instead, and "total/available" to "total/free".

PR #566, opened 2022-12-06
Update Belarusian localization

lidacity:
## Summary
update Belarusian language-files

## Checklist
- [*] I have followed the [contributing guidelines](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTING.md).

## Details
update Belarusian language in main farlang.templ.m4
update Belarusian language in plugins *_bel.hlf and *_bel.lng files


plugins/arclite/arclite_bel.hlf
```diff
@@ -0,0 +1,270 @@
+﻿.Language=Belarussian,Belarussian (Беларуская)
```
alabuzhev, line 1: This doesn't really look like a Belarusian localization.

plugins/macroview/MacroBel.hlf
```diff
@@ -0,0 +1,518 @@
+﻿.Language=Belarusian,Belarusian (Беларуская)
```
alabuzhev, line 1: This doesn't really look like a Belarusian localization.

Note: this plugin does not work with the current Far version since 2012 or so and it is highly unlikely that it ever will, so I suggest skipping it (unless you have already translated it of course).

plugins/helloworld/HelloWorld_bel.hlf
```diff
@@ -0,0 +1,22 @@
+﻿.Language=Russian,Russian (Русский)
```
alabuzhev, line 1: Wrong header

plugins/arclite/arclite_bel.hlf
alabuzhev, line 1: When you add new files, make sure that `.Language` is spelled exactly the same as in the [main template](https://github.com/FarGroup/FarManager/blob/cbd4fcf1f338f0e85b65148d45d970e34d5182c5/far/farlang.templ.m4#L23), i.e. with one `s` in this case, otherwise it won't be picked.

lidacity: і чым мая беларуская мова адрозніваецца ад той беларускай мовы, якую я дасылаў у папярэдні раз?

and how does my Belarusian language differ from the Belarusian language that I sent the previous time?

PR #568, opened 2022-12-07
Decoupling fuzzy search from case sensitive search. Part 1.

MKadaner:
## Summary
Combining case sensitive and fuzzy search options in a single tri-state checkbox in search dialog is inconvenient and limits the functionality.
- Ever since the introduction of fuzzy search option, my decades-old muscle memory became useless; `Alt+C` does not toggle `Case sensitive` checkbox anymore but does something more complex instead.
- Case sensitive fuzzy search sometimes may be useful. For example, `Uber` will match `Überwald` but not `überwald`.

In this PR, I propose fully decoupling these search options.

## Checklist

## Details
Decoupled case sensitive and fuzzy search in underlying search utility functions and `Search` / `Replace` dialog; will fix the `Find file` dialog in a later PR.


far/stddlg.hpp
```diff
@@ -121,10 +120,11 @@ int GetSearchReplaceString(
 	string& ReplaceStr,
 	string_view TextHistoryName,
 	string_view ReplaceHistoryName,
-	search_case_fold* pCaseFold,
+	bool* pCaseSensitive,
 	bool* pWholeWords,
 	bool* pReverse,
 	bool* pRegexp,
+	bool* pFuzzy,
```
alabuzhev, line 127: I like the idea of this PR, but not quite sure about reintroducing bools all over the place: functions with half a dozen boolean parameters is a pain to support.

Maybe instead:
- Add `fuzzy_icase` to the `search_case_fold` enum
- Add a helper function to convert UI checkboxes state into `search_case_fold`, e.g.:
```C++
search_case_fold ui_to_search_mode(bool IsIcase, bool IsFuzzy)
{
	return IsIcase?
		IsFuzzy? search_case_fold::fuzzy_icase : search_case_fold::icase :
		IsFuzzy? search_case_fold::fuzzy : search_case_fold::exact;
}
```
?

This would probably make the PR smaller as well.

far/FarEng.hlf.m4
```diff
@@ -3208,6 +3208,14 @@ $ #Editor: search/replace#
  Treat input as Perl regular expression (~search~@RegExp@ and ~replace~@RegExpRepl@).
 Each line is processed individually, so multi-line expressions and line break characters will not be found.
 
+ #Fuzzy search#
+ The search will be diacritical insensitive (for example, #ёлочка# will be found in #белочка#),
```
alabuzhev, line 3212: Nice example, but it probably won't be obvious to the target audience that the words "spruce" and "squirrel" are unrelated and the search is false positive in this context.
Maybe something like "ape"/"canapé", "he"/"cliché" [etc.](https://en-academic.com/dic.nsf/enwiki/3894487)?

far/strmix.hpp
```diff
@@ -188,36 +188,36 @@ bool IsCaseMixed(string_view Str);
 
 [[nodiscard]]
 bool SearchString(
-	string_view Haystack,
-	string_view Needle,
+	string_view const Haystack,
```
alabuzhev, line 191: https://stackoverflow.com/questions/38660537/why-is-const-unnecessary-in-function-declarations-in-header-files-for-parameters

far/stddlg.hpp
MKadaner: I think combining two (out of six) independent Boolean parameters in an enumeration, even if it covers all possible combinations, is still cumbersome. I defined dedicated structs to hold all search options, which nicely enables the use of [Designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers). Please see if you like it.

far/FarEng.hlf.m4
MKadaner: Well, I did not mean to demonstrate a false positive case, just wanted something a marginally interesting in the example. I decided not to be a smart aleck and opted for "déjà vu" instead.

far/stddlg.cpp
```diff
@@ -202,17 +186,17 @@ int GetSearchReplaceString(
 		ReplaceDlg[dlg_button_selection].Flags |= DIF_HIDDEN;
 	}
 
-	if (!pCaseSensitive)
+	if (!Options.CaseSensitive.has_value())
```
MKadaner, line 189: Yes, I know, `std::optional::operator!` does exactly the same thing. However, with `std::optional<bool>` omitting `has_value()` may be confusing to a human reader. I will remove it if you like the terser notation.

MKadaner:
When approved, I will squash + rebase to master for the final merge.

far/strmix.hpp
alabuzhev: I used to do that too, but these days static analyzers often complain about useless qualifiers in such cases, which is annoying, but reasonable: they do not provide any extra information to the caller and can be easily confused with `const ref&` versions.

far/FarEng.hlf.m4
alabuzhev: I mean it is perfectly fine in FarRus.hlf, but probably less interesting in other languages.
Anyway, `déjà vu` is a great example too and probably universally understandable.

far/stddlg.cpp
alabuzhev: This is fine, it can be confusing indeed.

far/string_utils.hpp
```diff
@@ -180,6 +180,7 @@ class icase_searcher final: public i_searcher
 	exact_searcher m_Searcher;
 };
 
+template <typename Searcher>
```
alabuzhev, line 183: I'd prefer to not templatize the whole thing: fuzzy_searcher methods don't depend on the exact type of the underlying searcher and it would only duplicate the code. Maybe something like
<details>

```diff
diff --git a/far/string_utils.cpp b/far/string_utils.cpp
index 7a19e560c..b0e95a8ab 100644
--- a/far/string_utils.cpp
+++ b/far/string_utils.cpp
@@ -319,21 +319,19 @@ static void normalize_for_search(string_view const Str, string& Result, string&
 	Result.resize(End - Zip.begin());
 }
 
-static string normalize_for_search(string_view const Str, string& Intermediate, std::vector<WORD>& Types)
+fuzzy_searcher_impl::fuzzy_searcher_impl(i_searcher const* const Searcher):
+	m_Searcher(Searcher)
 {
-	string Result;
-	normalize_for_search(Str, Result, Intermediate, Types);
-	return Result;
 }
 
-template <typename Searcher>
-fuzzy_searcher<Searcher>::fuzzy_searcher(string_view const Needle, bool const CanReverse):
-	m_Searcher(normalize_for_search(Needle, m_Intermediate, m_Types), CanReverse)
+string fuzzy_searcher_impl::needle(string_view Needle) const
 {
+	string Result;
+	normalize_for_search(Needle, Result, m_Intermediate, m_Types);
+	return Result;
 }
 
-template <typename Searcher>
-std::optional<std::pair<size_t, size_t>> fuzzy_searcher<Searcher>::find_in(string_view const Haystack, bool const Reverse) const
+std::optional<std::pair<size_t, size_t>> fuzzy_searcher_impl::find_in(string_view const Haystack, bool const Reverse) const
 {
 	const auto Result = find_in_uncorrected(Haystack, Reverse);
 	if (!Result)
@@ -358,16 +356,12 @@ std::optional<std::pair<size_t, size_t>> fuzzy_searcher<Searcher>::find_in(strin
 	return Result;
 }
 
-template <typename Searcher>
-std::optional<std::pair<size_t, size_t>> fuzzy_searcher<Searcher>::find_in_uncorrected(string_view Haystack, bool Reverse) const
+std::optional<std::pair<size_t, size_t>> fuzzy_searcher_impl::find_in_uncorrected(string_view Haystack, bool Reverse) const
 {
 	normalize_for_search(Haystack, m_HayStack, m_Intermediate, m_Types);
-	return m_Searcher.find_in(m_HayStack, Reverse);
+	return m_Searcher->find_in(m_HayStack, Reverse);
 }
 
-template class fuzzy_searcher<exact_searcher>;
-template class fuzzy_searcher<icase_searcher>;
-
 #ifdef ENABLE_TESTS
 
 #include "testing.hpp"
diff --git a/far/string_utils.hpp b/far/string_utils.hpp
index 381b5e0f8..92a254d9a 100644
--- a/far/string_utils.hpp
+++ b/far/string_utils.hpp
@@ -180,20 +180,37 @@ private:
 	exact_searcher m_Searcher;
 };
 
-template <typename Searcher>
-class fuzzy_searcher final: public i_searcher
+class fuzzy_searcher_impl: public i_searcher
 {
-public:
-	NONCOPYABLE(fuzzy_searcher);
-	explicit fuzzy_searcher(string_view Needle, bool CanReverse = true);
+protected:
+	explicit fuzzy_searcher_impl(i_searcher const* Searcher);
 	std::optional<std::pair<size_t, size_t>> find_in(string_view Haystack, bool Reverse = {}) const override;
+	string needle(string_view Needle) const;
 
 private:
-	std::optional<std::pair<size_t, size_t>> find_in_uncorrected(string_view Haystack, bool Reverse = {}) const;
+	std::optional<std::pair<size_t, size_t>> find_in_uncorrected(string_view Haystack, bool Reverse) const;
 
 	mutable string m_HayStack, m_Intermediate;
 	mutable std::vector<WORD> m_Types;
 
+	i_searcher const* m_Searcher;
+};
+
+template <typename Searcher>
+class fuzzy_searcher final: public fuzzy_searcher_impl
+{
+public:
+	NONCOPYABLE(fuzzy_searcher);
+
+	explicit fuzzy_searcher(string_view const Needle, bool const CanReverse = true):
+		fuzzy_searcher_impl(&m_Searcher),
+		m_Searcher(needle(Needle), CanReverse)
+	{
+	}
+
+	using fuzzy_searcher_impl::find_in;
+
+private:
 	Searcher m_Searcher;
 };
 
```
</details>
?

MKadaner: Good point. No reason duplicating the entire `find_in` implementation. I did it slightly differently though, because:
1. Giving the base class constructor a pointer to a field of the derived class, while technically allowed, is a bad form.
2. Improved encapsulation of `fuzzy_searcher` inner workings.
3. Avoided redundant copying of the `Needle` while constructing `fuzzy_searcher`.

See if you like it.

far/stddlg.hpp
```diff
@@ -51,10 +51,19 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //----------------------------------------------------------------------------
 
 enum class lng : int;
-enum class search_case_fold;
 class regex_exception;
 struct error_state_ex;
 
+struct SearchReplaceDlgOptions
+{
+	std::optional<bool> CaseSensitive{};
```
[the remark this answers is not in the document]
MKadaner: Sure! No point explicitly default-initilize `std::optional`.

far/stddlg.hpp
MKadaner: All checks passed, I did not bother looking deeper; my bad. I applied the fix.

BTW, how to find gcc 8.1 build starting from PR? It looks like "continuous-integration/appveyor/pr" check points to Clang build. Do I miss something?

At a higher level:
> [E]ventually it all boils down to a flat list of searchers...

I was more concerned about `GetSearchReplaceString`. A function taking multiple positional parameters of the same type, especially sequentially, is a really bad thing. A dedicated struct combined with designated initializers is a nice way to imitate named parameters available in other languages. A bit funny-looking, but we are talking C++ after all.

Unrelated: Which build is official (published at https://www.farmanager.com/), MSVC, gcc, or Clang? Just curious.

alabuzhev: > how to find gcc 8.1 build starting from PR?

See the buttons in the "AppVeyor" column on the [main page](https://github.com/FarGroup/FarManager).

> points to Clang build

Yes, it looks like the GitHub check does not support multiple AppVeyor instances and gets overwritten by the last one of them unfortunately.

> Which build is official

MSVC 2019, usually slightly outdated (19.29.30136 currently). You can find that in `far:about`.

far/string_utils.hpp
alabuzhev: Looks good, thank you.

far/string_utils.hpp
```diff
@@ -180,28 +180,54 @@ class icase_searcher final: public i_searcher
...
+
+	private:
+		string m_Result, m_Intermediate;
+		std::vector<WORD> m_Types;
+	};
+}
+
+template <typename Searcher>
+#ifdef __cpp_concepts
+	requires std::is_constructible_v<Searcher, string_view, bool>
```
alabuzhev, line 199: Is it really needed? It doesn't look like we need to constraint the template here to solve some overload resolution issues or to help the users avoid confusion when instantiating it with arbitrary types on a daily basis.

If it is needed anyway, please see the `REQUIRES()` macro.

far/string_utils.cpp
```diff
@@ -279,61 +279,58 @@ std::optional<std::pair<size_t, size_t>> icase_searcher::find_in(string_view con
 	return m_Searcher.find_in(m_HayStack, Reverse);
 }
 
-static void normalize_for_search(string_view const Str, string& Result, string& Intermediate, std::vector<WORD>& Types)
+namespace
```
alabuzhev, line 282: I don't mind replacing static with an anonymous namespace, but it requires extra indentation and makes the diff larger and a bit harder to review.

If you're touching it anyway, maybe merge the whole thing into `fuzzy_searcher_impl::normalize`? It is only used there and in the test below.

far/string_utils.cpp
```diff
@@ -279,61 +279,58 @@ std::optional<std::pair<size_t, size_t>> icase_searcher::find_in(string_view con
...
-	zip const Zip(Result, Types);
-	const auto End = std::remove_if(ALL_RANGE(Zip), [](const auto& i)
-	{
-		return
-			!flags::check_any(std::get<1>(i), C3_ALPHA | C3_LEXICAL) &&
+		zip const Zip(Result, Types);
+		const auto End = std::remove_if(ALL_RANGE(Zip), [](const auto& i)
+			{
+				return
+				!flags::check_any(std::get<1>(i), C3_ALPHA | C3_LEXICAL) &&
```
alabuzhev, line 317: Looks misaligned now.

far/stddlg.hpp
MKadaner: Got it! Thank you.

far/string_utils.cpp
MKadaner: > If you're touching it anyway, maybe merge the whole thing into `fuzzy_searcher_impl::normalize`?

I had this at some point, then stumbled upon the test and thought that maybe the policy call for testing of such functions in strict isolation. I will happily apply this change.

far/stddlg.hpp
MKadaner: What's the point of supporting old compilers / language versions?

alabuzhev: > What's the point of supporting old compilers / language versions?

My personal opinion:
- There is no point. This is an end user project, not a library, a typical user takes prebuilt binaries and does not need to build it in some arcane environment, so we can just require the latest and be done with it.

Project maintainer opinion:
- We need to take into account that newest compilers often have more bugs than features
- There are multiple contributors to the project, supposedly not all of them have the latest or are able/willing to upgrade
- Someone has to reconfigure all the CI workflows to use newer compilers, possibly switching to newer VM images, possibly dealing with issues etc., it takes time and effort
- Since it requires some work, it should bring some visible benefits, e.g. at least make the code significantly simpler in multiple places, remove old workarounds etc.

Given the above, the usual primary targets are:
- MSVC: the **previous** major release (e.g. 2019) + the latest updates.
  If possible, we don't intentionally break the support for (previous - 1) and try to stay compatible with it for a while if people ask so. E.g. the last year it was still possible to compile Far in VS2017, but now it's broken and no one complains about it, so it's time to drop it for good.
  The current major release (e.g. 2022) is supported too of course.
  MSVC is used in production, i.e. it produces binaries published here and on farmanager.com
- GCC: the oldest that is more or less feature-compatible with the currently used MSVC.
  Not used in production, mostly for validation, cross checks, ensuring standard compatibility and extra warnings.
- Clang: the latest will do.
  Not used in production, mostly for validation, cross checks, ensuring standard compatibility and extra warnings.

far/string_utils.cpp
alabuzhev: > maybe the policy call for testing of such functions in strict isolation

Not really, more like "the test should work without initializing the rest of the codebase, e.g. config, language etc."
Sometimes it's easier to do so by extracting some code to free functions, but in this case the class is trivial and you've already added fuzzy_searcher_impl to the test body, so why not.

far/string_utils.cpp
MKadaner: Fixed.

far/string_utils.cpp
MKadaner: Done!

far/string_utils.hpp
MKadaner: I removed redundant `requires`.

far/stddlg.hpp
MKadaner: There are two different aspects here: eagerly chasing the latest and greatest compiler / language versions, and ruthlessly cutting off old versions as soon as migration to a new one is completed. As you pointed out, regular upgrading requires resources and should be done only when migration offers enough benefits. As much as I enjoy writing code in "modern C++" (whatever it is), I cannot commit to keeping the project at the technology cutting edge, so not asking for anything here.

On the other hand, supporting old versions is just a liability. I think the only reason to maintain compatibility with gcc 8.1 (especially since gcc 12.2.0 runs as part of the regular checks), is catering for contributors who are unable/unwilling to upgrade. (The overhead of build environment reconfiguring is only partially relevant, as most of the effort belongs to the upgrade side.) Frankly, I think that with the latest versions of all three compilers available for free and reasonably easy to install, it would be very beneficial to persuade such contributors to stop hindering the progress of software development industry. 😀

alabuzhev (approved):

MKadaner:
Squashed. Ready to merge. Thank you for the review.

alabuzhev:
Thank you!

MKadaner:
Thank you again. If there are issues related to this change reported outside of GH, please ping me directly. Unfortunately, I can miss them sometimes.

PR #584, opened 2022-12-30
Unifying Viewer and Editor search.

MKadaner:
## Summary
Unifying Viewer and Editor search.
Part 1: Added Fuzzy search to Viewer text search.

## Checklist
If not checked, I accept that this work might be rejected in favor of another great big ineffable plan.<br/>

### Done
- Added "Fuzzy" checkbox to the Viewer Search dialog and rearranged controls. How do you like it?
- Used `i_searcher` in text search mode.
- Minor refactoring.

### Undone
- Add Hex search controls and capabilities to `GetSearchReplaceString` dialog and use it in Viewer.
- Change layout of Find File dialog, put "Hex" search at the top.
- Use `i_searcher` in Viewer search in Regex and Hex mode.
- More refactoring of Viewer search code.
- Implement proper abstraction for mapping decoded text to encoded bytes in Viewer.


far/farlang.templ.m4
```diff
@@ -21226,32 +21226,32 @@ MViewSearchFor
...
-"Ricerca Di &Testo"
-"Шукати &текст"
-"Шукаць &тэкст"
-"Ieškoti &teksto"
+"&Текст"
+"&Text"
+"&Text"
+"&Text"
+upd:"&Szöveg"
+upd:"Szukaj &tekstu"
```
MKadaner, line 21234: @BestiaPL, please help with Polish translation here.

far/farlang.templ.m4
```diff
@@ -21226,32 +21226,32 @@ MViewSearchFor
...
-"Ricerca &Esadecimale"
-"Шукати 16-ричний &код"
-"Шука&ць hex"
-"Ieškoti &hex"
+"16-ричный &код"
+"&Hex"
+"He&x"
+"&Hex (xx xx …)"
+upd:"&Hexákra"
+upd:"Szukaj &wartości szesnastkowych"
```
MKadaner, line 21248: @BestiaPL, ... and here.

far/strmix.cpp
```diff
@@ -803,7 +803,7 @@ namespace
 	{
 		const auto BlankOrWordDiv = [&WordDiv](wchar_t Ch)
 		{
-			return std::iswblank(Ch) || contains(WordDiv, Ch);
+			return std::iswspace(Ch) || contains(WordDiv, Ch);
```
alabuzhev, line 806: As far as I remember, iswblank is used here because only space and \t were checked initially:

https://github.com/FarGroup/FarManager/blob/3f560a1d5f5b25f0795f92fef4d927fe839e2223/far/const.hpp#L49

Supposedly it was enough because \r, \n etc. are usually filtered out earlier.
iswspace looks reasonable if it is now used in Viewer, but it is probably worth checking other usages of iswblank just in case.

far/viewer.cpp
```diff
@@ -1007,11 +1011,20 @@ void Viewer::SetStatusMode(int Mode)
 	ShowStatusLine=Mode;
 }
 
+static constexpr std::array extra_word_divs{ encoding::bom_char, encoding::replace_char };
```
alabuzhev, line 1014: Come to think of it, should we consider replace_char non-word at all?
The domain of word characters is way larger than the domain of punctuation characters, so any encoding errors are likely to be within words.

far/viewer.cpp
```diff
@@ -1007,11 +1011,20 @@ void Viewer::SetStatusMode(int Mode)
...
 static bool is_word_div(const wchar_t ch)
 {
-	static const wchar_t extra_div[]{ encoding::bom_char, encoding::replace_char };
-	return !ch || std::iswspace(ch) || contains(Global->Opt->strWordDiv.Get(), ch) || contains(extra_div, ch);
+	return !ch || std::iswspace(ch) || contains(Global->Opt->strWordDiv.Get(), ch) || contains(extra_word_divs, ch);
+}
+
+static const string& get_word_div()
+{
+	static string word_div;
```
alabuzhev, line 1023: Should it really be static?

We rewrite it on every invocation anyway (which is correct - Opt->strWordDiv might change), so the only benefit is reusing the allocated storage.

alabuzhev:
> rearranged controls. How do you like it?

Looks good. Maybe do the the same in Find File for consistency?

far/viewer.cpp
MKadaner: Even though an arbitrary corrupted byte is more likely to fall on a word character, it feels that a false positive (found when should not) is more appropriate than false negative (not found when should) when interactively presenting results to a human being.

That said, I'd rather do not change the logic here in this PR. The risk of breaking something is high enough as it is. Let's make this change in a separate commit to facilitate rollback if necessary.

far/viewer.cpp
MKadaner: That was the reason. Why make an allocation when we can avoid it altogether with zero overhead? If this function would be defined in `string_utils` or such, where it could be called in unrelated scenario while the buffer is still in use, that would be a different story. As it is, I think a warning comment here is sufficient. Please let me know whether you prefer returning a newly allocated string every time or just add a comment.

MKadaner:
> Maybe do the the same in Find File for consistency?

Makes sense. Separate PR for sure. OK?

far/strmix.cpp
MKadaner: First, I am confused by your link. I could not find this definition of `isspace(x)` on master.

Second, I looked quickly through the existing usages of `iswblank`. Most places seem appropriate; a few are less obvious. Audit of the latter requires either better knowledge of the codebase or much more time to understand the expected semantics. Given that all existing places now work with `iswblank`, it seems reasonable to suppose that the change is not required anywhere else. I strongly suggest leaving everything as is and wait for user's complaints.

So far, I renamed `GetSpaces` to `GetBlanks` and added all-new `GetSpaces` to make the naming consistent with the standard.

far/farlang.templ.m4
BestiaPL: Can you put here screenshot of search dialog (even in English) and tell me how much characters can I use in strings.

BestiaPL, line 21234: &Tekst

far/farlang.templ.m4
BestiaPL: S&zesnastkowo

(I assume the text will fit in the window).

far/farlang.templ.m4
MKadaner: Thank you. Here you are!

ViewSearch Pol New
ViewSearch Eng New

If in the future we want to add the "Hex" option to the Search dialog in Editor, we may need to fit all this into 72 characters:
```
Search for  (•) Text  ( ) Hex  [ Word ] [ Selection ]
```

@alabuzhev, what do you think of changing radio button to checkbox here? It'll be easier to fit in one line in some languages.

BestiaPL: Please, try to set:
&Tekst
S&zesnastkowo

far/farlang.templ.m4
MKadaner: Yes, it will. Thank you!

far/farlang.templ.m4
MKadaner: Done! Thanks!

MKadaner: Done!

far/farlang.templ.m4
```diff
@@ -21281,6 +21281,20 @@ MViewSearchWholeWords
 "Толькі по&ўныя словы"
 "Visi ž&odžiai"
 
+MViewSearchFuzzy
+"Нестроги&й поиск"
+"Fu&zzy search"
+upd:"Fu&zzy search"
+upd:"Fu&zzy search"
+upd:"Fu&zzy search"
```
MKadaner, line 21289: @BestiaPL, oh, forgot to ask about "Fuzzy search". Translate it please.

far/viewer.cpp
alabuzhev: I'd rather not premature optimize object lifetimes unless necessary, but if it is called many times it is probably reasonable.

However, should it be called multiple times?
Looks like it is used in search_text_forward & search_text_backward, called by Search in a loop. 
While we should query Opt->strWordDiv before every search to make sure it is relevant, there is probably no need to do so on every iteration within a search.
Perhaps get_word_div() can be a class member `string m_WordDiv`, updated in `Viewer::Search` only?

alabuzhev:
> Separate PR for sure. OK?

Of course. Thank you.

far/viewer.cpp
alabuzhev: Let's leave it as is for now.

far/strmix.cpp
alabuzhev: > First, I am confused by your link. I could not find this definition of isspace(x) on master.

It is no longer there of course.
The link is just to explain why in many places we use iswblank and not iswspace, even though iswspace might appear more idiomatic in general.

> I strongly suggest leaving everything as is

I'm not suggesting changing anything in this PR, just making sure there are no other places that are accessed by Viewer::Search and potentially need an update.

And we probably should rename BlankOrWordDiv accordingly.

far/farlang.templ.m4
alabuzhev: > what do you think of changing radio button to checkbox here? It'll be easier to fit in one line in some languages.

Let's solve one thing at a time. Editor does not have a hex search yet so there is nothing to worry about :)

In general, "hex" is probably idiomatic / well known enough to leave it as is?
If not, I suppose it can be shortened in other languages too, same as we don't write "hexadecimal" in English here.

far/string_utils.cpp
```diff
@@ -399,11 +412,16 @@ TEST_CASE("string.case")
 
 TEST_CASE("string.utils")
 {
-	for (const auto& i: GetSpaces())
+	for (const auto& i: GetBlanks())
```
alabuzhev, line 415: I think this whole test case is redundant - everything in it is already checked in string.blanks, string.spaces, string.eols above. Could you remove the whole thing please?

far/viewer.cpp
MKadaner: That's a good point. Changed to call it once per search only in Text mode.

far/strmix.cpp
MKadaner: Thanks for the explanation. As far as I can tell, `Viewer::Search` is clear. I renamed `BlankOrWordDiv` to `SpaceOrWordDiv`.

I also looked at all uses of `IsBlankOrEos` and `GetBlanks`. There are not many of them, and everywhere using "Space" instead of "Blank" may be reasonable, except for (at least in UCRT):
- `iswsapce` includes U+0085 (NEL) and U+00A0 (NBSP). It's an interesting question whether we want to break words at NBSP, even in Viewer Search. Maybe we should everywhere use `GetSpaces() + GetEols()` instead of `iswsapce`?
- For codepoints outside UNICODE C0/C1 (> U+00FF), `iswsapce` calls `GetStringTypeW` which returns who knows what. Again, switching to `GetSpaces() + GetEols()` may work more predictably.

What do you think?

far/farlang.templ.m4
MKadaner: ACK

far/string_utils.cpp
```diff
@@ -50,11 +50,16 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 //----------------------------------------------------------------------------
 
-string_view GetSpaces()
+string_view GetBlanks()
 {
 	return L" \t"sv;
 }
 
+string_view GetSpaces()
```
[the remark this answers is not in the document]
MKadaner: Yeah! I removed `GetSpaces` for now. However, see my comment about the difference between `iswbalnk` and `iswspace`. One day we my want it back.

Please consider everything and make the final call, I will do whatever you say. My personal opinion is that we should consistently use `GetSpaces() + GetEols() + Global->Opt->strWordDiv.Get()` as word separators, but I very well may miss some subtle scenarios. I will change Viewer Search accordingly. Other places better to change in a separate PR. Please create and issue and assign to myself.

I added `IsSpaceOrEos` for the symmetry (and used it in `is_word_div`). Then I noticed this record in changelog 🤣

```
svs 26.05.2009 23:56:23 +0300 - build 948
. . .
2. Добавлена IsSpaceOrEos(), руками не мацать, нужна.
```

far/string_utils.cpp
MKadaner: Done. I pulled `isblank` test up into `"string.blanks"` though, just in case.

far/string_utils.hpp
```diff
@@ -50,10 +50,13 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 constexpr bool IsEol(wchar_t x) noexcept { return x == L'\r' || x == L'\n'; }
 
 [[nodiscard]]
-inline bool IsBlankOrEos(wchar_t x) noexcept { return std::iswblank(x) || !x; }
+inline bool IsBlankOrEos(wchar_t x) { return !x || std::iswblank(x); }
```
alabuzhev, line 53: No noexcept?

far/strmix.cpp
alabuzhev: > It's an interesting question whether we want to break words at NBSP, even in Viewer Search

Why not. NBSP is about preventing line breaks, not word breaks.

far/string_utils.cpp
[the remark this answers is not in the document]
alabuzhev: Let's go with minimal changes that would allow the viewer search to work as expected.
Blanks vs spaces is not the goal of this PR after all. We can explore possible approaches later if needed.

far/string_utils.hpp
MKadaner: Yes. `iswblank` / `iswspace` may call `GetStringTypeW`, so who knows whether it can throw. And if it does throw, the process will be terminated with prejudice, which is likely not good.

far/strmix.cpp
MKadaner: I often use NBSP in contexts like "64&nbsp;KiB" or "1776&nbsp;г." or "T.&nbsp;Pratchett". Not sure if "KiB" should be found in whole words mode. But then, why not? Let's keep `iswspace` for now.

far/string_utils.cpp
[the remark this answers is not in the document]
MKadaner: Sure, Removing `IsSpaceOrEos` in the next iteration.

far/string_utils.hpp
alabuzhev: C functions, as well as Win32 APIs, don't throw C++ exceptions by definition.

far/string_utils.cpp
```diff
@@ -348,11 +348,12 @@ std::optional<std::pair<size_t, size_t>> detail::fuzzy_searcher_impl::find_in(co
...
 #include "testing.hpp"
 
-TEST_CASE("string.spaces")
+TEST_CASE("string.blanks")
 {
-	for (const auto& i: GetSpaces())
+	for (const auto& i: GetBlanks())
 	{
 		REQUIRE(std::iswblank(i));
+		REQUIRE(std::isblank(i));
```
alabuzhev, line 356: Redundant (and UB if `i` is not within unsigned char range anyway).
It was used in another test because of a typo.

far/viewer.cpp
```diff
@@ -1007,20 +1011,25 @@ void Viewer::SetStatusMode(int Mode)
...
+	static constexpr std::array extra_word_div{ encoding::bom_char, encoding::replace_char };
+
+	auto word_div{ Global->Opt->strWordDiv.Get() };
+	word_div.append(extra_word_div.cbegin(), extra_word_div.cend());
+	return word_div;
 }
 
 void Viewer::ReadString(ViewerString *pString, int MaxSize, bool update_cache)
 {
+	auto WordDiv{ Global->Opt->strWordDiv.Get() };
```
alabuzhev, line 1030: const maybe?
Maybe even const& - we probably don't need a copy here?

far/viewer.cpp
```diff
@@ -1007,20 +1011,25 @@ void Viewer::SetStatusMode(int Mode)
...
+	auto IsWordDiv{ [&WordDiv](wchar_t ch) { return IsBlankOrEos(ch) || contains(WordDiv, ch); } };
```
alabuzhev, line 1031: Now we have is_word_div and IsWordDiv, it might get confusing.

far/string_utils.cpp
MKadaner: My bad. :(

far/string_utils.hpp
MKadaner: I know; still somehow get caught up in it. :(

far/viewer.cpp
MKadaner: Oh! I somehow thought `Get()` returns `string`; `const&` it is!

far/viewer.cpp
MKadaner: Makes sense; changed to `CanWrapLineAt`. As an aside, I wonder if a couple of `IsBlankOrEos`s below in `Viewer::ReadString` (but likely not the third one) should also be `CanWrapLineAt`. I did not look closely.

alabuzhev (approved):

MKadaner:
Thank you for the review. Squashed, rebased, ready.

alabuzhev:
Thank you

far/viewer.cpp
```diff
@@ -130,12 +130,15 @@ Viewer::Viewer(window_ptr Owner, bool bQuickView, uintptr_t aCodePage):
 	ViOpt(Global->Opt->ViOpt),
 	Reader(ViewFile, (Global->Opt->ViOpt.MaxLineSize*2*64 > 64*1024 ? Global->Opt->ViOpt.MaxLineSize*2*64 : 64*1024)),
 	strLastSearchStr(Global->GetSearchString()),
-	LastSearchCaseSensitive(Global->GlobalSearchCaseSensitive),
-	LastSearchWholeWords(Global->GlobalSearchWholeWords),
-	LastSearchReverse(Global->GlobalSearchReverse),
-	LastSearchHex(Global->GetSearchHex()),
-	LastSearchRegexp(Global->Opt->ViOpt.SearchRegexp),
-	LastSearchDirection(Global->GlobalSearchReverse? -1 : +1),
```
alabuzhev, line 138: @MKadaner looks like this one was here for a reason: reverse search doesn't work as expected now.

STR:
- view a file with multiple occurrences of some pattern
- F7, enter it, enter
- Shift+F7 a few times
- Alt+F7 once. It goes back.
- Alt+F7 again. It goes forward.
- Alt+F7 again a few times. It loops.
- F7. Note that `[x] Reverse search` is selected.

Restoring this variable fixes the issue.

alabuzhev: No problem, let's wait till the weekend.

MKadaner: Done. See #599.

PR #628, opened 2023-02-17
Added help article about custom language files.

MKadaner:
## Summary
As promised, added help article "Customizing UI elements."

## Checklist
- [X] I have discussed this with project maintainers: #616 <br/>


alabuzhev:
Thanks

HamRusTal:
Does it make sense to suggest proofreading corrections to the new article without submitting a separate PR?
I see at least one spelling error (in the Russian variant).

far/FarRus.hlf.m4
```diff
@@ -5657,6 +5661,43 @@ $ #Информация о версиях#
...
+Manager’а, определены в файлах #.lng# (например, #FarRus.lng#).
+Пользователь может изменить эти элементы, как ему удобно. К примеру,
+можно изменить горячие клавиши или надписи на каком-нибудь диалоге.
+Чтобы переопределить элементы пользовательского интерфейса, необходимо
+создать файл с именем #Far<Lng>.lng.custom# в папке #%FARHOME%# или
+#%FARPROFILE%# и задать новые значения для тех UI ID’ов, которые нужно
+изменить.
+
+ Например, для того, чтобы типы ссылок файлов и папок на панелях Far
+выглядели также, как в листинге DOS’овской команды DIR, можно создать
```
MKadaner, line 5676: It would be simpler if you just pointed the error and asked to fix it. :)

```suggestion
выглядели так же, как в листинге DOS’овской команды DIR, можно создать
```

Is it this? Please add more if I am missing something else.

This PR is already merged and closed, so a new commit is required. I can create a new PR, but somebody would need to approve / merge it anyway, so, @alabuzhev, could you please push the correction to reduce the hustle all around?

> ... the wording looks too promising to me ...

Why, you can also modify hotkeys, the feature that prompted the whole discussion and this PR, by the way. :)

To be serious, may I ask you to suggest more precise wording? At least a sketch? I can polish it later. Just open a bug issue or an idea discussion and post what you would like to see in this (or any other) article. With maintainers' help, I will handle it as time permits.

HamRusTal: > Is this it? Please add more if I am missing something else.

Yes, that one. Another error I see is in `двумя способами, на двух отдельных строках или на одной строке` — IMHO a colon should be used here instead of a comma, as we have a general name followed by a list of specific names (I'm not sure these terms are linguistically correct but I hope you got the idea).

Also I suggest dropping the mentions of DOS, as strictly there's no DOS subsystem in any modern Windows and the DIR command we have just uses the same name and serves the same purpose. (And I doubt the genuine DIR command even of the latest MS DOS versions marked junctions or symlinks as such even if the OS supported NTFS which I doubt too.) So I suggest `listing of DIR command` and `листинге команды DIR` (or maybe `listing of cmd.exe's DIR command` and `листинге команды DIR в cmd.exe`) for English and Russian respectively.

MKadaner: To the best of my knowledge, using of colon here is not mandatory, tough I understand the idea. I have no strong opinion either way.

As for the use of "DOS," it was intentional, a bit of tong in cheek, if you wish. "Cmd.exe" sounds somewhat mouthful (both in English and Russian). Besides, somewhere down the line it was "command.com." I do not remember whether it was on one of the old versions still amazingly supported by Far; and do not want to investigate. :) Again, I am not very strong on this, just explaining my choice.

HamRusTal: I have no strong feelings on these either, just shared my thoughts.

As for `DOS`, I'm still not clear what was the purpose, but that's not too important.

PR #643, opened 2023-02-27
gh-626: Added build checker for changelog headers.

MKadaner:
## Summary
What can I say. Added.

 
## References
#626

## Checklist


misc/build-checks/hlf-checker/CommandLine.cs
```diff
@@ -5,7 +5,7 @@
 using System.Linq;
 using System.Text.RegularExpressions;
 
-namespace HlfChecker
+namespace hlf_checker
```
yegor-mialyk, line 8: I would leave it as is, a Pascal Case is the correct (read "widely used") name convention for C# namespaces and classes.

MKadaner: Yeah! Just wanted to experiment with the style of project names and directories which are more in line with everything else under FarManager\misc. It looks ugly indeed. Let me change it. It's still a draft, isn't it? :)

MKadaner: So much better!

MKadaner:
CodeFactor complains about [`Complex Method complexity = 18`](https://www.codefactor.io/repository/github/fargroup/farmanager/pull/643). I think it's a false positive. Is there any way to suppress the diagnostic in this place?

alabuzhev:
> Is there any way to suppress the diagnostic in this place?

No idea, but the method does look somewhat complex indeed. Maybe there is a way to simplify it?

.github/workflows/build.yml
```diff
@@ -21,20 +21,39 @@ jobs:
         working-directory: far
         run: python tools/source_validator.py
 
+#------------------------------------------------------------------------------
+  changelog-checks:
```
alabuzhev, line 25: Do we really need to run this and hlf-checks separately on different VMs?

Yes, running things in parallel is good in general, but in this case both checks will likely take only a few seconds, way less than the preparations.
Maybe both can be steps in one workflow?

MKadaner: Not at all. The only real advantage of having two separate checks is accounting. One can immediately see in the PR "Checks" section which check failed, `hlf` or `changelog`. It was the way to ask whether we want it for the price of two VMs. I like them combined as well.

MKadaner:
> the method does look somewhat complex indeed.

Please see if you like it better. CodeFactor is happy, but whether the code became simpler is a matter of taste.

alabuzhev:
Thank you.

PR #648, opened 2023-03-07
Fix broken link in macroapi manual

johnd0e:
Reported here: https://forum.farmanager.com/viewtopic.php?p=171968#p171968


enc/enc_lua/macroapi_manual.en.tsi
```diff
@@ -997,6 +997,7 @@ mtime=3833370824
 #_@@@
 #_[acall]: 91.html
 #_[deserialize]: 104.html
+#_[EnumScrips]: 106.html
```
alabuzhev, line 1000: Shouldn't it be EnumScrip**t**s here and below?

Issue #542, opened 2022-08-15
Find the Pattern under Caret with Reverse Search


MKadaner:
### Description of the new feature or improvement

At some point between 3.0.5720.0 and now, the behavior of Internal Editor's reverse search changes slightly. In the olden days, if I reverse search for `abc` with the caret at the line `abc` in the position of the `b` or `c`, the pattern I am at will be found. Now it is skipped, and the previous occurrence, if exists, is found or "Could not find..." dialog is displayed.

I almost started implementing the change but happened to check the behavior of MSVC editor. Not entirely surprisingly, it also skips the pattern under the caret when searches backward. It does make some sense. After all, if I search _forward_, the pattern under caret is not found, and it feels quite natural. My only problem is that somehow I am fine with this behavior in VS, but expect different behavior from Far. It breaks my workflow. 🙂 Just recently, I was sorting out through a large list of long similar-looking strings and for some time could not understand why I could not find a certain string, even though I've just found it searching forward from the beginning of the file.

What is general opinion about reverting the old behavior?

### Proposed technical implementation details (optional)

I can implement this change if people here think it's a good idea.

alabuzhev:
@MKadaner I am fine with either behavior, as long as:
1. it is consistent between the regular search and the regex search (looks like currently it is not)
2. next/previous searches do not stuck, regardless of the `Cursor at the end` option.

P.S. I do not remember changing this intentionally. It could be accidental or an attempt to fix #⁠2.

MKadaner:
Thank you, @alabuzhev, I'll look at it then. (I did not even remember there was this option. 🙂)

What do you mean "stuck?" The search does not progress on repeated pressing of `Shift+F7`/`Alt+F7`?

alabuzhev:
> What do you mean "stuck?" The search does not progress on repeated pressing of `Shift+F7`/`Alt+F7`?

Yes, and `F7 Enter`.

MKadaner:
Please assign this issue to me. I am knee deep in search anyway.

HamRusTal:
> > What do you mean "stuck?" The search does not progress on repeated pressing of `Shift+F7`/`Alt+F7`?
> 
> Yes, and `F7 Enter`.

IMHO `F7 Enter` should **not** move forward if the caret is at a match already (in a sense that this exact match should me found). And if the `Cursor at the end` option is off, then even the caret should not move in this case.
[...]

PR #651, opened 2023-03-09
gh-542: Editor Search (plain text): Shift+F7 / Alt+F7 now finds the nearest match

MKadaner:
## Summary
Adjusted calculation of the search start position, so that `Shift+F7` / `Alt+F7` do not skip the nearest match if it is overlapping with the previously found match.

Also, Fixed regex search in cursor-at-the-end mode. `Shift+F7` now continues from the next character.

## Checklist
- [X] I have discussed this with project maintainers: gh-542, gh-660<br/>

## Details
Based on the discussion here, overall regex search behavior did not change.


alabuzhev:
@MKadaner is it about
find: aa
in: aaaa
results: **aa**aa a**aa**a aa**aa**
instead of 
results: **aa**aa aa**aa**
?

If so, then yes, it makes perfect sense, this is how it initially was and the current behavior is a bug.

It looks like the fix doesn't cover "Find all", could you check please?

MKadaner:
Hm-mm... It's vise versa. Current behavior:
find: aa
in: aaaa
search forward results: **aa**aa a**aa**a aa**aa**
search backwards results: aa**aa** **aa**aa

First of all, it is inconsistent between searching forward and backwards. I prefer the forward-search behavior and attempted to make backwards-search behave symmetrically.

I spot-checked Far30b2738.x64.20120618. There was no "Cursor at the end" then. The behavior matches my expectations: _all_ overlapping patterns are found in either direction. 

Sure, I'll fix "Find all" when we decide what behavior is desirable.

P.S. I've just realized that this change does not actually address #542 😞.

"Cursor at the end" is _off_

```
line:  aaaa
caret:    ^
```

`F7` `aaa` `Alt+U` (search backwards)

Actual:
```
line:  aaaa
caret: ^
```

Expected:
```
line:  aaaa
caret:  ^
```

I'll see what I can do with this, but again, after we decide what to do in general.

HamRusTal:
In Far 3.0.6099, searching in Editor gives AAaa, aaAA; searching in Viewer gives AAaa, aAAa, aaAA.
I'm strongly for them to be consistent, and what the Viewer gives looks correct.

As for the caret position, IMHO, when the found match is selected, search forward/backward should continue with no regard to the caret position, based only on the selection's beginning/end respectively.
This becomes trickier when the caret is disjoint from selection (e.g. with persistent blocks). Probably the search should continue from the caret in this scenario.

An interesting edge case is when a new selection is made (and the caret “touches” the selection or is inside it). Should the search continue based on the selection or based on the caret? Maybe Far should memorize the last highlighted match and continue search iff that match is still re-/selected, and search from the caret otherwise?

rohitab:
@MKadaner I just tried out your changes. It's working the way you described. Backwards search now behaves the same as Forward search.

I noticed another issue, that is present in the current version of _Far_, without your changes. Since it seems related to the changes you're working on, I'm reporting it here.

1. Enter the text **abcabcabc**, i.e. three groups of **abc** without any spaces
2. Press `F7` to search, enter **abc** and press `Enter`
3. Remove any selection and position the cursor to the first character of the above text: **<ins>a</ins>bcabcabc**
4. Press `Shift+Right` three times to select the first **abc** group
5. The cursor should now be positioned on the second group: **`abc`<ins>a</ins>bcabc**
6. Press `Shift+F7` to continue search
7. The second **abc** group is found/selected: **abc`abc`abc**
8. Repeat steps 3 to 6
9. Note that instead of the second group, the third group is now found/selected: **abcabc`abc`**
10. Each time you repeat steps 3 to 5, the found match alternates between the second and third group

Note that it happens even without selecting the text, and just positioning the cursor on the first character of the second group. I hope I was able to explain it properly.

MKadaner:
> In Far 3.0.6099, searching in Editor gives AAaa, aaAA; searching in Viewer gives AAaa, aAAa, aaAA.

Correct with a note that in Editor, this happens when continuing searching _backwards_. Continuing _forward_ search moves caret one character at a time, just like in Viewer.

> I'm strongly for them to be consistent, and what the Viewer gives looks correct.

Agreed on both points.

>As for the caret position, IMHO, when the found match is selected, search forward/backward should continue with no regard to the caret position, based only on the selection's beginning/end respectively.

Yeah! In the "Select found" mode, search _continue_ should skip the entire selected pattern. This is not implemented in the prototype here. I will work on it.

> This becomes trickier when the caret is disjoint from selection (e.g. with persistent blocks). Probably the search should continue from the caret in this scenario.

This is how it should be now, though I have not tested.

> An interesting edge case is when a new selection is made (and the caret “touches” the selection or is inside it). Should the search continue based on the selection or based on the caret? Maybe Far should memorize the last highlighted match and continue search iff that match is still re-/selected, and search from the caret otherwise?

This is more or less how it is implemented now. Far keeps track of the last found match (selected or not). Then the following algorithm is used:

1. If caret is _not_ on the same line with the last found, search from caret.
2. Define _Active_ end of the last found as the beginning of the match if "Cursor at the end" is _off_; otherwise _Active_ is the end of the match.
3. If caret is not at the Active end, search from caret.
4. If the search is in the direction of the match relative to the caret (== Active end), continue search from the next character in the search direction (the next character into the match), or from the caret if it is a new search (after showing the dialog).
5. Otherwise, search from the _Passive_ (the opposite to Active) end of the match plus one character in the search direction.

Incidentally, this is literally what I implemented in the new function, only in C++ instead of English. Some observations:

- If caret is at the active end, new search
  - In the _Natural_ direction always starts from the caret. _Natural_ means forward if "Cursor at the beginning," or backward if "Cursor at the end." This is how it has always worked.
  - In the _Unnatural_ direction always start from one character into the match from the _Passive_ end.
  - This trick allows to find overlapping patterns.
- If caret was moved around and eventually returned to the Active end of the match (though the user has no visual clue), `Shift+F7` / `Alt+F7` _continue_ search from last found, i.e., search start is punted by one character.
  - I think that this case should be treated as continue search from an arbitrary place, i.e., search should start from the caret.
  - Implementing this requires "forgetting" last found if caret was touched after last successful search.
  - @alabuzhev, can you suggest how to do the trick (other than finding all places where `m_it_CurLine` is modified? There should be some place where plugins are notified about caret movement.

MKadaner:
> I noticed another issue...

🤣 This is exactly how it should work, except for it is of course dumb. This happens because Far remembers the last found match, but never visualizes it. Selection does not matter. It's all about caret position relative to the invisible last found match. Whether caret was moved between the searches also does not matter as long as it was returned to the last found. Please read my long explanations (which crossed with your bug report). Even though caret is both times at the same position, the beginning of the second occurrence, the last found alternates between the second and the third occurrences.
- If the last found is the second occurrence, _continue_ search find the third one, otherwise we would be stuck at the second and could never have a progress.
- If the last found is the third occurrence, _continue_ search find the second one, because caret is nowhere near the last found.

This stupid behavior reiterates the importance of dropping last found if the user touched the caret (maybe did something else, need to think), between consecutive continue searches.

MKadaner:
> P.S. I've just realized that this change does not actually address #542 😞.

It was in the morning when I am too sleepy to think clearly. My change _does_ address #542 in _continue search_ case. A new search always starts from the caret because we cannot step back (towards the "passive" end of the last match) as I did in this prototype, because there is no last match during new search.

MKadaner:
And thank you for testing, @rohitab.

MKadaner:
@HamRusTal Sure thing. The purpose of this draft PR, along with #542, is exactly to discuss how it should work.

Jumping ahead, one thing I'd like to reaffirm is that the behavior should be symmetrical in two dimensions: search direction and "Cursor at end" mode. My prototype fits the bill (and it's trivially verifiable in the code). 😄 BTW, the proposed behavior is very similar to the current one, except for continue search backwards finds overlapping patterns; the current behavior being annoyingly _asymmetrical_ in this respect.

alabuzhev:
> should be symmetrical in two dimensions

Yes, as much as possible in *all* dimensions: editor/viewer, forward/backward, cursor in the front/end, non-/persistent blocks etc., unless there are complexities with the implementation or valid logical reasons to do it differently.

rohitab:
@MKadaner Thanks for the detailed explanation.
> This stupid behavior reiterates the importance of dropping last found if the user touched the caret (maybe did something else, need to think), between consecutive continue searches.

I agree. If the user moves the caret, the search should be from that position. This is how Visual Studio does it as well. If you run the test case I described above, Visual Studio will always find the second group. You can also try this in [Visual Studio Code for the Web](https://vscode.dev/), which behaves the same, as does Google Chrome.

There is another issue with the current implementation as well. If you select a word from right-to-left and continue search, nothing happens the first time, and there is no feedback given to the user. You need to continue search a second time for the next match to be found. This happens with the _All matching entries_ menu as well -- if you press `Enter` in the menu to go to the matched entry, and then press `Shift+F7` to continue search, nothing happens the first time. Here too, you need to continue search a second time for the next match to be found. Both issues are probably the same, since the caret is at the beginning of the selected text when you continue search.

One last thing -- in most editors I use, if you press `Right` (after finding a word) the caret is moved to the position after the selected text; in _Far_, the caret is moved to the next character of the selected text. Also, when text is found, the caret is usually positioned at the end of the text, unlike _Far_, which positions it at the beginning. Thought I'd bring this up for discussion as well. :grinning:

HamRusTal:
> One last thing -- in most editors I use, if you press `Right` (after finding a word) the caret is moved to the position after the selected text; in _Far_, the caret is moved to the next character of the selected text.

This is not specific to searching, hence out of scope here. Also, many Far users are accustomed to this behavior, so it should not change without proper discussion on the Far forum.

HamRusTal:
@MKadaner Sorry for the delay. Here are my thoughts (all of them IMHO; also, please see P.S. and P.P.S. at the end on the possible sources of confusion):

> In the "Select found" mode, search _continue_ should skip the entire selected pattern.

I disagree strongly! The matches being found should **not** depend on the “Select found” option. Neither should they depend on the “Cursor at the end” option. In other words, search _continuation_ should iterate all matches, overlapping or not.
Of course, this only applies when the selection matches the last found occurrence and the caret is _attached_ to the selection (that is, the caret is inside the selection or touches it from the right).

> 1\. If caret is _not_ on the same line with the last found, search from caret.

I disagree with the “same line” criterion. A line can be very long, the selection may be out of the visible area still on the same line. That's why I suggested the “caret is disjoint (detached) from the selection” criterion.

> 3\. If caret is not at the Active end, search from caret.

I disagree again. Where the caret is within the selection should not affect what (overlapping) matches are found by search continuation. Hence my note on the “Cursor at the end” option above.

> 4\. from the caret if it is a new search (after showing the dialog)

I agree, but this is irrelevant here as we're discussing search continuation only.

> 5\. Otherwise, search from the _Passive_ (the opposite to Active) end of the match plus one character in the search direction.

I disagree yet again. What end of the match is used as the base position for search continuation should depend only on the search direction, not on the caret position (again, as long as it's attached to the selection) or on the “Cursor at the end” option. Or do you see scenarios when my “simplistic” approach would give unintuitive results?

> If caret was moved around and eventually returned to the Active end of the match (though the user has no visual clue), `Shift+F7` / `Alt+F7` _continue_ search from last found, i.e., search start is punted by one character.
> * I think that this case should be treated as continue search from an arbitrary place, i.e., search should start from the caret.

I agree.

> Selection does not matter. It's all about caret position relative to the invisible last found match.

This is counter-intuitive. If the found match is still selected (and the caret is still attached to it), it should be the selection that drives search continuation. Only if the selection changed or caret detached from the (preserved) selection should the search continuation act as a new search from caret position.

P.S.
Reading your explanations, I find it hard to tell the difference between the logic that published Far builds (which exactly?) have already, the logic you've implemented in the draft being discussed (at the moment of writing), and the logic that you intend to implement. Can you please make this distinction clearer in future comments?

P.P.S.

> This is exactly how it should work, except for it is of course dumb.

When you say “should”, do you mean “how it is implemented” or “how I-as-a-user prefer”? To have a productive discussion, we need to be very clear about this difference every time.

MKadaner:
> If the user moves the caret, the search should be from that position.

I skimmed through the code. It looks like, unfortunately, there is no easy way to trigger an action (e.g., reset last found) on a cursor move or a text change. This improvement (one may call it bug fix) should be implemented by honestly scanning the entire Editor.cpp and inserting `ResetLastFound()` everywhere where it should be. As such, this fix is absolutely out of scope of this PR. @rohitab, could you please create a separate issue for this? I'll think of it.

>  If you select a word from right-to-left and continue search, nothing happens the first time, and there is no feedback given to the user...
> ... the caret is at the beginning of the selected text when you continue search.

Yes, this is how Far Editor search works. I do not think we should change it. At least not without thorough discussion. The confusion is because there is no visual indication of the last found match. Internally, Editor knows about it and operates logically and appropriately, but the user has no clue. 😞 Currently, I am thinking about how the last found could be made visible _and_ distinguishable from the regular selection. This is how VS works, and there is no confusion there. **If anyone has any ideas, please chime in.**

> [I]n most editors I use, if you press Right (after finding a word) the caret is moved to the position after the selected text; in Far, the caret is moved to the next character of the selected text...
> ... when text is found, the caret is usually positioned at the end of the text, unlike Far, which positions it at the beginning.

There are two Editor options, "Select found," and "Cursor at the end." Together, they make Far Editor's behavior closer to other editors (but probably still somewhat different). The key thing here (just like with everything around Editor Search) is where the caret is currently positioned. Selection does not matter. @rohitab, could you please try these options (if you already have not). Does it make more sense to you? If you see the room for improvement, please create separate issue(s). As @HamRusTal pointed out, this behavior is very traditional and should not be changed without proper discussion.

HamRusTal:
> It looks like, unfortunately, there is no easy way to trigger an action (e.g., reset last found) on a cursor move or a text change.

How so?! There are events `EE_CHANGE` (upon every text change) and `EE_REDRAW` (before every redraw, e.g. upon a selection change and hopefully upon caret movement). As Far emits these events to plug-ins, there _should_ be a way to subscribe to them within Far's own source code… If there's no such way, that's a significant shortcoming. @alabuzhev Any ideas?

MKadaner:
> There are events `EE_CHANGE` ... and `EE_REDRAW`

Thanks. I'll have a look. What I did notice thought was that `Editor::TextChanged(bool State)` is called _manually_ all over the place. This suggests that either there is no one central place to handle such event or that the use of this function predates plugin API. Anyway, if I add resetting the last found as necessary, it'll be in a separate PR.

P.S. I'll reply to your comments later today or tomorrow.

MKadaner:
> The matches being found should not depend on the “Select found” option.

In fact, I personally in favor of this requirement. So, let's keep the existing behavior for now. However, I can understand the users who prefer different behavior depending on whether the found string becomes selected or not. If a search operation positions the cursor and does nothing more, the feeling is that the overlapping pattern near the cursor should be found. On the other hand, if the search operation selects the found string, the next found should be beyond the selection. Here I use "should" in the sense of how it looks more natural, at least to some users. Again, currently, found matches do not depend on the “Select found” option, and I am going to keep it this way.

> Neither should they depend on the “Cursor at the end” option. In other words, search continuation should iterate all matches, overlapping or not.

In my implementation, repeated searches find all matches, no matter the search direction or where the cursor is positioned.

> Of course, this only applies when the selection matches the last found occurrence and the caret is attached to the selection (that is, the caret is inside the selection or touches it from the right).

Sorry, I do not follow you. If search does not depend on selection, why any difference? Maybe you can explain with examples?

> > 1. If caret is _not_ on the same line with the last found, search from caret.
>
> I disagree with the “same line” criterion. A line can be very long...

If the caret is not on the same line with the last found, it is detached from the selection (if the selection even exists), hence you criterion is satisfied. This statement says nothing about what happens if caret _is_ on the same line.

I think there is miscommunication here. Those numbered items represent steps of the (actually implemented) algorithm, not the description of user's experience.

> > 3. If caret is not at the Active end, search from caret.
>
> I disagree again. Where the caret is within the selection should not affect what (overlapping) matches are found by search continuation.

Again, it's just one step in the algorithm, not the final user experience. Please note also that we are _not_ talking about _selection_ here, only about _the last found match_ which is itself not visible and may or may not be selected depending on the “Select found” and partly "Persistent blocks" options.

> ... we're discussing search continuation only.

Nope! We are discussing a general question of where to start search depending on many different conditions, including first-search vs, continue-search. They are all related and cannot be discussed separately.

> > 5. Otherwise, search from the Passive (the opposite to Active) end of the match plus one character in the search direction.
>
> I disagree yet again. What end of the match is used as the base position for search continuation should depend only on the search direction, not on the caret position ...

Again, this statement should not be considered on its own. It is meant to be the last step of the algorithm, after all other variants were already handled.

> > Selection does not matter. It's all about caret position relative to the invisible last found match.
>
> This is counter-intuitive. If the found match is still selected (and the caret is still attached to it), it should be the selection that drives search continuation. Only if the selection changed or caret detached from the (preserved) selection should the search continuation act as a new search from caret position.

I did not say that search continuation should _always_ start from the caret position. I said that the behavior depends on _the caret position relative to the invisible last found match._ In the current implementation and in my prototype, selection is not taken into account. As I noted at the beginning of this comment, it may be a good idea to consider selection and behave differently, but I think you said (and I agree) that the selection should not be considered.

> I find it hard to tell the difference...

Sure, will try to explain in the next comment. Short answer, there is no difference except for some annoying asymmetry is fixed.

> When you say “should”, do you mean “how it is implemented” or “how I-as-a-user prefer”?

Here I meant famous "by design" or “how it is implemented” if you wish. Sorry about imprecise language. When I see some silly behavior reported, sometimes I cannot help becoming sarcastic, and this mood does not promote clarity. Will try to keep more to the point.

MKadaner:
OK. Let me try to describe the current behavior of Far Editor Search at the very high level. There are two ways to search: `F7` or `Shift+F7` / `Alt+F7`. In the previous discussion we called them first-search and continue-search.

The first-search starts from the current cursor position. Period. Full stop.

Particularly, it means that if the cursor is at the first character of the matching string when we search forward, or at the last character of the matching string when we search backward (past the last from the naïve user's point of view), we immediately find this string. Cursor does not move.

In its turn, continue-search may be performed in two different states: immediately after the previous successful search, or after the user did something else (moved around or modified the text, maybe something else). In the latter case, search starts from the current cursor position. Period. Full stop.

The same note about finding the match at the cursor (and the cursor does not move) applies. As an aside, sometimes, this behavior (either after first-search or after continue-search-from-elsewhere) surprises or even annoys the user. It is perceived as if the search "stuck," does not make any progress. The user may complain that he has to press `Shift+F7` again.

Now we are left with continue-search right after the pattern was found. Let's call it repeat-search. The term may be somewhat confusing, but since we are going to discuss only this case, let's live with it.

The desired behavior in the repeat-search case is obvious. We want to find the very next match, even in the case when the matches overlap. However, we do not want to find the match we've just found (we are in `Shift+F7` - `Shift+F7` situation, or `Shift+F7` - `Alt+F7`, or similar). Otherwise, the search is really stuck. If the same match is found on the second repetition, it will be found on the third, etc.

The trick question is how to detect "continue-search right after the pattern was found" condition. From the user's point of view, it is obvious. He pressed `Shift+F7` twice in a row. Unfortunately, the existing code layout does not make it easy to implement this condition literally this way.

So, somebody came up with an idea of keeping track of the last found match, and if the cursor is at head of the last found (I'll explain in a moment). consider it repeated search. Works reasonably well in general, but sometimes throws up interesting glitches, like the one found by Rohitab.

It's important to note that this condition is treated very precisely, cursor at the head of the last found match. Not "attached" to the match. but exactly at the head. For example, if the cursor is in the middle of the last found match, it means the user moved the cursor, thus, he lost ~his innocence~ the state of repeat-search. Now it is mere continue-search from an arbitrary place, and we start searching from the cursor. Again, sometimes the result may look counterintuitive, but I do not see better alternatives here.

About the term "head of the match."
- In the "cursor at the beginning" mode, it's the left end of the string (beginning of the match).
- In the "cursor at the end" mode, it's the right end of the string (end of the match).

Why is the difference? Why we want dependency on "cursor at" whatsoever? Because this is where the cursor is positioned after the match is found, and to detect the repeat-search (i.e., the state when the user did not touch the cursor) we should compare against this place, not against the other end.

MKadaner:
Now about the current implementation, my fix, and the consistent behavior overall. Here, I am discussing repeat-search only (I'll make a note otherwise) and concentrate on overlapping patterns. Everything else is simple.

Viewer: repeat-search finds overlapping patterns in either direction. There is no notion of repeat-search because there is no cursor, and the found match is always selected. For example, if we search for `aaa` in the string `aaaaaaaaaa`, the last found match (the same as selection) will move one character at a time in the direction of the search, no matter the direction. Not much room for a counterintuitive behavior.

Editor, same string and pattern, current behavior:
- Search forward, cursor at the beginning (normal case):
  Just like in Viewer, the last found match moves forward one character at a time. There is or there is no selection (depending on "Select found"). But the cursor always jumps forward one character at a time. Nice, everyone is happy (hopefully).
- Search backward, cursor at the beginning:
  _Unlike_ in the Viewer, the last found match moves backward one match length at a time. Again, there may or may not be the selection, but the cursor jumps backward by the long strides and misses some occurrences of the pattern. This is the most annoying case, at least to me, and one of the reasons why I am touching Far Search.
- Cursor at the end, either direction:
  Again, the search progresses by match length, not by one character. Do not know why it is asymmetrical compared to the "cursor at the beginning" state. I do not use "cursor at the end," so I care less, but of course I am fixing it as well.

Editor, same string and pattern, proposed behavior:
- Search _always_ progresses one character at a time, no matter search direction, cursor placement, or "Select found."

As an aside, finding the nearest match when searching against the grain (backward if cursor is at the beginning, or forward if the cursor is at the end) is a bit tricky. This is why the algorithm deals with the match tail (called "Passive" end in my previous comments). I am proud of how simple the final code is. I am going to rename some identifiers to make it even clearer,

MKadaner:
One last note. One may argue that if both "Select found" and "Cursor at the end" are _on_, the repeat-search (of overlapping matches) should progress by match length. After all, Visual Studio and many other editors effectively operate in this mode and behave this way. 

. . .

Just checked VS. It's even more complex. Particularly, the change of contniue-search direction (`F3` to `Shift+F3`) breaks the repeat-search mode. After that, when in repeat-search mode, it finds _all_ overlapping patterns, i.e., the selection moves one character at a time.

Anyway, I am not going to do anything special for this mode in Far unless somebody (a) convinces me that it is worth it, and (b) specifies the exact and complete algorithm.

HamRusTal:
@MKadaner Thanks for the extensive explanations!

> I think you said (and I agree) that the selection should not be considered

I believe I did not say that.

But never mind. If successive Find-Next commands traverse all (overlapping) matches regardless of direction and of the mentioned options, I'm fine with that. Thanks for making this happen!

P.S.
If we happen to continue the discussion (or have a new one on Find-Next), I suggest that we revise the terminology, as I feel the continue-search and repeat-search terms as meaning-swapped.

MKadaner:
How about "search-after-search"?

MKadaner:
OK. It's all fine for plain string search. Regex search is of course broken. Here is what happens:

#### Current behavior
Cursor is in the middle of an arbitrary non-empty line. Search _backwards_ for `.*` regex. Found match starts one character to the left of the original cursor position and ends at the line end. Search-after-search works more or less intuitively in either direction.

#### With my change to calculate the next search start position (this draft PR)
The new search works the same way. However, _backwards_ search-after-this-new-search finds the last two characters on the line. After that, the search is stuck.

This happens because we give regex search the entire line and the start position. When searching backwards, it finds the rightmost match which still starts to the left of the start position _in the entire line_.

Visual Studio behaves differently. Regex search backwards stops at the caret position, even though the characters to the right of the caret could be included in the match. After that, search-after-search works intuitively.

I am going to change the behavior of backwards regex search (and maybe make other tweaks (fuzzy search?) as necessary) to limit the found match by the original cursor position.

Any concerns?

MKadaner:
> ... the definitions here are more intuitive.

😄 The problem with intuition is that, unlike common sense, it's uncommon. To be serious, I perceive the meaning of these terms the way I used them. This was why I tried to introduce the awkward search-after-search thingy in the reply to your comment where you suggested to swap the meanings for the first time. At least it was something new, not loaded with the history. Maybe "search-next-prev"? Anyway, our discussions (if we need more) should be concentrated on the second bullet in your previous comment, so we may do with just one term and assume that it always has this meaning.

> ... in this case the matches found by successive Find-Previous commands will not be the same as ones found by successive Find-Next commands.

That is true. My point is that there is nothing wrong with this behavior when we are talking about regexes, which are inherently anisotropic (embed the notion of direction).

> Is it feasible to implement the backwards regex _continue-search_ in such a way that the _beginning_ of the match should shift to the left compared to the last search but the _length_ of the match is not limited by the caret position?

This is how it works currently in production. Of course, it is possible to keep this behavior. I am considering this option as well. It will require two completely different implementations of `CalculateStartPositionForSearchReplace`, but why not? The problem with this existing behavior is that it is also not entirely logical, (symmetrical). Consider search for the regex `.*`. Currently, search backwards _does not_ find the characters from the beginning of the line to the cursor, though they match the pattern and are in the direction of the search. My proposal is based on Visual Studio's behavior, which is more intuitive to me. 🤣.

Once again, I am not strong about it.

> Can you please give examples when fuzzy search backwards works in an counter-intiutive way in its current “stock” implementation?

No. Not off the top of my head. I mentioned fuzzy because it can match strings of different lengths, which is a necessary (not a sufficient) condition for the trouble. I did not have time to test and just put it here as a reminder.

HamRusTal:
> Currently, search backwards _does not_ find the characters from the beginning of the line to the cursor, though they match the pattern and are in the direction of the search.

But, first of all, why should the search backwards find such a match?! With the `.*` regex, the match should last past the caret, till the EOL. So I don't see a problem in this case.  
IMHO finding the same matches regardless of direction is the primary requirement.

MKadaner:
> But, first of all, why should the search backwards find such a match?!

Because when I search backwards, I expect everything to be found _to the left of the caret._

> With the `.*` regex, the match should last past the caret, till the EOL.

Where the "past the caret" assumption comes from? If we search forward, we start at the caret. If we search backwards, we stop at the caret.

HamRusTal:
> Because when I search backwards, I expect everything to be found _to the left of the caret._

I am not sure this is what most Far users expect.



> Where the "past the caret" assumption comes from?

From the primary requirement that the same matches are found regardless of direction.

> If we search backwards, we stop at the caret.

I'm not so sure. Anyway, such a change in the search logic warrants prior discussion on the forum, maybe even a poll.

MKadaner:
> > If we search forward, we start at the caret.
>
> When we talk about Find-Next, it's simply not true, as you explained above.

It is true, in the "cursor at the beginning" mode. In the "cursor at the end" mode, the logic should be inversed, if we want symmetry.

> [T]he primary requirement that the same matches are found regardless of direction.

It does not work this way universally. With `.*` regex and caret at the end, first time forward we find everything from the caret to EOL, next time forward we find the entire next line. If we reverse and search backwards, we sequentially find all string suffixes, including those we skipped on the second search forward.

Even without regexes, when the caret was moved to the middle of a potential match, we find it searching backwards, but not forward (or vice versa if cursor is at the end). Search direction matters.

If the idea is that the search in Far finds the matches no matter search direction and caret positioning, searching for `.*` should find the entire line always.

> [S]uch a change in the search logic warrants prior discussion on the forum, maybe even a poll.

Do you mean forum.farmanager.com or ru-board? Personally, I think that, forums being in Russian, potential audience here is wider. But who knows the actual coverage, since the audience is probably mostly Russian-speaking as well.

HamRusTal:
> Do you mean forum.farmanager.com or ru-board? Personally, I think that, forums being in Russian, potential audience here is wider.

I mean the official forum at forum.farmanager.com, and it has a subforum for English speakers (although with very low activity).

I'll answer other points tomorrow. What I see right now is that it's hard again to understand whether you make statements on the current “stock” implementation, on this PR implementation, or on your vision of how things should work (to a lesser extent here). Can you please try to make these very clear every time?

MKadaner:
In my last comment, I meant current production behavior, for example Far Manager, version 3.0.6116.0 x64.

And let me clarify my first remark, I feel it may not be entirely clear.

> > > If we search forward, we start at the caret.
> >
> > When we talk about Find-Next, it's simply not true, as you explained above.
>
> It is true, in the "cursor at the beginning" mode. In the "cursor at the end" mode, the logic should be inversed, if we want symmetry.

My statement should read:

Current production behavior. Option "Cursor at the end" is _off_. Either regex or plain text search. Search forward. Search starts at the current cursor position. Always. The found match lies entirely to the right of the current cursor position.

I do hope this time I managed to make myself unambigous.

MKadaner:
I think that we should treat these two issues separately. Plain text search should find all matches left and right, no matter any other options, like "Cursor at end." This was the main purpose of this PR (and my beef for years). I believe we agree on the proposed change.

Regex search behavior appears to be controversial, surprising to me at first, but it takes all sorts to make a world. Let's agree to disagree. I'll keep the existing behavior for regexes. It should always be safe to keep the existing behavior. At the very least it's not my fault. 😄 I'll create a separate issue / discussion.

Have a great weekend. Stay tuned.

HamRusTal:
> Let's agree to disagree. I'll keep the existing behavior for regexes. It should always be safe to keep the existing behavior.

Agreed. Thank you! And sorry for not having answered earlier.

MKadaner:
I reverted regex search behavior. Seem to be working as in production.

@rohitab, when you have time, would you be so kind as to test regex search in Editor? Especially, compare new search up / down and `Shift+F7` / `Alt+F7` behavior vs. production behavior. Thank you in advance!

rohitab:
@MKadaner Sure thing, but I'm currently swamped with work, and might only be able to do it over the weekend.

MKadaner:
@rohitab, No problem. I am not sure I'll have time to clean up or fix anything before the weekend either.

MKadaner:
I fixed _Search All_ to match the behavior of _Search Forward._ @rohitab, Could you please test it as well?

If nothing is broken, it should be all that is needed for this PR. So far, it is still a draft.

MKadaner:
I happened to have a bit of time to spare, so I tested the new behavior best I could. I think the PR is ready for review / merge. If Rohitab finds any issues, I'll fix them separately.

far/editor.hpp
```diff
@@ -289,6 +289,8 @@ class Editor final: public SimpleScreenObject
 	enum class SearchReplaceDisposition;
 	SearchReplaceDisposition ShowSearchReplaceDialog(bool ReplaceMode);
 	void DoSearchReplace(SearchReplaceDisposition Disposition);
+	int CalculateSearchStartPosition(bool Continue, bool Backward, bool Regex) const noexcept;
```
alabuzhev, line 292: Let's not mark things `noexcept` unless necessary (e.g. it's a part of a contract or such a hot loop that it makes a difference).

MKadaner: Sure, but why? I've always thought it a good idea to [[u]se `noexcept` when exiting a function because of a throw is impossible or unacceptable](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-noexcept).

The only reason not to blindly slap `noexcept` on a function which _cannot_ throw I am aware of is a public library, where breaking API changes are expensive, impractical or unacceptable, and the author cannot be sure future implementations can be kept `noexcept`. The key word here is _public library_.

alabuzhev: Exactly for the reason quoted: in principle, throw is neither impossible nor unacceptable in this case.
Indeed, it does not explicitly throw now because there is only some integer math inside, but it is an implementation detail than might change one day for any reason: we might call a function that throws or change that "assert" to "throw logic_error" etc.

If we blindly slap noexcept on every internal function that doesn't throw *today*, someone will have to go through the codebase, inspect the call tree and manually remove noexcept from all the affected functions every time the error handling logic changes somewhere, and there are more entertaining things in the world than that.

There are about 600 occurrences of noexcept in Far today and about 750 with plugins. Most of them were probably added by me, so I'm not rejecting the concept per se. However, noexcept is a commitment and should be treated with care to not cause more harm than good.

As a rule of thumb, if a function is longer than one line, if probably should not be noexcept.

MKadaner: > If we blindly slap noexcept on every internal function that doesn't throw today, someone will have to go through the codebase, inspect the call tree and manually remove noexcept...

I see your point (and I kinda expected this argument), however:
Removing `noexcept` should happen when the function's implementation changes, not while fixing a crash. For internal functions, it's similar to renaming them or otherwise changing their signatures as things evolve. The `noexcept` spec is not stuck there forever. Anyway, I noticed your opinion and will follow the rules of the house of course (and if I forget, which may happen, please be patient with me).

What do you think of merging this PR? I have the next in the pipeline. :)

MKadaner: P.S. My liking of `noexcept` stems from a real-world experience, when I traced enormous build times to huge code bloat caused by lack of `noexcept` on many relatively large but provably non-throwing functions. Even with forced heavy inlining, compiler could not prove everything and had to painstakingly generate megabytes of code to execute destructors as necessary. This issue is exacerbated by the (good) style of writing many small functions.

alabuzhev: > Removing `noexcept` should happen when the function's implementation changes

Indeed, but unfortunately it is not enforced by the compiler and thus it is too easy to get it wrong. We're only humans.
I wish `noexcept(auto)` was a thing, or better the default.

> What do you think of merging this PR?

LGTM, I can merge if you have nothing else to add to it (the discussion got quite long and I didn't follow everything).

MKadaner: Please merge. I believe that my discussion with @HamRusTal is concluded (to mutual respectful disagreement), so there should not be problems here. I also tested the new behavior as best as I could. If @rohitab (or anybody else) finds any issues, I'll fix them separately.

MKadaner:
Thank you!

alabuzhev:
Thank you too.

PR #662, opened 2023-03-27
Forum: Search / Replace dialog facelift.

MKadaner:
## Summary
On popular demand, renamed `Search down` and `Search up` buttons to `Find next` and `Find prev` respectively, added separators, and renamed related identifiers in the code.

## Checklist
- [X] I have discussed this with project maintainers: [Far support forum](https://forum.farmanager.com/viewtopic.php?t=13158).<br/>


MKadaner:
Rebased, ready to merge. Thanks.

MKadaner:
Ditto.

MKadaner:
Thanks.

HamRusTal:
I just had a look at the new design… Can we please have the `Find/Replace previous` buttons **to the left** of the `Find/Replace next` ones, please? Otherwise their order is counter-intuitive to the vast majority of Far users (writing left-to-right; it was discussed on the Forum recently that Far is not suitable for RTL writing). The `Find/Replace next` buttons would stay default though.

MKadaner:
I can swap _Find/Replace Next/Prev_ buttons. Personally, I do not care much because I use hotkeys anyway. My only concern is who will be next to complain and suggest a better solution? :)

I added the vertical separator to make _Search/Replace_ dialog follow the pattern of _Find file_ one. Yes, I also noticed that because _Search/Replace_ dialog is shorter, and because of the horizontal half-separator, the overall result looks too much like a griddle. I certainly can remove it and reduce the width of the half-separator. By how many characters? Will be great if you post a mush up. And the same concern about who is next to complain. :)

I am still out of touch with the people here and especially on the forum. @HamRusTal, @alabuzhev, could you please confirm the request for these changes, and I'll effect them in a few days.

w17:
Ough! You didn't know... Search actually is Pandora box.

MKadaner:
In fact, I knew. Or at least could guess pretty well. :)

Thought more explanations needed.

1. I do not mind making changes. I just want to minimize UI churn for others who is not involved in these discussions.
2. I mean, I still do not know whose say goes here. Who is the PM so to say? Or should such changes be voted, or what? Again, it's not a big deal to make a change. But extra changes would frustrate more people, so better to get the final word before going.

w17:
Personally I am agreed with **HamRusTal** -- Prev on the left, Next on the right.
It is more intuitive especially if customize Next/Prev as arrow symbols (maybe I'll do that).
For my opinion dialog is good enough now.
And it is not possible to make all happy so just take it easy...

HamRusTal:
> Will be great if you post a mush up.

I suggest this:
```
╔═════════════════════════════ Replace ══════════════════════════════╗
║ Search for                                  [ Word ] [ Selection ] ║
║ S                                                                 ↓║
║ Replace with                                                       ║
║ R                                                                 ↓║
╟────────────────────────────────────────────────────────────────────╢
║ [ ] Case sensitive                [ ] Regular expressions          ║
║ [ ] Whole words                   ──────────────────────────────── ║
║ [ ] Fuzzy search                  [ ] Preserve style               ║
╟────────────────────────────────────────────────────────────────────╢
║          [ Replace previous ] { Replace next } [ Cancel ]          ║
╚════════════════════════════════════════════════════════════════════╝
```
The metaphor is that the checkboxes are all in the same section laid out in two columns, and there's a horizontal separator between the search-related and the replace-related ones.

Also I suggest that the `Find/Replace next` buttons are given their own hotkey too (e.g. `N`), for people to be able to activate it by left hand alone.

alabuzhev:
> Can we please have the `Find/Replace previous` buttons to the left of the `Find/Replace next` ones, please? Otherwise their order is counter-intuitive

Such order would be more logical indeed, but it's a double-edged sword: the user will need to take less logical steps to get there, e.g. Shift+Tab instead of Tab.

> could you please confirm the request for these changes

Personally, I don't care that much.

HamRusTal:
> the user will need to take less logical steps to get there, e.g. Shift+Tab instead of Tab

Can you please elaborate in which scenarios you foresee this?

alabuzhev:
> Can you please elaborate in which scenarios you foresee this?

Searching up, assuming there are no hotkeys.
Enter text, Tab (to leave the edit control), End (to go to the default button), Shift-Tab.

HamRusTal:
> Enter text, Tab (to leave the edit control), End (to go to the default button), Shift-Tab.

Oh, this requires esoteric knowledge (as this function of `End` is missing from the Help's section on Dialogs in the `@MiscCmd` page). And it will be more natural to use `Left` there rather than `ShiftTab`.

Issue #618, opened 2023-01-28
Editor Replace Button Hotkey


alabuzhev:
### Discussed in https://github.com/FarGroup/FarManager/discussions/616

<div type='discussions-op-text'>

<sup>Originally posted by **rohitab** January 27, 2023</sup>
Most dialogs in _Far_ do not have hotkeys for the default button, since you can just press `Enter`. For example, in the _Search_ dialog in Viewer/Editor/Help, the _Search_ button does not have a hotkey. However, in the _Replace_ dialog in the Editor, the _Replace_ button has an `Alt+R` hotkey.

What do you think about removing the hotkey from the _Replace_ button and assigning `Alt+R` to the _Replace with_ edit control instead? It seems natural to me, to use `Alt+S` and `Alt+R` to switch between _Search_ and _Replace_ with boxes. It might inconvenience people that are accustomed to using the `Alt+R`/`Alt+E` hotkeys though.

> [MKadaner](https://github.com/MKadaner): I also noticed this discrepancy. I think this is a good idea. The decision is mostly for the language maintainers, except for there is no dedicated maintainer for English localization.</div>

HamRusTal:
I have R assigned to “Replace with” locally for years, so I'm both hands for this.
Also, I assigned E to “Regular expressions” so that I don't need to use both hands to toggle it (doing it «одной левой»).
```
// For builds after 6095
MSearchReplaceReplaceWith = "&Replace with"
MSearchReplaceRegexp = "R&egular expressions"
MSearchReplaceReplace = "Replace"
```

MKadaner:
Thank you, @alabuzhev. I did not realize that a discussion cannot be assigned. Sorry for all the hassle. It would be way simpler if you just pushed this change.

PR #666, opened 2023-04-01
Continuation of 6122. Follow-up on the discussion in gh-662.

MKadaner:
## Summary
Swapped buttons on the `Search / Replace` dialog and removed vertical separator.

 
## References

- #616
- #618

## Checklist
- [X] I have discussed this with project maintainers: #662 <br/>

## Details
I did not assign hotkey to the `Find/Replace next` buttons because we seemed to agree (#616, #618) that default dialog buttons should not have hotkeys. Besides, `N` is not exactly within the left hand's reach (though close).

@HamRusTal, if you insist on assigning the hotkey, please chime in. I would rather prefer keeping it consistent with other dialogs.


alabuzhev:
Thanks

HamRusTal:
Thank you!

> if you insist on assigning the hotkey
I don't. After all, I can assign it locally.

MKadaner:
Thank you all.

PR #673, opened 2023-04-13
Propose fareditor macros to start/end selection with hotkeys/menu

sfkulyk:
## Summary
Add FarEditor macro to mark selection start/stop using Ctrl+( or Ctrl+).
Also these actions available from menu which allows user to use them during macro recording (Ctrl+.)
If persistent blocks are enabled and you have active selection, macros automatically detect selection type (lines/blocks) and change existing selection end or start position.

## References
It is no bug/issue.

## Checklist
This was discussed several times in semiofficial telegram channel, also we have few topics on Forum with Macros, but as it is not a part of main FAR code, there were no official discussion.

Nevertheless macros can be very useful for different purposes. common usecase example:
Ctrl+(  F7(find something including using regexp for findings) Ctrl+) Ctrl+C

## Details
If you want to reject this one, kindly spend 5 minutes and try to use this macro to understand how useful it can be. Thank you

### DIFF
--- a/extra/Addons/Macros/Descript.ion
+++ b/extra/Addons/Macros/Descript.ion
@@ -10,6 +10,7 @@ Edit.Notepad.lua Editor: Notepad-like shortcuts
 Edit.OpenURL.lua Editor: Open URL or UNC under cursor
 Edit.SaveAndExit.lua Editor: Save File and Exit
 Edit.SaveFile.lua Editor: Save File
+Edit.SelectionStartEnd.lua Editor: Start/end selection with separate hotkeys
 Editor.ColorWord.moon Editor: Color Word Under Cursor
 F3.lua Use internal editor as viewer
 F9DeactivateMenu.lua Use F9 to deactivate main menu
--- a/extra/Addons/Macros/Edit.SelectionStartEnd.lua
+++ b/extra/Addons/Macros/Edit.SelectionStartEnd.lua
@@ -1,0 +1,81 @@
+local F = far.Flags
+local Selection = {}
+
+Event {
+  group="EditorEvent";
+  action=function(id,Event)
+    if Event==F.EE_CLOSE then
+      Selection[id] = nil
+    end
+  end
+}
+
+local function changeBlock(x,id,CurLine,CurPos)
+  local sel = editor.GetSelection(id) or Selection[id] or {}
+  sel[x.."Line"] = CurLine
+  sel[x.."Pos"] = CurPos
+  Selection[id] = sel
+  if not sel.EndLine or not sel.StartLine then
+    return
+  end
+  local startpos, endpos
+  if sel.BlockType==F.BTYPE_COLUMN then --https://bugs.farmanager.com/view.php?id=2759
+    startpos = editor.RealToTab(id, sel.StartLine, sel.StartPos)
+    endpos = editor.RealToTab(id, sel.EndLine, sel.EndPos)
+  else
+    startpos = sel.StartPos
+    endpos = sel.EndPos
+  end
+  local width = endpos-startpos+1
+  local height = sel.EndLine-sel.StartLine+1
+  editor.Select(id, sel.BlockType or F.BTYPE_STREAM, sel.StartLine, startpos, width, height)
+end
+
+local function selectStart()
+  local info = editor.GetInfo()
+  changeBlock("Start", info.EditorID, info.CurLine, info.CurPos)
+end
+
+local function selectEnd()
+  local info = editor.GetInfo()
+  changeBlock("End", info.EditorID, info.CurLine, info.CurPos-1)
+end
+
+Macro {
+  description="Mark selection start";
+  area="Editor";
+  key="CtrlShift9";
+  action=selectStart;
+}
+
+Macro {
+  description="Mark selection end";
+  area="Editor";
+  key="CtrlShift0";
+  action=selectEnd;
+}
+
+-- Following menu items are useful when recording macros
+-- Otherwise do not include them into menu
+
+MenuItem {
+  description="[macro] Selection start &(";
+  menu="Plugins";
+  area="Editor";
+  guid="AB0E7F16-C942-4780-AD33-E8493CF66196";
+  text=function()
+    return far.MacroGetState()~=F.MACROSTATE_NOMACRO and "&( Selection start"
+  end;
+  action=selectStart;
+}
+
+MenuItem {
+  description="[macro] Selection end &)";
+  menu="Plugins";
+  area="Editor";
+  guid="6CC1CDA4-A550-48EB-BC65-DD6EE177E2E2";
+  text=function()
+    return far.MacroGetState()~=F.MACROSTATE_NOMACRO and "&) Selection end"
+  end;
+  action=selectEnd;
+}
--- a/misc/msi-installer/features.wxs
+++ b/misc/msi-installer/features.wxs
@@ -76,6 +76,7 @@
           <ComponentRef Id="Edit.OpenURL.lua" />
           <ComponentRef Id="Edit.SaveAndExit.lua" />
           <ComponentRef Id="Edit.SaveFile.lua" />
+          <ComponentRef Id="Edit.SelectionStartEnd.lua" />
           <ComponentRef Id="F3.lua" />
           <ComponentRef Id="F9DeactivateMenu.lua" />
           <ComponentRef Id="F9Table.lua" />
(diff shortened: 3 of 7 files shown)


alabuzhev:
@sfkulyk LGTM (unless @shmuz has anything to add), thanks.
Could you please also update the descript.ion in the same directory and the installer?
1e017a10d4461db2a6d613d8a84aa80c950770b6 could be used as a reference (no need to update far/changelog).

MKadaner:
The `Ctrl+]` combination is traditionally assigned to jumping to the matching bracket (FarColorer, Brackets plugin, Visual Studio). Is it possible to use different accelerator keys for this functionality, for example `Ctrl+(` / `Ctrl+)`? Of course, one can adjust the keys before installing the macro, but it would be nice if there are no conflicts out of the box.

shmuz:
@alabuzhev IMO it's OK (though I would do it w/o editor event handler).
WRT to placing it to Addons - I have no opinion, decide for yourself.

bitraid:
> The `Ctrl+]` combination is traditionally assigned to jumping to the matching bracket (FarColorer, Brackets plugin, Visual Studio). Is it possible to use different accelerator keys for this functionality, for example `Ctrl+(` / `Ctrl+)`? Of course, one can adjust the keys before installing the macro, but it would be nice if there are no conflicts out of the box.

And Far uses `Ctrl+[`/`Ctrl+]` to insert the path from left/right panel.
(Also, lines 12, 14 mixes spaces with tabs for indentation)

HamRusTal:
The macros' descriptions are `"^Ctrl[ Selection start/end"` — the `^` character is misleading.

extra/Addons/Macros/SelectionStartEnd.lua
```diff
@@ -0,0 +1,96 @@
...
+  action=startSelection;
+}
+
+MenuItem {
+  description = "Selection &] end";
+  menu   = "Plugins";
+  area   = "Editor";
+  guid   = "A435D561-AD64-4DD1-8C61-28CB90358818";
+  text   = "Selection &] end";
+  action=startSelection;
```
johnd0e, line 95: `=` alignment here is not consistent.
(I suppose we need no alignment, as no other macro has it)

johnd0e, line 95:
```suggestion
  text   = "Selection &] end";
  action=endSelection;
```

extra/Addons/Macros/SelectionStartEnd.lua
```diff
@@ -0,0 +1,96 @@
...
+        StartPos=1,
+       	EndLine=1,
+        EndPos=0
+      }
+    end
+    if Event==far.Flags.EE_CLOSE then 
+      Selection[info.EditorID]=nil
+    end
+  end
+}
```
johnd0e, line 22: This is totally redundant.

johnd0e:
By the way, it was never mentioned, but the scrips is mostly copy-pasted/adapted from [`Ctrl-K-blocks.lua`](https://forum.farmanager.com/viewtopic.php?t=8815).

sfkulyk:
Hello all and thank you for all comments there!

I have updated alignment, It is a bit strange as in local copy I never used tabs..
Changed hotkeys to Ctrl( and Ctrl)
Fixed copypaste issue in menu item

I am not sure that I know how to make good workaround without events - if you have multiple editor windows opened, event looks like the most simple and stable feature to manage independent selection buffers in each editor...

johnd0e:
> I am not sure that I know how to make good workaround without event - if you have multiple editor windows opened

It is completely pointless to store/keep/update `Selection` in event handler, as in either case you are ending up getting editor info on every macro call.

sfkulyk:
> > I am not sure that I know how to make good workaround without event - if you have multiple editor windows opened
> 
> It is completely pointless to store/keep/update `Selection` in event handler, as in either case you are ending up getting editor info on every macro call.

Removed event handler, thanks for point me at this moment.

HamRusTal:
A few more notes on the proposed code:
1. The top line comment is clearly outdated. Also it mentions some Macros Menu which is not a thing, and the script adds its items to the regular plugin commands menu.
2. The top-level `Color` and `Selection` variables should clearly be `local`.
3. The records are only being added to the `Selection` table and never removed, which can cause (a) unwanted results in new editors and (b) memory leaks in extreme cases.
4. The hint is displayed too briefly, so there's no chance to actually read it.
5. The selection type is still being reset to Stream which is **not** what a user expects when having a Block selection.

extra/Addons/Macros/SelectionStartEnd.lua
```diff
@@ -0,0 +1,114 @@
+﻿-- Selection Start/End with hot-keys Ctrl+( / Ctrl+) or using Menu
+local Color = far.AdvControl(far.Flags.ACTL_GETCOLOR,far.Colors.COL_HELPTOPIC)
```
yegor-mialyk, line 2: If we removed `Hint`, `Color` is not needed either unless I'm missing something.

sfkulyk:
Hello all,

Thanks for the comment!
Let me try to respond.

1. I have removed hint. If was very useful for debugging, but really not needed in everyday usage
2. all variables are local now
3. I have returned event handler to initialize and clear up table
4. About a stream/block, This macro have no sinergy with manual selection, because it should know the position of start and end. So you can't select block, then change block selection using macros - it will just select the place between start and end.
But I have added two separate functions for start/end block selection. I put them without hotkeys, and I am not sure I can find correct hot keys for them. Maybe will find good idea in future versions...
5. Updated description

About menu:
One of the main purpose of this macros - is make a possible to record macro with Ctrl+. recording.
And it is impossible to call macros from hotkey during recording, but it is a way to call them from plugin menu using hotkeys "(" and ")" which are highlighted in menu. So menu items are very important for this use case.

HamRusTal:
> you can't select block manually, then change it using macros

In practice, that scenario works and is very handy. If it is not supported and will likely break at some point, it's too significant a limitation IMHO. Scripts like these should work based on the existing editor selection (if any).

sfkulyk:
Such scenario will works only in Persistent blocks are enabled.
In my scenario, the most of use cases are record macro and call macros from plugin menu during macro recording.
Never works with Persistent blocks...

Do you want me to do such investigation?

HamRusTal:
> only in Persistent blocks are enabled

Not necessarily, as there are scripts that store the selection whenever it changes (in particular when it gets cleared) and one can restore the saved selection by a shortcut. After that it's possible to apply the suggested macros based on the  restored selection.

> Do you want me to do such investigation?

Well, investigation is too ambitious a term, yet some learning will likely be necessary.

sfkulyk:
> > only in Persistent blocks are enabled
> 
> Not necessarily, as there are scripts that store the selection whenever it changes (in particular when it gets cleared) and one can restore the saved selection by a shortcut. After that it's possible to apply the suggested macros based on the restored selection.

IMHO It is absolutely different use case.
With this macro I want definitely to mark start and end position of block selection, which is not related to manual selection.

So actually I can even manipulate with two selections in parralel
Ctrl( on required text block begin, then edit text and even select and copy/paste something during editin. then go to required block of text end, press Ctrl) and it will select the text exactly between Ctrl( and Ctrl), neverless that I have used selection after that, because block begin was remembered before. I think it is a good approach - just tried to do that.

HamRusTal:
> So my proposition is to keep current behavior as is.

Then I'm no longer interested. Good luck!

sfkulyk:
> This is clearly not what a user like me needs.

I have made some investigation...
I present a great reworking of the script.
1. If persistent blocks are enabled, and selection is already exists, it change the existing selection. In this case correctly change selected area according to the existing block/lines selection type
2. If persistent blocks are not enabled or no selection exists, by default select block type is BTYPE_STREAM
3. Removed extra functions as the single function detects existing block type
4. Inside handler I am checking the handler type and invoce editor.Getinfo only for editor open/close file
5. Some refactoring to make script easier to read

Actually I am really happy that script behavior is pretty logical according to persisting block option enabled/disabled. With disabled option, it works as I like initially, with enabled option it works as @HamRusTal  asking.

Kindly provide your comments/approval about the last commit.

extra/Addons/Macros/SelectionStartEnd.lua
sfkulyk: I have moved editor.GetInfo inside if/end, so it will calls only during file opend/close

sfkulyk:
@johnd0e , @yegor-mialyk script has been reworked according to comments.
Kindly review/provide your approval for merge

johnd0e (approved):

shmuz:
@alabuzhev 
I'm trusting @johnd0e so count me too as an "approver".

alabuzhev:
Ok.
FYI I renamed it to Edit.SelectionStartEnd.lua for consistency with other Editor macros there.

Issue #613, opened 2023-01-21
Regular expressions search option not synchronized between Viewer and Editor


rohitab:
### Far Manager version

3.0.6094.0 x64

### OS version

10.0.22623.1180

### Steps to reproduce

1. Open Editor
2. Press `F7` to open _Search_ dialog
3. Check _Regular expressions_ option
4. Press `Enter` to search
5. Ignore any messages and close Editor
6. Open Viewer
7. Press `F7` to open _Search_ dialog
8. Uncheck _Regular expressions_ option
9. Press `Enter` to search
10. Ignore any messages and close Viewer
11. Open Editor again
12. Press `F7` to open _Search_ dialog

### Expected behavior

_Regular expressions_ option should be unchecked.

### Actual behavior

_Regular expressions_ option is checked. 

Note that all other search options _Case sensitive_, _Whole words_, _Reverse search_ and _Fuzzy search_ are synchronized between the Viewer and Editor. _Regular expressions_ is the only one that is managed separately. It is also the only option that is saved between restarts of _Far_. It appears that these settings are stored in `Editor.SearchRegexp` and `Viewer.SearchRegexp`.

One other minor difference between Viewer and Editor -- in Viewer, when you check _Regular expressions_, _Whole words_ and _Fuzzy search_ are disabled; in Editor they are not.

I am aware that the search is being unified, so most of these issues will most likely be fixed. Decided to report them anyway.

MKadaner:
Thank you for the report, @rohitab, very useful. You are right, the unified search / replace dialog will enable / disable controls as necessary. Since you've tested the PR version of the improved dialog and did not complain, I assume that this functionality works satisfactorily. If I missed something, I will always be able to fix the behavior later, and it will be fixed for all three, Editor, Viewer, and Help.

The _Regex_ option is different. @alabuzhev, do you know why it is handled differently than all other ones? Again, it is easy to fix, but hard to tell who'll shout out.

alabuzhev:
> why it is handled differently than all other ones?

Probably an overlook: regex was added way later than the other options and, as far as I remember, not simultaneously to Viewer / Editor.
I see no reason for it to behave differently.

> in Viewer, when you check Regular expressions, Whole words and Fuzzy search are disabled; in Editor they are not

[...]

Issue #612, opened 2023-01-21
Search options do not save when searching from Help


rohitab:
### Far Manager version

3.0.6094.0 x64

### OS version

10.0.22623.1180

### Steps to reproduce

1. Restart _Far_. This is required to reset search options.
2. Press `F1` to launch _Help_ from any Panel
3. Press `F7` to open _Search_ dialog
4. Enable the option _Case sensitive_
5. Press `Enter` to search
6. Press `Esc` to close _Help_
7. Press `F1` to relaunch _Help_
8. Press `F7` to open _Search_ dialog

### Expected behavior

_Case sensitive_ search option should be enabled.

### Actual behavior

_Case sensitive_ search option is disabled. Note that this happens with _Whole words_, _Regular expressions_ and _Fuzzy search_ options as well.

Next, please do the following:
1. Open the Viewer or Editor
2. Press `F7` to open the _Search_ dialog
3. Enable the _Case sensitive_ option
4. Press `Enter` to search
5. Ignore any error messages and close the Viewer or Editor
6. From any Panel press `F1` to launch _Help_
7. Press `F7` to open the _Search_ dialog
8. Note that _Case sensitive_ is now enabled
9. Turn off the _Case sensitive_ option
10. Press `Enter` to search
11. Press `Esc` to close _Help_
12. Press `F1` to relaunch _Help_
13. Press `F7` to open _Search_ dialog
14. Note that _Case sensitive_ is still enabled, even though it was turned off in Step 9 above.

To summarize, the search options in _Help_ are always initialized based on the last state from the Viewer/Editor. The changes made to the search options from _Help_ are not saved.

MKadaner:
Thank you for reporting, @rohitab. Yes, this is how it works. The fix would be simple. As usual, I am very wary about changing the behavior which has been there for a long time.

Not saving search options, including search string, on closing Help may have been very intentional. Help and Editor / Viewer are two very different contexts. I can imagine a scenario when I am analyzing something juggling Find File and several Editors and keep (a part of) the context of my investigation in the search term available in all Search dialogs everywhere. If during such exercise I happen to open help and search for something, I will lose my context if Help overrides the last search string. Admittedly, this is less of an issue with search history available.

@alabuzhev, what do you think? Maybe we should keep a separate Help Search context between Help invocations? This way, something will be preserved, and it would be obvious to the user that it's intentionally _not_ the same as Editor / Viewer context.

alabuzhev:
[...]

PR #678, opened 2023-04-30
gh-612, gh-613: Rationalized and unified global Search / Replace parameters.

MKadaner:
## Summary
1. gh-612: Help Search parameters are preserved between Help sessions.
2. gh-613: Search patterns and parameters are shared between Viewer, Editor, and Find File.

 
## References
- #612
- #613

## Checklist
- [X] I have discussed this with project maintainers: #612, #613
 <br/>


far/macro.cpp
```diff
@@ -3912,7 +3912,6 @@ void FarMacroApi::editorsetFunc() const
 		ShowScrollBar           = 15,
 		EditOpenedForWrite      = 16,
 		SearchSelFound          = 17,
-		SearchRegexp            = 18,
```
MKadaner, line 3915: - Should I remove this option / index elsewhere, including documentation? I searched for `Regexp` in likely places, but I could miss some places altogether.
- What about backward compatibility? Will it break existing macros?

far/global.hpp
```diff
@@ -42,14 +42,58 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+	if constexpr (requires { ClientParams.SearchStr; })     SaveParam(GlobalParams.SearchStr, ClientParams.SearchStr);
+	if constexpr (requires { ClientParams.SearchBytes; })   SaveParam(GlobalParams.SearchBytes, ClientParams.SearchBytes);
+	if constexpr (requires { ClientParams.ReplaceStr; })    SaveParam(GlobalParams.ReplaceStr, ClientParams.ReplaceStr);
+	if constexpr (requires { ClientParams.Hex; })           SaveParam(GlobalParams.Hex, ClientParams.Hex);
+	if constexpr (requires { ClientParams.CaseSensitive; }) SaveParam(GlobalParams.CaseSensitive, ClientParams.CaseSensitive);
+	if constexpr (requires { ClientParams.WholeWords; })    SaveParam(GlobalParams.WholeWords, ClientParams.WholeWords);
+	if constexpr (requires { ClientParams.Regex; })         SaveParam(GlobalParams.Regex, ClientParams.Regex);
+	if constexpr (requires { ClientParams.Fuzzy; })         SaveParam(GlobalParams.Fuzzy, ClientParams.Fuzzy);
+	if constexpr (requires { ClientParams.PreserveStyle; }) SaveParam(GlobalParams.PreserveStyle, ClientParams.PreserveStyle);
```
alabuzhev, line 95: This is an interesting pattern, but it looks rather fragile: if someone renames a parameter in ClientParams (to be more descriptive or whatever), this place can be easily overlooked because the compiler won't complain at all.

Maybe just make the struct copyable & assignable and let the clients have their copies?

If we decide to proceed with the suggested above hex logic encapsulation,
the data-only part of the structure can be extracted into a base structure and shared with the clients.

far/macro.cpp
MKadaner: I can do that. But why regexes only? How about case sensitive and fuzzy and the whole gamut of them?

Apparently, option `SearchPickUpWord` (19) was [removed from the code back in 2015](https://github.com/FarGroup/FarManager/commit/15d9f2fcade62d3c41b6cadb6a4032c5437dd6dc#diff-c125991267d8110ae22f0d586120f3c49298cb7bbe7a29fb4209b6f356a6c11c), but nobody updated the doc.

Maybe it's better to deprecate the option than maintaining some of parameters but not the others?

Up to you. Please let me know how to proceed.

far/global.hpp
```diff
@@ -42,14 +42,58 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
 // External:
 
 //----------------------------------------------------------------------------
 
-enum class lng;
-enum class search_case_fold;
+struct global_search_replace_params
+{
+	NONCOPYABLE(global_search_replace_params);
+	NONMOVABLE(global_search_replace_params);
```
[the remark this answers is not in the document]
MKadaner: Ah, I thought of these things differently. More like a high-level design choice, which may also reflect performance considerations. Something on the lines of, "this thing, while technically copyable (or movable), do not need to be copied (or moved) around by design. To prevent accidental expensive operations, let's outright ban them. If you really think you need a copy, change the code here, but you better have a really good reason."

Anyway, I will remove `NONMOVABLE`.

far/stddlg.cpp
```diff
@@ -76,6 +76,8 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 //----------------------------------------------------------------------------
 
+template void operator<<<>(global_search_replace_params&, const SearchReplaceDlgParams&);
```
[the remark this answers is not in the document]
MKadaner: Apparently, last night I screwed it up (`extern template` in the header is important) and also overcomplicated (do not need `<>` at all). Thank you for calling out this place.

far/stddlg.cpp
```diff
@@ -190,17 +192,10 @@ SearchReplaceDlgResult GetSearchReplaceString(
 
 	const auto SetFlagIf{ [&](const item_id Item, const auto Flag, const bool Condition) { if (Condition) DlgItems[Item].Flags |= Flag; } };
 	const auto SetSelected{ [&](const item_id Item, const bool Selected) { DlgItems[Item].Selected = Selected; } };
-	const auto SetStringIf{ [&](const item_id Item, const string_view String, const bool Condition) { if (Condition) DlgItems[Item].strData = String; } };
+	const auto SetStringIf{ [&](const item_id Item, auto&& String, const bool Condition) { if (Condition) DlgItems[Item].strData = std::move(std::forward<decltype(String)>(String)()); }};
```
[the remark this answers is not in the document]
MKadaner: The idea was to avoid expensive `BlobToHexString` call for `dlg_edit_search_hex` if the caller has no Hex option. If you think the optimization is not worth complexity, let me know, I'll revert it.

far/findfile.cpp
```diff
@@ -173,16 +173,34 @@ namespace
 		messages::percent
 	>;
 
-	struct FindFilesOptions
-	{
+	struct SearchParams
+	{
+		COPY_CONSTRUCTIBLE(SearchParams);
+		NONMOVABLE(SearchParams);
```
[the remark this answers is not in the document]
MKadaner: It goes together with using `SearchParams` by `const&`. Will change both.

far/findfile.cpp
```diff
@@ -1497,25 +1507,20 @@ intptr_t FindFiles::FindDlgProc(Dialog* Dlg, intptr_t Msg, intptr_t Param1, void
 				}
 				else
 				{
-					string strSearchStr;
+					string strSearchingIn{ Params.Hex ? BlobToHexString(Params.SearchBytes, 0) : Params.SearchStr };
 
-					if (!strFindStr.empty())
+					if (!strSearchingIn.empty())
```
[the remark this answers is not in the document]
MKadaner: Refactored.

far/findfile.cpp
```diff
@@ -3247,23 +3247,21 @@ FindFiles::~FindFiles()
...
 
 	m_Owner(Owner),
 	m_Autodetection(),
-	strFindStr(std::move(FindString)),
 	SearchMode(SearchMode),
 	m_FindFileArcItem(FindFileArcItem),
 	CodePage(CodePage),
 	SearchInFirst(SearchInFirst),
-	Options(Options),
+	Params(Params), // Consider: It should be OK to access m_Owner->Params directly, through a const&.
```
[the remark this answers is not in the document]
MKadaner: I guessed that.

> Assuming that Params can't change during search...

That was actually my question. 😀

OK. As far as I understand Find File async machinery, that should be the case. Let's try it and see how it crashes, then we'll know where the problem is, and I'll fix it.

far/macro.cpp
alabuzhev: > But why regexes only? How about case sensitive and fuzzy and the whole gamut of them?

Unlike case etc. `[x] regex` usually changes the whole meaning of the search string, which sort of makes it more important.

However, I cannot immediately imagine a scenario when this might be useful. Let's remove and see how it goes, we can always put it back if something breaks. Could you remove 19 too please?

far/findfile.cpp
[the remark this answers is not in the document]
alabuzhev: Please make sure that editor / viewer, opened and closed during the search, don't screw up anything.

far/stddlg.cpp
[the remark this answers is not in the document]
alabuzhev: > The idea was to avoid expensive `BlobToHexString` call

Ah, I misread it - `String` is actually a callable!
The optimization makes sense, but:
- Please rename `auto&& String` to make it more obvious that it's a callable rather than an actual string, e.g. `StringGetter` or something like that.
- `std::move(std::forward<decltype(String)>(;` is redundant, `String()` will do the same thing.
- `auto&&` can be replaced with `function_ref<string()> const` to avoid templatizing.

far/global.hpp
```diff
@@ -42,14 +42,58 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+
+	global_search_replace_params() = default;
+
+						// It is client's responsibility to maintain invariants:
+	string SearchStr;   // If Hex, should be BytesToString(SearchBytes, CodePage)
+	bytes SearchBytes;  // If not Hex, should be empty
```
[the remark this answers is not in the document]
MKadaner: :sigh: This will incur dependency of Global on code page and strmix. Even when I introduced this dependency some time ago, and in global.cpp only, I thought it was wrong, but I deferred cleaning up to this change. Now, with the templated function, the dependency would be in the header, which is even worse. Besides, there is no way to pass code page to the operator, but this is minor, of course.

On the other hand, the clients need this invariant for their own sanity, and they have all necessary dependencies anyway so not a big problem for them.

The philosophy is that the global params are just storage, they have no behavior. It is the clients that cooperate among themselves to maintain proper state. The `operator<<` is provided for convenience only, to reduce copy-paste of making repeating assignments on clients' sides.

If you think the dependency is acceptable, let me know, I'll add this logic here.

far/global.hpp
```diff
@@ -42,14 +42,58 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+	bool Hex{};
+	bool CaseSensitive{};
+	bool WholeWords{};
+	bool Regex{};
+	bool Fuzzy{};
+	bool PreserveStyle{};
+};
+
+template <typename client_params_t>
+void operator<<(global_search_replace_params& GlobalParams, const client_params_t& ClientParams)
```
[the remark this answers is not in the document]
MKadaner: Mostly because the semantics does not fit well with the conventional meaning of assignment. Another, less important consideration was to discourage accidental unintentional assignment. Maybe making it a named function is a better idea, especially if we want to pass in the code page. What do you think?

far/global.hpp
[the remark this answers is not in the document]
alabuzhev: Ah I see. Indeed, it is a hairy problem.

Conceptually, global.?pp is a workaround to address issues with initialization order of complex core objects, not a home for each and every global/static variable in the project.
Ideally we should tear it down, not build it up.
`global_search_replace_params` does not really belong here and, since it consists of simple vocabulary types, it does not need a special treatment.
We can probably move it to stddlg.?pp and do the usual thing there, e.g.:
```C++
global_search_replace_params& params()
{
    static global_search_replace_params s_Params;
    return s_Params;
}
```
It should also be possible to move the hex logic into a non-templated helper function to not pollute the header.

far/global.hpp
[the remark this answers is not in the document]
alabuzhev: I'd probably prefer a named function (save, store, persist etc.) if it does not fit well.
Overloading operators is great for vocabulary types, global_search_replace_params is unlikely one of them.

far/global.hpp
MKadaner: There are three places where we need these parameters in somewhat different capacities:

1. The clients: View, Edit, Help, Find File. Each of them needs different subset of parameters.
2. The Search / Relace dialog. It needs to have all parameters and deal with either combination of them to be optional. Currently this is modelled with `std::optional`. It can also be done with templates.
3. The Global. It heeds to have all parameters, but the save operation must not override those which do not exist in the client, so we need some way to indicate which parameters exist in this particular client, and which should be ignored (to preserve the values saved by the clients who actually have them).

So, the client's parameter subset serves as a filter between the dialog and the global. The proper design is to have three distinct types: full set of non-optional parameters in global, full set of optional params in the dialog, and proper subset of non-optional params in each of the clients. Then, we need two mappings, between client's and dialog's, and between client's and global's. The mappings should be smart enough to leave optional params in dialog's empty if there are no client's counterparts, and to not reset global's params with non-existent client's params.

I actually prototyped this approach. There are two disadvantages. First of all, it is unwieldy, lots of repeatable code even with templated everything. Second, it obviously involves unnecessary copying of values at run time.

The other extreme implementation is to use `SearchReplaceDlgParams` (properly renamed) in all three places. There are some issues here as well.
1. The use of optional in global seems to be utterly wrong. All global params always have some values (even if `false` immediately after start).
2. The use of optional in Find File is completely redundant. and will only clutter the code.
3. The clients still need to indicate which parameters are applicable, so somewhere each of them needs to list all parameters anyway.

I prototyped this approach as well. It does not feel elegant either. Let me try it once again, and let's see.

alabuzhev: Another approach is perhaps just setting the required fields explicitly?
```C++
SearchReplaceParamsCommon.Regex = LastSearchDlgParams.Regex;
SearchReplaceParamsCommon.CaseSensitive = LastSearchDlgParams.CaseSensitive;
...
```
A bit more typing, but probably more robust than smart templates. And it's not like we need to do it 100 times, only 4.

far/help.cpp
```diff
@@ -1429,7 +1433,7 @@ bool Help::ProcessKey(const Manager::Key& Key)
 			// не поганим SelTopic, если и так в FoundContents
 			if (StackData->strHelpTopic != FoundContents)
 			{
-				if (GetSearchReplaceString({}, LastSearchDlgParams, L"HelpSearch"sv, {}, CP_DEFAULT, {}, &HelpSearchId) == SearchReplaceDlgResult::Cancel)
+				if (GetSearchReplaceString({}, SearchDlgParams(), L"HelpSearch"sv, {}, CP_DEFAULT, {}, &HelpSearchId) == SearchReplaceDlgResult::Cancel)
```
MKadaner, line 1436: Can Help be opened recursively? I hope it cannot but wanted to double check. Otherwise, besides static Params, we need an instance copy of Params.

far/macro.cpp
MKadaner: Thanks. Done.

far/stddlg.cpp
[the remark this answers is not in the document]
MKadaner: Removed all this template stuff.

far/stddlg.cpp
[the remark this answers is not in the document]
MKadaner: Renamed and removed `move-forward` nonsense (do not know what happened).

`function_ref<string()>` involves type erasure and indirection, hence most likely actual call, while `auto&&` should be nicely inlined. Let's keep it this way.

far/findfile.cpp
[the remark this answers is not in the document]
MKadaner: Seems to work. More testing is needed though. See my last comment below.

far/global.hpp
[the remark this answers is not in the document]
MKadaner: > Conceptually, global.?pp is a workaround...

That changes everything! I reworked the whole as you suggested (plus a separate instance for help). Please see if you like it.

far/global.hpp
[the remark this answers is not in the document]
MKadaner: Done.

MKadaner:
I completely reworked how global params are stored and copied to / from instances of the clients.

I consider the code complete. It compiles and does not immediately crash (I tried all four components). However, I have not had time to test it at all, let alone thoroughly. Also, refactoring artifacts may still exist.

Could you please review the change conceptually and let me know if it will do. If you notice some rough edges, of course, call them out as well. I will clean up and finalize the change later, though I may have less time during the week.

far/help.cpp
alabuzhev: F1 F7 F11 F1 F7 F11 F1 F7 F11 F1 F7 F11...

alabuzhev (approved):
Looks good, thank you.

far/help.cpp
MKadaner: Ugh! Will model Help Search params after all other components'.

far/findfile.cpp
```diff
@@ -515,9 +518,9 @@ void background_searcher::InitInFileSearch()
 	readBufferA.resize(readBufferSize);
 	readBuffer.resize(readBufferSize);
 
-	if (!Options.SearchHex)
+	if (!Params.DlgParams.Hex)
```
rohitab, line 521: There appears to be a typo here, causing this code to not be executed, which results in a never ending file search in `background_searcher::LookForString`. I tested this fix and the file search appears to be working correctly now.

```suggestion
	if (!Params.DlgParams.Hex.value())
```

far/findfile.cpp
```diff
@@ -3107,8 +3102,8 @@ FindFiles::FindFiles():
 
 		FindAskDlg[FAD_EDIT_MASK].strHistory = L"Masks"sv;
 		FindAskDlg[FAD_EDIT_TEXT].strHistory = L"SearchText"sv;
-		FindAskDlg[FAD_EDIT_HEX].strMask = L"HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH HH"sv;
-		FindAskDlg[FAD_CHECKBOX_NOTCONTAINING].Selected = Options.NotContaining;
+		FindAskDlg[FAD_EDIT_HEX].strMask = HexMask(24);
```
rohitab, line 3105: Not sure if you noticed, but the original mask was `23` bytes, which I believe was done so that the content fit in the _Containing_ Hex control without scrolling. `24` bytes causes the control to scroll at the very last byte. Just thought I'd point it out.

```suggestion
		FindAskDlg[FAD_EDIT_HEX].strMask = HexMask(23);
```

far/findfile.cpp
MKadaner: Ouch! Thank you! As I said, haven't had time to test it. Of course, you are right. That's going to be a bug farm; I should think of something less error-prone.

far/findfile.cpp
MKadaner: Hm-mm... Yes, I did it intentionally. I thought that 23 was a weird number, so I adjusted it to something more "round." 24 at least is divided by 8. 🙂 The actual purpose of this number has never occurred to me. I'll change it back. Thank you.

far/help.cpp
MKadaner: Done!

far/findfile.cpp
MKadaner: I could not find a good solution which would not introduce lots of clutter in some other place(s). Leaving it as is for now (this place is fixed, of course).

far/findfile.cpp
MKadaner: Done!

MKadaner:
Ready to merge (unless more issues found).

far/stddlg.cpp
```diff
@@ -76,6 +76,69 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 //----------------------------------------------------------------------------
 
+
+namespace
+{
+	template<SearchReplaceDlgParams::SharedGroup group>
```
alabuzhev, line 82: Looks like templatizing doesn't contribute anything here except producing unique instances.
Maybe we can avoid templates completely with something like

```C++
auto& GetShared(SharedGroup const Id)
{
    static SearchReplaceDlgParams s_SharedParams[SharedGroup::count];
    return s_SharedParams[Id];
}
```
?

I normally wouldn't care since templatized parts are already minimal, but those explicit instantiations below make things look more complex than they could be.

MKadaner: Well, using templates here has some advantages:
- Logical: instantiation of the shared params of different groups are not tied together. In this case it's irrelevant, but in general we may want to avoid instantiating `help` params only because somebody needs `common` params.
- Performance: no indexing, no indirections.
- Encapsulation: no implementation details, like `count`, is exposed to the clients.

Each of these considerations are, of course, minor. However, all of them together made my choice, especially since I do not see any extra complexity in explicit template specialization.

Anyway, I change these functions to use indices.

far/stddlg.hpp
```diff
@@ -65,16 +65,30 @@ struct SearchReplaceDlgProps
...
+	std::optional<string> ReplaceStr;
 	std::optional<bool> Hex;
 	std::optional<bool> CaseSensitive;
 	std::optional<bool> WholeWords;
-	std::optional<bool> Regexp;
+	std::optional<bool> Regex;
 	std::optional<bool> Fuzzy;
 	std::optional<bool> PreserveStyle;
+
+	enum class SharedGroup { common, help, _count };
```
alabuzhev, line 77: Technically `_count` is not a reserved identifier, because it is not in the global namespace.

However, names like that give an impression that we are doing something fishy, like accessing some platform-specific bits, which is not the case. Maybe just `count`?

MKadaner: Done!

alabuzhev:
Thank you

MKadaner:
Thank you!

Issue #685, opened 2023-05-14
Search in archives option not saved in Find File search


rohitab:
### Far Manager version

3.0.6142.0 x64

### OS version

10.0.22624.1755

### Steps to reproduce

1. Press `Alt+F7` to open _Find file_
2. Press `Alt+I` to enable _Search in archives_
3. Press `Enter` to search
4. Press `Esc` to close _Find file_ results
5. Press `Alt+F7` to reopen _Find file_

### Expected behavior

_Search in archives_ option is enabled (checked).

### Actual behavior

_Search in archives_ option is disabled (unchecked).

Please note that this happens regardless of whether the _Containing_ input box has text. The other options like _Case sensitive_ and _Whole words_ are only saved if the _Containing_ input box has text, but _Search in archives_ option is never saved. This was working correctly in version `3.0.6134.0 x64`.

MKadaner:
That's weird. I do not think I touched _Search in archives_, at least not intentionally. Will have a look.

rohitab:
Thanks! By the way, this issue is also present with the _Not containing_ checkbox.

https://github.com/FarGroup/FarManager/blob/59231c2b931a2cb70f25ab75b301a65bcc323ea2/far/findfile.cpp#L191-L192

In previous versions (where it worked), these options were stored in a static variable so they were persisted across multiple searches.

https://github.com/FarGroup/FarManager/blob/d76ab4d5a6b755b22e5239de8a71532af82fb156/far/findfile.cpp#L3034-L3036

MKadaner:
Yeah! I already noticed that. Fixing it as well. Stay tuned.

MKadaner:
Fixed in #686.

rohitab:
Just tested version `3.0.6144.0 x64`. The issue is fixed now. Thank you!

Issue #681, opened 2023-05-08
Find file search text not saved when cleared


rohitab:
### Far Manager version

3.0.6136.0 x64

### OS version

10.0.22624.1690

### Steps to reproduce

1. Press `Alt+F7` to launch _Find file_
2. Press `Alt+T` to switch to _Text_ mode
3. Type in `hello` in the _Containing_ input box
4. Press `Enter` to start the search
5. Press `Esc` to close _Find file_ results
6. Press `Alt+F7` to re-launch _Find file_
7. Note that the _Containing_ input box shows the previous search `hello`
8. Press `Alt+T`, then `Del` to clear the search text
9. Press `Enter` to start the search
10. Press `Esc` to close _Find file_ results
11. Press `Alt+F7` to re-launch _Find file_

### Expected behavior

The _Containing_ input box should be empty.

### Actual behavior

The _Containing_ input box shows the previous non-empty search text `hello`.

Open the _Editor_, press `F7` to search, press `Del` to clear out the `hello` text from the _Search for_ input box, press `Enter` to search. Now close the _Editor_ and press `Alt+F7` to launch _Find file_ again. You'll notice the _Containing_ input box is now empty.

Appears to be a recent issue which is not present in version `3.0.6134.0 x64`.

rohitab:
Might be an issue with this piece of code that only saves if the string is not empty.

https://github.com/FarGroup/FarManager/blob/c085d23040482e25472d631688a84f0150d40a74/far/findfile.cpp#L3209-L3212

It seems that the Editor saves it regardless of its content.

https://github.com/FarGroup/FarManager/blob/c085d23040482e25472d631688a84f0150d40a74/far/editor.cpp#L220-L223

MKadaner:
Thank you for the report, @rohitab.

I guess I know what happened here. Originally, the similar check was probably intended to avoid modifying the global search options if the user searched for files by name without regard to the content. The thinking went along the lines of, "the user did not supply search pattern, so all search options, like case sensitive, whole words, etc. were not used and should not affect the global settings even if the user accidentally flipped some checkboxes." The behavior was as if the user cancelled the regular Search dialog.

However, before my change, the search pattern was eventually saved to the find-file-specific static buffer _unconditionally_, so the empty search string would be picked up the next time the user opens find file dialog.

I need to think how to imitate old behavior, which is probably expected, without re-introducing yet another piece of global state.

[...]

PR #686, opened 2023-05-15
Fixes of Viewer / Editor / Find File search options persisting.

MKadaner:
## Summary
- Fallout of 6135: Persist forgotten Find File dialog options.
  - `gh-681`: Persist empty search pattern.
  - `gh-685`: Persist "Search in archives" option.
  - `gh-681`: Find File content search parameters (search pattern, hex, case sensitive, whole words, fuzzy) are not shared with Viewer / Editor. However, if the search pattern is not empty, Find File content search parameters are copied to Viewer / Editor search.
- `gh-681`: Do not allow empty search pattern in Viewer and Editor search dialog.

## References
- gh-681
- gh-685

## Checklist
- [X] I have discussed this with project maintainers: `gh-681`, `gh-685`<br/>

## Details
Yet another rather big but relatively straightforward change.


MKadaner:
@rohitab, while I tested the changes locally, it would be great if you tested it as well.

rohitab:
@MKadaner I just ran some quick tests and your changes looks good. I didn't find any issues.

### File Mask

I noticed you changed some code related to the file mask, so I thought I'd point out a slight inconsistency that has been there for a while. In the _Find file_ dialog, the _file mask_ is `*.*` by default. However, if you clear it, it is set to `*` for the next search. I think it should be the same in both cases  `*.*`. This is completely unrelated, but since it is a minor change, I thought maybe you could add it in as well. If you think it makes sense, of course.

far/help.cpp
```diff
@@ -1453,7 +1453,8 @@ bool Help::ProcessKey(const Manager::Key& Key)
 			// не поганим SelTopic, если и так в FoundContents
 			if (StackData.strHelpTopic != FoundContents)
 			{
-				if (GetSearchReplaceString({}, m_SearchDlgParams, L"HelpSearch"sv, {}, CP_DEFAULT, {}, &HelpSearchId) == SearchReplaceDlgResult::Cancel)
+				if (GetSearchReplaceString({ .AllowEmptyPattern = true }, m_SearchDlgParams, L"HelpSearch"sv, {}, CP_DEFAULT, {}, &HelpSearchId)
```
alabuzhev, line 1456: Why do we allow this in Help?

rohitab: I'm guessing because an empty search displays all topics from the _Help_, not just the ones listed in _Contents_. I have used this feature in the past.

MKadaner:
Thank you for testing, @rohitab.

> File Mask

Yeah, I've seen this inconsistency in the code; did not want to touch it because any change in this area causes a real big discussion. 😀

Since you mentioned it, I decided to fix it. Now it's `*.*` everywhere.

far/help.cpp
MKadaner: Done!

MKadaner:
Merged, rebased, ready!

alabuzhev:
Thanks

MKadaner:
Thank you!

far/help.cpp
rohitab: :bulb: If anyone is interested in seeing the old format, the one that showed up when searching with an empty string, just search for `#`. You'll get the same list as before, with all the topics listed in the same order as the `.hlf` _Help_ file.

Issue #669, opened 2023-04-03
`Home` key doesn't work in Editor's Find/Replace dialogs


MKadaner:
### Far Manager version

3.0.6122.0

### OS version

10.0.22621.1485

### Other software

N/A

### Steps to reproduce

1. Open _Search_ or _Replace_ dialog in the internal Editor.
2. Move focus to one of the checkboxes (use `Tab` or a hotkey).
3. Press `Home` key.

### Expected behavior

Focus is set to the `Search for` edit control.

### Actual behavior

Focus stays where it was.

MKadaner:
Originally reported by @HamRusTal in [#668](https://github.com/FarGroup/FarManager/pull/668#issuecomment-1493322197).

Please assign to me.

MKadaner:
Fixed in 6151 (gh-688).

HamRusTal:
I suspect this change caused #697…

MKadaner:
Sigh... Reopening. See [this comment to #688](https://github.com/FarGroup/FarManager/pull/688#issuecomment-1629111037) for details.

PR #688, opened 2023-05-16
gh-669: Now `Home` key works in the Search / Replace dialog.

MKadaner:
## Summary
Added new `FARDIALOGITEMFLAGS`: `DIF_HOMEITEM`.
This flag indicates the item which should take focus on `Home` key press.
Note: The first "focusable" item is not always the best target for `Home` key.
It is a good idea to set this flag on the item which initially has `DIF_FOCUS`.

## References
- gh-669

## Checklist
- [X] I have discussed this with project maintainers: `gh-669`<br/>

## Details
Experimental. Where should I replicate the definition of `DIF_HOMEITEM`? Any documentation to update? Is it even a good idea to introduce this flag?


alabuzhev:
> Is it even a good idea to introduce this flag?

Given that
> It is a good idea to set this flag on the item which initially has `DIF_FOCUS`.

maybe the dialog could just remember the first focused item and go there on Home?

MKadaner:
> ... maybe the dialog could just remember the first focused item and go there on Home?

It will not work. For example, in the Search / Replace dialog, `DIF_FOCUS` is assigned dynamically. Say, it is initially set to the Text edit control, we snap it at the beginning, then the user switches to Hex, Text becomes hidden, and `Home` key moves focus to the Text / Hex radio.

alabuzhev:
> It will not work.

Unfortunately a new flag will not work too - at least until we and 3rdparty plugin authors add it everywhere it should be, which will probably take eons.

MKadaner:
> ... at least until we and 3rdparty plugin authors add it everywhere ...

Can you elaborate please? Everywhere where? In all dialogs, or do you mean something else? If the new flag is not set on any controls, the behavior of the `Home` key is not changed. All existing dialogs and plugins should work as they work now.

alabuzhev:
Ah, I probably misunderstood you.
The whole idea here is to fine-tune Home behavior in rather complex dialogs, which are relatively rare anyway, right?

MKadaner:
That is correct. I hope I did not break any existing dialogs. Just provided means to define home item when it is not the first on the dialog. Could you please double check the changes in `Dialog::Do_ProcessFirstCtrl()` function?

I think the new feature is very useful if `DIF_FOCUS` is initially set to an item in the middle, like it happens in the Search / Replace dialog. In such cases, the `Home` key should move focus to the initially focused item, and not to the first focusable item.

In fact, after I fixed the actual problem which made the `Home` key in the Search / Replace dialog useless, the `Home` key moved focus to the `[ Word ]` button instead of the `Search for` editor. This looked really awkward. I wonder nobody complained though in the Viewer the `Home` key worked and the behavior was similarly ugly.

alabuzhev:
Looks good.

> Is it even a good idea to introduce this flag?

On the one hand, such dialogs are rare and Home can be handled in DlgProc if needed.
On the other hand, having it won't harm, so why not.

> Any documentation to update?

Add a new entry [here](https://github.com/FarGroup/FarManager/tree/master/enc/enc_rus/meta/dialogapi/flags) please.

MKadaner:
I added encyclopedia articles. Please see if I screwed something up. Otherwise, ready to merge.

alabuzhev:
Thank you

MKadaner:
Thank you!

johnd0e:
@MKadaner 
1. `DIF_HOMEITEM` is still missing on this page https://api.farmanager.com/ru/dialogapi/index.html
2. It also should be add to each page linked from here: https://api.farmanager.com/ru/dialogapi/controls/index.html
3. English Encyclopedia is completely [unmaintained](https://github.com/FarGroup/FarManager/commits/master/enc/enc_eng). It is still about 1.75 api (I suppose), so there is no sense to update it with new features of Far 3.

Another issue: https://www.google.com/search?q=DIF_HOMEITEM+site:api.farmanager.com
Not found, despite there are 2 pages with this:
- https://api.farmanager.com/ru/dialogapi/flags/index.html
- https://api.farmanager.com/ru/dialogapi/flags/dif_homeitem.html

The same is for https://fargroup.github.io/api/ru/index.html
@alabuzhev, may be you have any idea?

MKadaner:
Thank you for checking, @johnd0e. I reopened #669.

> Another issue: https://www.google.com/search?q=DIF_HOMEITEM+site:api.farmanager.com

Bing is better: https://www.bing.com/search?q=%2bDIF_HOMEITEM+site%3aapi.farmanager.com. I guess crawling frequency for api.farmanager.com is very low because the site is updated very rarely. I submitted DSAT:

image

All we need to do is to wait for another year or so. 😀

> English Encyclopedia is completely [unmaintained](https://github.com/FarGroup/FarManager/commits/master/enc/enc_eng). It is still about 1.75 api (I suppose), so there is no sense to update it with new features of Far 3.

Maybe remove it altogether to avoid confusion? Should I revert my changes for EN?

alabuzhev:
> Maybe remove it altogether to avoid confusion? Should I revert my changes for EN?

No, we will get there one day.

PR #691, opened 2023-05-18
Fix Panel.Select

johnd0e:
Ref: https://forum.farmanager.com/viewtopic.php?p=174783#p174783


johnd0e:
@shmuz

shmuz:
Looks good but you need to build Far and test.
Here is my test (assuming TmpPanel with files as [here](https://forum.farmanager.com/viewtopic.php?p=174735&sid=842ea79c43c703007a9605e1f659befe#p174735) is on the active panel):
```lua
local a={}
for i=2,APanel.ItemCount do
  a[#a+1] = Panel.Item(0, i)
end
for k=1,3 do
  local m=k==1 and 1 or k==2 and 0 or 2 -- select, deselect, invert
  Panel.Select(0, m, 2, table.concat(a,'\n'))
  far.Text()
  win.Sleep(2000)
end
```

alabuzhev:
This is a breaking change.
Currently `Panel.Select(0, 1, 2, "file.ext")` works on TmpPanel, after the change it won't, and the documentation [explicitly allows](https://github.com/FarGroup/FarManager/blob/master/enc/enc_rus/meta/macro/macrocmd/prop_func/panels.html#L511&:~:text=возможно,путями) omitting full paths.

We should fall back to the current logic if the item is not found, something like
```DIFF
-	const auto Pos = FindFile(PointToName(i), true);
+	auto Pos = FindFile(i, false);
 	if (Pos == -1)
-		continue;
+	{
+		Pos = FindFile(PointToName(i), true);
+		if (Pos == -1)
+			continue;
+	}
```

shmuz:
> Currently Panel.Select(0, 1, 2, "file.ext") works on TmpPanel

I wouldn't call that "works" as in case there are several "file.ext" on TmpPanel only 1 of them is processed.

shmuz:
See my backward-compatibility fix: https://github.com/shmuz/far2m/commit/4067c5280a1fbb88a4859f51c4557d3024b73fb7

shmuz:
@johnd0e 
Вместо LGOOD_SLASH поставьте `L'\\'`

far/filelist.cpp
```diff
@@ -982,7 +982,7 @@ long long FileList::VMProcess(int OpCode,void *vParam,long long iParam)
 					if (i.empty())
 						continue;
 
-					const auto Pos = FindFile(PointToName(i), true);
+					const auto Pos = FindFile(i, !wcschr(namePtr,LGOOD_SLASH));
```
alabuzhev, line 985: It won't compile.
`!wcschr(namePtr,LGOOD_SLASH)` -> `!i.contains(path::separator)`.


This is still a bit incompatible: previously it was possible to select names like `"q:\non_existing_or_completely_unrelated_path\file.ext"`, as long as there is `file.ext` on the panel, but we can probably live with that.

johnd0e: > but we can probably live with that.

Hmm.. I rather see that as breaking change

far/filelist.cpp
```diff
@@ -982,7 +982,7 @@ long long FileList::VMProcess(int OpCode,void *vParam,long long iParam)
 					if (i.empty())
 						continue;
 
-					const auto Pos = FindFile(PointToName(i), true);
+					const auto Pos = FindFile(i, !i.contains(path::separator));
```
alabuzhev, line 985: Sorry, just realized this won't compile too (or at least not everywhere).

`!i.contains(path::separator)` -> `!contains(i, path::separator)`

johnd0e:
> This is still a bit incompatible: previously it was possible to select names like "q:\non_existing_or_completely_unrelated_path\file.ext", as long as there is file.ext on the panel, but we can probably live with that.

@alabuzhev 
Could you please provide sample code illustrating that incompability?
Because for me that seems still working.
```lua
lua:Panel.Select(0,1,2,"file.ext")
```

alabuzhev:
> Could you please provide sample code illustrating that incompability?

```
cd %farhome%
lua:Panel.Select(0, 1, 2, "banana\\far.exe")
```

I'm not saying it makes any sense though.
However, the documentation also mentions `Panel.Select(0,1,2,clip(0))`, so one might imagine a scenario like "select files that I copied into clipboard in some other directory".

johnd0e:
Ah, I see now.
So in my opinion this new behavior is more correct (and corresponds to docs more).

@shmuz ?

shmuz:
In my opinion the current solution is OK.

BTW, when I and **zg** were changing the macrosystem to Lua, **zg** removed all macroapi functions (he thought they were unnecessary as we had LuaFAR API). Later I put them back.

shmuz:
I made another change (not committed yet; as I'm writing the test for Panel.Select).
See the snippet below and pay attention to the new variable `RegularPanel`:
```c++
auto SelectByList = [&] (ps_action act)
{
	Result=0;
	const wchar_t *name;
	bool RegularPanel = !GetPluginHandle();
	for(size_t I = 0; (name=itemsList.Get(I)) != nullptr; ++I)
	{
		int Pos;
		if ((Pos=FindFile(name, RegularPanel || !wcschr(name,LGOOD_SLASH))) != -1)
		{
			Select(ListData[Pos], act==remove?FALSE : act==add?TRUE : !ListData[Pos]->Selected);
			Result++;
		}
	}
};
```

shmuz:
The commit is [here](https://github.com/shmuz/far2m/commit/960138bb8c50d24dc8d4b9bc978a1af650e57098). Also, a 90-line test function was added.

alabuzhev:
Thanks, synchronized.

shmuz:
> Thanks, synchronized.

I'm a little confused. There are some API differences between Far3 and far2m and such a synchronization is a ... non-trivial task.

alabuzhev:
I've picked what I could.

shmuz:
I didn't (yet) look at your commit but for example in the test added today:
far2m: `panel.GetPanelInfo(1)`
Far3: should be `panel.GetPanelInfo(nil,1)`

shmuz:
@alabuzhev 
Well done!

Issue #697, opened 2023-05-26
Help/Viewer/Editor Search/Replace “Search for” text field's completion menu disappears upon Up/Down arrows


HamRusTal:
### Far Manager version

3.0.6151.0

### OS version

10.0.19045.2604

### Steps to reproduce

1. Start Far: `far.exe -s "%TEMP%\FarProfile\2023-05-26" -p`
2. Open e.g. its `changelog` for editing.
3. Start a couple of replacements to populate the history (no need to actually replace anything): `aa` to `aa`, `ab` to `ab`.
4. Open the Replace dialog.
5. Observe the following behavior: type `a` into the `Replace with` field — the completion drop-down menu opens, press Up/Down — the menu stays, its current item follows the arrows pressed.
6. Now type `a` into the `Search for` field, then press the Up/Down arrows.

### Expected behavior

The completion drop-down menu stays, its current item follows the arrows pressed — just like in the `Replace with` field.

### Actual behavior

The completion drop-down menu disappears but the dialog behaves as if the menu is still shown: the Up/Down arrows change the current field text, `Esc` does not close the dialog.

MKadaner:
Ough! Apparently, I broke it in 6144 (gh-686). Will look during the weekend. Thank you for reporting.

MKadaner:
Fixed in 6156 (#394)

PR #698, opened 2023-05-29
gh-697: Fixed disappearing autocomplete list in Search / Replace dialog.

MKadaner:
## Summary
The bug was introduced in 6144 (gh-686).

## References
- gh-697
- gh-686

## Checklist
- [X] I have discussed this with project maintainers: gh-697<br/>

## Details
The issue happens when `DN_EDITCHANGE` comes because of changing items in the autocomplete list (Up, Down, Mouse), and we send `DM_ENABLE` to dialog buttons. In this case, the dialog is refreshed but the autocomplete list is not, so the list disappears from the screen but still has focus. It seems to be a bug in the autocomplete list logic.

The proposed change is not really a fix, rather dodging the problem by avoiding sending unnecessary `DM_ENABLE`s when the "empty" status did not change (which is the right thing to do anyway). It works because while the autocomplete list is open, the Search string never becomes empty, so we never send `DM_ENABLE`.


MKadaner:
@alabuzhev and everyone. I could not, in reasonable time, find why the autocomplete list is not refreshed in the scenario described above. If somebody can give me a hint how to fix the problem properly, I'll be happy to do it.

alabuzhev:
Thanks.

> why the autocomplete list is not refreshed in the scenario described above

Autocomplete list is a bit hacky and might behave unexpectedly in corner cases.
Does it happen elsewhere?

MKadaner:
Thank you.

> Does it happen elsewhere?

Not that I noticed. On the other hand, This scenario is new. If, as you said elsewhere, dialogs never update other controls in response to `DN_EDITCHANGE` sent for an edit box with autocomplete enabled, this issue could not happen. If, as you suggested at the same place, we want to introduce this behavior in other dialogs, we should better fix autocomplete list.

johnd0e:
> Autocomplete list is a bit hacky and might behave unexpectedly in corner cases. Does it happen elsewhere?

Actually it does. On DN_EDITCHANGE my DlgProc sends DM_SETTEXT (to other unrelated control), and Autocomplete list disappears.
I have tried to force redraw list with DM_REDRAW but no luck.
ACTL_REDRAWALL does not work either.

The only workaround I found is changing list size, forth and back. Ugly but works.
```lua
  local wi = far.AdvControl(F.ACTL_GETWINDOWINFO)
  if wi.Type==F.WTYPE_VMENU then
    local hList = wi.Id
    local r = hList:send(F.DM_GETDLGRECT)
    local w = r.Right - r.Left + 1
    local h = r.Bottom - r.Top + 1
    hList:send(F.DM_RESIZEDIALOG,0,{X=w,Y=h+1})
    hList:send(F.DM_RESIZEDIALOG,0,{X=w,Y=h})
  end
```

johnd0e:
Here is sample code to reproduce the bug:
```lua
local F = far.Flags
far.Dialog("",-1,-1,20,1,nil,{
  {F.DI_TEXT,0,0,1,0,0},
  {F.DI_EDIT,3,0,20,0,0,0,"test",F.DIF_EDITPATH+F.DIF_EDITPATHEXEC,""}
},nil,function(hDlg,msg)
  if msg==F.DN_EDITCHANGE then
    hDlg:send(F.DM_SETTEXT,1,math.random(9))
  end
end)
```

MKadaner:
@johnd0e, could you create an issue for this glitch? I promise nothing here, but I am poking around this place now (unfortunately, I cannot devote enough time to it), will keep it in mind.

PR #699, opened 2023-06-05
More ways to scroll VMenu lists horizontally.

MKadaner:
## Summary
If the lines in the Find All list are long (think of thousands of characters), it is tedious to shift them horizontally with `Alt+Arrow` one character at a time. In this PR, I am adding new accelerator keys to all Far lists (`VMenu`):
- `Ctrl+Alt+Arrows` shift all items by 20 characters.
- `Ctrl+Shift+Arrows` shift the selected item by 20 characters.

Also, added accelerator keys to align the selected item:
- `Alt+Shift+Home` aligns the selected item to the left.
- `Alt+Shift+End` aligns the selected item to the right.

Also, described horizontal positioning of list items in the `@MenuCmd` help article.

## Checklist
- [ ] I have discussed this with project maintainers: <br/>

## Details
For the reference, existing accelerator keys were:
- `Alt+Arrows` shift all items by 1 character.
- `Alt+Shift+Arrows` shift current item by 1 character.
- `Alt+Home` aligns all items to the left.
- `Alt+End` aligns all items to the right.


johnd0e:
> it is tedious to shift them horizontally with Alt+Arrow one character at a time.

But it is easy to write macro with same functionality.
And everyone can adapt assignments to own preferences.

MKadaner:
@johnd0e, how can I limit the scope of the macro to Find All list only (or at least to any VMenu)? I do not want to override these hotkeys in Editor or anywhere else.

There is also performance consideration. Just today, I was working with a Find All list with some 250K items, each several hundred characters long. With the macro, a shift by 20 characters will be executed as 20 shifts by one character. If window refresh is suppressed, it should be faster than 20 manual shifts, but still somewhat slow. If the operation is implemented natively, it is as fast as a one-character shift.

johnd0e:
>  how can I limit the scope of the macro to Find All list only (or at least to any VMenu)?

```lua
Macro { description="Scroll menu items by 20 chars";
  area="Menu"; key="CtrlAltLeft CtrlAltRight";
  action=function()
    local key = mf.akey(1,1):match"Left$" and "Left" or "Right"
    Keys("20*Alt"..key)
  end;
}
```

It is possible to further limit the scope to particular menu, but why.

> There is also performance consideration.  I was working with a Find All list with some 250K items

That is very uncomfortable to work with so many items, even Up/Down is slow. And 20*scroll is terrible slow.
But again: instead of hardcoding this functionality to this single menu, why not to propose general plugins API?
Like DM_LISTSCROLL or DM_LISTGET/SETHPOS.

MKadaner:
Thank you for the macro, I'll give it a try tomorrow and report performance.

> It is possible to further limit the scope to particular menu, but why.

I asked in case somebody considers it should not affect other menus; personally I would be fine with all menus.

> That is very uncomfortable to work with so many items,

I do it at least weekly, often daily. It is actually not bad (except for one-char H-scroll), at least on my work machine (which is far from high-end, by the way). In fact, I have other improvements to Find All list in mind, like `Ctrl+^` to align all found strings vertically. In due time I'll open a discussion on this and maybe some other features.

> [I]nstead of hardcoding this functionality to this single menu

For the record, my proposal here is for all `VMenu`s, which of course makes perfect sense.

> [W]hy not to propose general plugins API?

Well, to be honest, I do not believe in macros. Not sure I have good reasons, besides performance maybe (which is obviously not in issue with the APIs you suggested). Somehow, I cannot bring myself to perceive macros as first-class citizens, just like I consider byte-code-based languages inferior to proper compiled ones (even though I think C# second-best). 😀

If you strongly oppose implementing this feature natively, could you please give me some hints on how to implement a new plugin API? A model PR would be the simplest. Something on the lines of "check out such and such token in so-and-so files" will also do.

MKadaner:
> But I still believe that having API for this is more powerful.

To me, exposing an API with the sole purpose of calling it from a single dedicated macro is overkill. I do not see how this API can be useful to any plugins beyond giving the user a way to scroll menus horizontally. Well, people may want to adjust scroll size maybe... @alabuzhev may I ask your opinion here?

> I would start from looking how similar API is implemented...

Thank you, I will have a look, probably on the weekend. One way or the other, I need this feature, and I can live with the API + macro solution, though the change proposed here is probably an order of magnitude simpler to implement. 🙂

P.S. As promised, I measured performance of the macro which sends multiple keypresses. Nothing surprising, of course.

I had a menu of 248120 items, each about 300 characters, includes 6 tab characters). Pressing `Alt+Arrow` once takes on average about 75ms. The rate is comparable to the normal keyboard key repeat, so the result, visually, appears "normal." As expected, sending 20 `Alt+Arrow`s at a time, takes on average 1500ms. The only difference is that _the screen is frozen for the whole 1.5 seconds_, which is absolutely unacceptable.

MKadaner:
So, I hear implicit request to implement the API for all listboxes, not only for menu. That probably makes sense.

When I said about the limited usefulness of the API, I meant end-user scenarios. Of course, plugins / macros can do lots of interesting things, but let's consider the user's perspective. Are there any other menus or listboxes, native or plugins-owned, which will benefit from such functionality? What other lists are composed from many dynamically added, potentially very wide items? If they exist (or at least envisioned), it may warrant for the API.

MKadaner:
I checked out existing plugin APIs for listboxes. Just like there is no API for scrolling listbox vertically, the API for scrolling items horizontally is unnecessary. Scrolling of a listbox (either vertically or horizontally) does not affect any meaningful property of the UI control. Conceptually, all listbox items are fully visible to the user at all times. The fact that some (parts) of list items do not fit into the viewport, is incidental. The user is provided with the controls to compensate for the limited size of the viewport, but it does not mean that these controls should be customizable.

To compare, the scrolling behavior is very different from, say, `DM_LISTSETCURPOS`. With `DM_LISTSETCURPOS`, a plugin can select a listbox item based on a domain-specific criteria or user gesture, while scrolling of a listbox is a generic feature. Putting it the other way around, unlike getting or setting the selected listbox item, the leftmost visible character position of a listbox item has no semantical meaning. It is more like the topmost visible listbox item which also has no dedicated API.

johnd0e:
> there is no API for scrolling listbox vertically

See `TopPos`
https://api.farmanager.com/ru/structures/farlistpos.html

MKadaner:
> See `TopPos`

And may I ask who is using `TopPos` for anything other than implementing traditional list scrolling and support other operations on list item internally?

Once again, are there any user-level scenarios other than list scrolling which is implemented internally anyway?

johnd0e:
Well, I can understand your point. But for me there are a lot of cases, where I use api for such user-level things, where are just no other options. E.g. scrolling in editor, and in panels - there are not always dedicated keys for everything. But thanks to api (and macros) - I have the possibility to implement everything myself, in that or another way.

So if you have strong opinion opposite using plugins api for such kind of purposes - I have alternative solution in mind. What do you think about doing this in macro api? It is user-level thing by definition. If you need example, look for `Far.Scroll_Window` function: it is meant for changing console viewport position (quite similar purpose, isn't it?)
And it was introduced exactly for same reasons I told you before: best flexibility instead of dozen of predefined (and potentially conflicting) keyboard shortcuts.

MKadaner:
Plugin API probably makes more sense than macro API, isn't it?

OK. I'll think of it.

johnd0e:
> Plugin API probably makes more sense than macro API

Pro: Plugins can this api
Contra: Macro api is easier for user.

Compare:
`Menu.Scroll(n,...)`, where additional arg can define whether scroll the whole menu, or current line, or some other.

MKadaner:
Should we have both? 🙂

For plugin API I thought of these three:
```
DM_LISTHSCROLL: // Param1=ID Param2=Shift (for each Item, Item.ShowPos += Shift)
DM_LISTGETITEMHPOS: // Param1=ID Param2=Index Ret: ActualShowPos
DM_LISTSETITEMHPOS: // Param1=ID Param2=FarListPos: Index, ShowPos (<0 - rightmost visible counting from the end) Ret: ActualShowPos
```

How should it be defined in terms of macro API?

Can you help me with sample test macros, both Lua and regular Far macro?

HamRusTal:
> we have weird keys mapped to niche actions all over the place

Non-uniform key mappings (e.g. `Ctrl+Enter` in Shell vs `Shift+Enter` in dialogs' text edit fields, `Ctrl+M` in Shell vs `Ctrl+U` in Editor) is one of the worst parts of FAR's UX. I beg that we at least don't make it even worse.

MKadaner:
OK. Let's try once again.

- If somebody do not like these key assignments, they can change them with a macro.
- If somebody suggests universally better assignment, I will be happy to change the code.

johnd0e:
@alabuzhev 
What do you think about macro function to scroll list items?

alabuzhev:
@johnd0e I believe I have already expressed my opinion above, but I can reiterate:
- If there are signs of high demand for that such a function, e.g.
  - people ask for it
  - people come up with ugly workarounds to emulate it
  - people create PRs/fork Far to implement it
then we probably should consider adding it.
For example, see the recent "folder changed" event.
- If there are no such signs, we probably should wait for them first.

> *pluralitas non est ponenda sine necessitate*

johnd0e:
Ok, what I've really missed was per-item scrolling. Shame on me - it is already implemented via Alt+Shift+Arrows.
(It is not yet in help)

far/FarEng.hlf.m4
```diff
@@ -3416,6 +3416,18 @@ $ #Editor: All matching entries menu#
  #LeftCtrl+0…9#
  Go to the bookmark 0…9.
 
+ #Alt+Left#, #Alt+Right#
+ Scroll all items horizontally.
+
+ #Alt+Shif+Left#, #Alt+Shif+Right#
```
johnd0e, line 3422: Shif**t**, here and there

johnd0e (changes requested):
Should be implemented for every menu/listbox and documented in `@MenuCmd`.

One more thing to document - Alt+Home/End.

far/FarEng.hlf.m4
MKadaner: Ouch! Thank you for pointing this out, @johnd0e.

I moved the description to `@MenuCmd`, added description of the existing keys, and of course fixed the typo.

MKadaner:
All comments addressed. Ready for review / merge.

MKadaner:
Rebased, resolved conflicts. Ready for review / merge.

far/changelog
```diff
@@ -1,3 +1,12 @@
+--------------------------------------------------------------------------------
+MZK 2023-06-20 19:53:29-07:00 - build 6167
+
+1. More ways to scroll horizontally Find All list items in the internal Editor.
```
alabuzhev, line 4: > Find All list items in the internal Editor

Why so precise?

far/FarRus.hlf.m4
```diff
@@ -325,8 +325,35 @@ $ #Клавиатурные команды#
 $ #Команды управления меню#
  #Общие команды для меню и выпадающих списков#
 
- Фильтровать элементы меню или списка               #Ctrl+Alt+F,RAlt#
- Зафиксировать фильтр                                    #Ctrl+Alt+L#
+ #Ctrl+Alt+F,RAlt#
+ Фильтровать элементы меню или списка
```
johnd0e, line 329: Некоторые предложения завершаются точкой, некоторые нет. Желательно привести к одному виду.

far/FarRus.hlf.m4
```diff
@@ -325,8 +325,35 @@ $ #Клавиатурные команды#
...
+ Зафиксировать фильтр
+
+ #Alt+Left#, #Alt+Right#
+ Прокручивать все элементы горизонтально.
+
+ #Alt+Shift+Left#, #Alt+Shift+Right#
+ Прокручивать выбранный элемент горизонтально.
+
+ #Ctrl+Alt+Left#, #Ctrl+Alt+Right#
+ Прокручивать все элементы горизонтально на 20 позиций.
```
johnd0e, line 341: Возможно здесь уместнее употребить перфект "прокрутить", поскольку чётко сказано на сколько позиций.

far/FarRus.hlf.m4
```diff
@@ -325,8 +325,35 @@ $ #Клавиатурные команды#
...
+
+ #Ctrl+Shift+Left#, #Ctrl+Shift+Right#
+ Прокручивать выбранный элемент горизонтально на 20 позиций.
+
+ #Alt+Home#
+ Выравнять все элементы налево.
```
johnd0e, line 347: "Выравнять налево/направо" мне кажется звучит как-то не очень по-русски.
Но лучший вариант в голову не приходит.

MKadaner: Есть ещё типографский термин [выключка](https://ru.wikipedia.org/wiki/%D0%92%D1%8B%D0%BA%D0%BB%D1%8E%D1%87%D0%BA%D0%B0). Не уверен, что это будет лучше. 🙂

far/changelog
MKadaner: If you mean why it is so narrowly limited to Find All list only, it's because most all other lists do not need this feature. I wanted to fix Find All. Actual implementation is in `VMenu`, so I added new keys there, but I still think of the feature as of Find-All-specific. Historical, you know.

Updated everything. Thanks.

far/FarRus.hlf.m4
MKadaner: My bad. Fixed. Thanks.

far/FarRus.hlf.m4
MKadaner: Yeah! Maybe... Changed to "Сдвинуть". Sounds better to me.

far/changelog
alabuzhev: I mean the implementation is generic and works everywhere, so saying that "it's find all specifically" can be confusing and just "menu items" (or "menu/listbox items" even) should cause less questions.

Thank you.

far/FarRus.hlf.m4
HamRusTal: > "Выравнять налево/направо" мне кажется звучит как-то не очень по-русски.

Может, «прокрутить все элементы в начало» (и аналогично «… в конец»)?

MKadaner: Maybe «Сдвинуть все элементы в начало / в конец»? Just like with scrolling in the final version.

@HamRusTal, please let me know, I'll push a separate PR for that.

MKadaner:
Thank you all.

far/FarRus.hlf.m4
HamRusTal: > Maybe «Сдвинуть все элементы в начало / в конец»?

У меня такая формулировка вызывает какие-то неправильные ассоциации, как будто что-то _небольшое_ двигается туда-сюда. В моём представлении мы однозначно оперируем _прокруткой_, а не сдвигом или выравниванием. Но всё же эти формулировки для меня не являются принципиальными, т.ч. не настаиваю.

PR #706, opened 2023-06-15
Allow to break directory scan started from a macro (Mantis#0002692).

johnd0e:
Fix by Shmuel: https://github.com/shmuz/far2m/issues/36


johnd0e:
@shmuz

far/dirinfo.cpp
```diff
@@ -111,40 +111,44 @@ int GetDirInfo(string_view const DirName, DirInfoData& Data, multifilter* Filter
...
+				return 0;
+
+			case KEY_LEFT:
+			case KEY_RIGHT:
+			case KEY_DOWN:
+			case KEY_UP:
+			case KEY_PGDN:
+			case KEY_PGUP:
+			case KEY_HOME:
+			case KEY_END:
```
alabuzhev, line 143: The idea of "enhbreak" is to cancel the scan when you move around the file panel with quick view open.

And there are way more ways to move around the file panel than these. For example:
- numpad keys
- mouse clicks (all 3)
- mouse wheel
- Alt+letter
- Folder shortcuts
- ...

On top of that, currently any command line input (i.e. any printable key) also cancels the scan.

johnd0e:
@alabuzhev 
But what is wrong with **nmake_clang_x64_\*** builds?

alabuzhev:
> But what is wrong with **nmake_clang_x64_*** builds?

Looks like they updated VS in the build image and something is no longer compatible.
Don't worry about it, it's not related to your changes in any way.

far/dirinfo.cpp
johnd0e: Reverted

far/dirinfo.cpp
```diff
@@ -111,40 +111,44 @@ int GetDirInfo(string_view const DirName, DirInfoData& Data, multifilter* Filter
 	// Временные хранилища имён каталогов
 	string strLastDirName;
 	os::fs::find_data FindData;
+
+	// Mantis#0002692
+	struct SuspendMacros
+	{
+		SuspendMacros() { Global->CtrlObject->Macro.SuspendMacros(true); }
+		~SuspendMacros() { Global->CtrlObject->Macro.SuspendMacros(false); }
+	} Suspend;
```
alabuzhev, line 120: Minor: let's do it like in the other place (purely for consistency):
https://github.com/FarGroup/FarManager/blob/0b3d91dab6263574048368cb13f2f934f3521e0f/far/filelist.cpp#L7063-L7064

far/dirinfo.cpp
```diff
@@ -111,40 +111,44 @@ int GetDirInfo(string_view const DirName, DirInfoData& Data, multifilter* Filter
...
+
 	while (ScTree.GetNextName(FindData,strFullName))
 	{
-		// Mantis#0002692
-		if (!Global->CtrlObject->Macro.IsExecuting())
-		{
-			INPUT_RECORD rec;
+		INPUT_RECORD rec;
```
alabuzhev, line 124: Can we keep the extra indentation of this block just to reduce the diff size and make it more obvious what exactly has been changed?

We can fix source formatting in one go later (we need to do it anyway).

far/dirinfo.cpp
```diff
@@ -111,40 +111,44 @@ int GetDirInfo(string_view const DirName, DirInfoData& Data, multifilter* Filter
...
+	struct SuspendMacros
+	{
+		SuspendMacros() { Global->CtrlObject->Macro.SuspendMacros(true); }
+		~SuspendMacros() { Global->CtrlObject->Macro.SuspendMacros(false); }
+	} Suspend;
+
 	while (ScTree.GetNextName(FindData,strFullName))
 	{
-		// Mantis#0002692
-		if (!Global->CtrlObject->Macro.IsExecuting())
```
alabuzhev, line 117: As far as I understand, this has been already attempted in f7ae5b23d1864bcb9e22c33112507a4e6a199b23 and reverted in 2f8d7e7a3585b41417b072990a7423bf746c3139 due to issues, discussed in [M#2692](https://bugs.farmanager.com/view.php?id=2692).
Are we positive that suppressing macros fixes those issues?

johnd0e: You can test it yourself.
1. Place cursor at some pretty large directory
2. Execute `lua:Keys"EnOut F3 F1"`
3. Wait a couple of seconds, then press `Esc´
   Note 2 things:
   - Operation is successfully interrupted
   - Help screen appears only after operation is finished (or interrupted), and not immediately, like was in f7ae5b23d1864bcb9e22c33112507a4e6a199b23

johnd0e: @alabuzhev

far/dirinfo.cpp
alabuzhev: I mean we can fix it for all the repo later in one commit.
It's ok to leave it as is now, i.e. without 0c189c91a58389591d62787353fba2927d28cab1.

johnd0e: @alabuzhev 
Done

alabuzhev:
Thank you

PR #711, opened 2023-07-05
Added `AlignAnnotations` to `VMemu`.

MKadaner:
## Summary
This draft PR proposes a new feature: vertically aligning found patterns in the _Find All_ list. Now `VMenu` will align items' annotations vertically on pressing caret character (`^`).

I, for one, very much need this feature to visually compare the near contexts of the found patterns in a large file. I am especially missing this feature when I analyze large log files and want to find something unusual in the list of otherwise similar log lines.

The new behavior can be experimented with on my dev branch. Any feedback is highly appreciated.

## Checklist
- [ ] I have followed the [contributing guidelines](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTING.md).

## Details
I have a few questions right away:
- The behavior of scrolling items horizontally (`Alt+Arrow`, `Alt+Shift+Arrow`) changed. Currently, short items do not scroll at all, and long items stick to the list's vertical edges, so that there is never "empty" space between the edge and the item text. With this PR, items can be scrolled far beyond the edges until one last character is visible. This behavior is very natural and desirable when scrolling horizontally the list with aligned annotations. However, it looks very strange on regular menus. Does it make sense to tune this behavior? Maybe add a new `VMENU_FLAG`?
- For now, to experiment with this new feature, I used caret character (`^`) as a shortcut key. Apparently, it is not very convenient to type (long hand moving). I'd prefer using `Numpad5` or maybe `Ctrl+Numpad5` but for some reason `Numpad5` is not passed to `VMenu` from [`Dialog::ProcessKey`](https://github.com/FarGroup/FarManager/blob/d18ddaf2575849426f70e2f965266d847a8113e9/far/dialog.cpp#L2531-L2548). Are there any reasons for this limitation? Would it be OK to enable `Numpad5` for `VMenu`?
- Actual code changes in this PR are pretty hacky and for demonstration purposes only. I'd like to seriously rewrite [`VMenu::ShowMenu`](https://github.com/FarGroup/FarManager/blob/d18ddaf2575849426f70e2f965266d847a8113e9/far/vmenu.cpp#L2209-L2363). Are there any reasons to print text piecemeal changing color as we go, like [here](https://github.com/FarGroup/FarManager/blob/d18ddaf2575849426f70e2f965266d847a8113e9/far/vmenu.cpp#L2294-L2312)? The code will become much simpler if we first print the entire line in normal color, and then print hotkeys and annotations over while setting the print position as necessary. It does not look like the function `GotoXY` is expensive. I can see that `ScreenBuf::Write` takes a lock, but with this approach we will have less calls to `ScreenBuf::Write`. What do you think?
- In this PR, the new command vertically aligns the first annotation. What should be done with subsequent ones? Currently the only place where annotations are used is the _Find All_ list, so the question is moot. However, in general something should probably be done. Any thoughts?

### DIFF
--- a/far/dialog.cpp
+++ b/far/dialog.cpp
@@ -2544,6 +2544,7 @@ bool Dialog::ProcessKey(const Manager::Key& Key)
 			case KEY_MSWHEEL_RIGHT:
 			case KEY_NUMENTER:
 			case KEY_ENTER:
+			case KEY_NUMPAD5:
 				auto& List = Items[m_FocusPos].ListPtr;
 				int CurListPos=List->GetSelectPos();
 				List->ProcessKey(Key);
--- a/far/editor.cpp
+++ b/far/editor.cpp
@@ -3773,10 +3773,10 @@ void Editor::DoSearchReplace(const SearchReplaceDisposition Disposition)
 
 				case KEY_CTRLENTER:
 				case KEY_RCTRLENTER:
-				// TODO: Need to handle mouse events with ProcessMouse(). This implementation is accpetable,
+				// TODO: Need to handle mouse events with ProcessMouse(). This implementation is acceptable,
 				// but a click on the border jumps to the currently selected item without changing the item.
 				// Normally, mouse click on the border frame does nothing. Mouse click on the border
-				// ouside of the frame starts dragging the dialog.
+				// outside of the frame starts dragging the dialog.
 				case KEY_CTRL|KEY_MSLCLICK:
 				case KEY_RCTRL|KEY_MSLCLICK:
 					{
--- a/far/vmenu.cpp
+++ b/far/vmenu.cpp
@@ -435,7 +435,7 @@ int VMenu::AddItem(MenuItemEx&& NewItem,int PosAdd)
 
 	NewMenuItem.AutoHotkey = {};
 	NewMenuItem.AutoHotkeyPos = 0;
-	NewMenuItem.ShowPos = 0;
+	NewMenuItem.HorizontalPosition = 0;
 
 	if (PosAdd <= SelectPos)
 		SelectPos++;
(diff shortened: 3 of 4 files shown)


HamRusTal:
Caret as a shortcut is far from ideal because it's a regular character one might wish to use to further filter the menu contents using the built-in menu filtering functionality. I suggest `Ctrl+Shift+A` or `Ctrl+Alt+A` (from “Align”).
Please note that shortcuts with single `Alt` are not good too because they are normally used to activate menu items' hotkeys.

alabuzhev:
I wanted to do something like this back in 2012. The only reason I didn't was my laziness.

General thoughts:
- I don't think we need to do it on a hotkey, aligning by default and unconditionally is good enough.
- If we align, this menu won't need horizontal scrolling: showing a limited number of characters to the left and to the right is good enough.
- This doesn't really belong in VMenu: annotations is a private feature that is not used anywhere else and is only housed in VMenu for convenience. Editor could simply supply a list of prealigned string fragments.

MKadaner:
> Caret as a shortcut is far from ideal...

Sure! As I said, it's for experimenting only. But how about `Ctrl+Numpad5`? To me, it feels like this operation somewhat resembles centering panels.

@alabuzhev, I strongly disagree. I use horizontal scrolling in the _Find All_ list left and right. This is why I so strongly insisted on adding fast scrolling. Once again, my most important use case is filtering tall and wide log files and analyzing the result. I eventually want something like [TextAnalysisTool.NET](https://textanalysistool.github.io/) in Far (the tool is horribly glitchy and not supported anymore). Granted, full functionality is better to implement as a plugin. However, I am not sure if I will ever have enough drive for that, while the _Find All_ list is almost enough. The only two things I miss are this aligning command and showing each found line exactly once with multiple annotations. Both are easy to implement. Well, maybe saving the filtered lines to a file, but that is probably unrelated to `VMenu`.

Aligning by default on open is an interesting idea. I'll consider it.

MKadaner:
Closing in favor of #789.

PR #723, opened 2023-08-23
Using reserve for optimize inserts and minor fixes

GermanAizek:
## Summary
Useful changes, additional insert were found at `emplace_back`, but memory didnt allocated for last insert after loop.

### DIFF
--- a/plugins/arclite/plugin.cpp
+++ b/plugins/arclite/plugin.cpp
@@ -702,8 +702,10 @@ public:
       });
       const auto error_log = std::make_shared<ErrorLog>();
       size_t im = 0, nm = matched_indices.size();
+      std::vector<UInt32> indices;
+      indices.reserve(nm);
       while (im < nm) {
-        std::vector<UInt32> indices;
+        indices.clear();
         auto parent = archive->file_list[matched_indices[im]].parent;
         while (im < nm && archive->file_list[matched_indices[im]].parent == parent)
           indices.push_back(matched_indices[im++]);
--- a/plugins/arclite/sfx.cpp
+++ b/plugins/arclite/sfx.cpp
@@ -352,7 +352,7 @@ public:
   bool show() {
     label(Far::get_msg(MSG_SFX_OPTIONS_DLG_PROFILE));
     std::vector<std::wstring> profile_names;
-    profile_names.reserve(profiles.size());
+    profile_names.reserve(profiles.size() + 1);
     for (unsigned i = 0; i < profiles.size(); i++) {
       profile_names.push_back(profiles[i].name);
     }
--- a/plugins/arclite/ui.cpp
+++ b/plugins/arclite/ui.cpp
@@ -1077,6 +1077,7 @@ private:
   void populate_profile_list() {
     DisableEvents de(*this);
     std::vector<FarListItem> fl_items;
+    fl_items.reserve(profiles.size() + 1);
     FarListItem fl_item{};
     for (unsigned i = 0; i < profiles.size(); i++) {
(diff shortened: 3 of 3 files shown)


plugins/arclite/archive.hpp
```diff
@@ -169,7 +169,7 @@ class ArcAPI {
   static const ArcCodecs& codecs() {
     return get()->arc_codecs;
   }
-  static size_t Count7zCodecs() {
+  static const size_t Count7zCodecs() {
```
alabuzhev, line 172: const is useless on return types.

plugins/arclite/plugin.cpp
```diff
@@ -702,8 +702,9 @@ class Plugin
       });
       const auto error_log = std::make_shared<ErrorLog>();
       size_t im = 0, nm = matched_indices.size();
+	  std::vector<UInt32> indices;
```
alabuzhev, line 705: This changes the semantics: previously on each iteration the vector was recreated, now it's outside the loop and will accumulate all the values.
If you want to optimize here, add `indices.clear()` to the loop.

plugins/arclite/archive.hpp
GermanAizek: fix it here https://github.com/FarGroup/FarManager/pull/723/commits/d832463aa807c429ca15719bcc7cb5939f306fbd

plugins/arclite/plugin.cpp
GermanAizek: u ar absolutely right, I have changed original behavior function.
fix it here https://github.com/FarGroup/FarManager/commit/d832463aa807c429ca15719bcc7cb5939f306fbd

GermanAizek:
@alabuzhev,
I fixed code tabulation by making it like in your other project sources.

plugins/arclite/archive.cpp
```diff
@@ -233,6 +233,7 @@ class MyCompressCodecsInfo : public ICompressCodecsInfo, public IHashers, privat
       }
       if (arc_lib.ComHashers) {
         UInt32 numHashers = arc_lib.ComHashers->GetNumHashers();
+        hashers_.reserve(numHashers);
```
alabuzhev, line 236: Thinking about it, all this code is a part of a bigger loop, and calling `reserve` in a loop is a known antipattern: it is too easy to pessimise exponential growth down to linear.

I suggest leaving it as is, it's not like like this place is a bottleneck anyway.

plugins/arclite/archive.cpp
```diff
@@ -400,6 +401,7 @@ void ArcAPI::load_codecs(const std::wstring& path) {
   const auto& add_hashers = [this](ArcLib &arc_lib, size_t lib_index) {
     if (arc_lib.ComHashers) {
       UInt32 numHashers = arc_lib.ComHashers->GetNumHashers();
+      arc_hashers.reserve(numHashers);
```
alabuzhev, line 404: Same as above: `add_hashers` is calling from a `for` loop below, and then again from a `while` loop.
`reserve`ing here would effectively kill exponential growth,

plugins/arclite/plugin.cpp
```diff
@@ -702,8 +702,11 @@ class Plugin
       });
       const auto error_log = std::make_shared<ErrorLog>();
       size_t im = 0, nm = matched_indices.size();
+      std::vector<UInt32> indices;
+      indices.reserve(nm);
       while (im < nm) {
-        std::vector<UInt32> indices;
+        if (!indices.empty())
```
alabuzhev, line 708: This condition is redundant, `clear` checks it internally.

PR #729, opened 2023-09-14
LuaFAR: update luautf8 library to 0.1.5-2-g182575d

johnd0e:
Unicode normalize algorithm implemented.

New functions:
- `utf8.isnfc(s) -> boolean`
  check whether `s` is in *Normal Form C* or not.
- `utf8.normalize_nfc(s) -> normal_string, was_nfc`
  convert `s` to *Normal Form C*.

See https://github.com/starwing/luautf8/pull/44/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 for more info.


johnd0e:
@shmuz

shmuz:
If `luamacro\macrotest.lua` passes then OK.

johnd0e:
@alabuzhev 
Should I forward sonarcloud's report to upstream?

alabuzhev:
> Should I forward sonarcloud's report to upstream?

No need to, looks like sonar is simply not aware that lua_error doesn't return (which is not surprising, since it is not marked as such).

alabuzhev:
There are more pressing issues - as you can see, it doesn't compile.
Probably we need to suppress a few warnings to unblock it.

johnd0e:
> Probably we need to suppress a few warnings to unblock it.

Perhaps better option would be send our "fixes" to upstream?

P.S.
Where these warnings enabled?
I've just create new VC project from scratch - and the code compiles without errors.

alabuzhev:
> Perhaps better option would be send our "fixes" to upstream?

You are very welcome to do so if you like, but I'm not going to come up with such fixes.

We use somewhat paranoid warning options across the project (because we can) and try to keep it warning-free, but for third party code this is usually a futile idea.

<details>
  <summary>This should fix the build</summary>
  
```diff
--- a/plugins/luamacro/luafar3.vcxproj
+++ b/plugins/luamacro/luafar3.vcxproj
@@ -45,8 +45,14 @@
     <ClCompile Include="luafar\lusercontrol.c" />
     <ClCompile Include="luafar\reg.c" />
     <ClCompile Include="luafar\service.c" />
-    <ClCompile Include="luafar\slnunico.c" />
-    <ClCompile Include="luafar\lutf8lib.c" />
+    <ClCompile Include="luafar\slnunico.c">
+      <WarningLevel>Level1</WarningLevel>
+      <SDLCheck>false</SDLCheck>
+    </ClCompile>
+    <ClCompile Include="luafar\lutf8lib.c">
+      <WarningLevel>Level1</WarningLevel>
+      <SDLCheck>false</SDLCheck>
+    </ClCompile>
     <ClCompile Include="luafar\uliolib51.c" />
     <ClCompile Include="luafar\uloadlib51.c" />
     <ClCompile Include="luafar\ustring.c" />
--- a/plugins/luamacro/makefile_lib_gcc
+++ b/plugins/luamacro/makefile_lib_gcc
@@ -39,6 +39,7 @@ USERC = \
 	-Wno-format-nonliteral \
 	-Wno-implicit-fallthrough \
 	-Wno-missing-declarations \
+	-Wno-pedantic \
 
 ifdef CLANG
 USERC += \
@@ -50,7 +51,6 @@ USERC += \
 	-Wno-format-non-iso \
 	-Wno-missing-prototypes \
 	-Wno-missing-variable-declarations \
-	-Wno-pedantic \
 	-Wno-shadow \
 	-Wno-strict-prototypes \
 	-Wno-unused-macros \
```
</details>

PR #730, opened 2023-09-14
Fixed bug with getting struct size PluginInfo using sizeof()

GermanAizek:
## Summary

Looking at name StructSize class member, it was implied by size structure itself, not pointer.

 
## References

https://www.codeproject.com/Questions/1249365/How-can-I-check-size-of-pointer-variable-in-Cplusp


alabuzhev:
Thanks

PR #708, opened 2023-06-20
Fix broken links in Encyclopedia

johnd0e:

PR #732, opened 2023-09-19
Simplify Enc building scripts

johnd0e:
Unify inet/chm versions of Encyclopedia *.html.
(Continuation of #708)

As I found, for chm cp1251 is required for `pluginsr.hhc` and `pluginsr.hhk` only.
*.html can be in utf-8, same as for web version.
Thus we do not need to have 2 different sets of *.html

### DIFF
--- a/enc/tools/tool.make_chm.py
+++ b/enc/tools/tool.make_chm.py
@@ -83,11 +83,11 @@ def make_chm_lang(lang):
   for root, dirs, files in walk(chm_meta_dir):
     for f in files:
       infile  = open(join(root, f), encoding="utf-8-sig")
-      outfile = open(join(root.replace(chm_meta_dir, chm_html_dir), f), "w", encoding="windows-1251")
+      outfile = open(join(root.replace(chm_meta_dir, chm_html_dir), f), "w", encoding="utf-8-sig")
       header_replaced = False
       for line in infile:
         if not header_replaced and header_match in line:
-          line = line.replace(header_match, header_replace)
+          #line = line.replace(header_match, header_replace)
           header_replaced = True
 
         #while link_match.search(line):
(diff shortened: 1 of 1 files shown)


yegor-mialyk (approved):

alabuzhev (approved):

alabuzhev:
Looks good, rebase please.

alabuzhev:
Thanks.
I've applied the same approach (cp1251 for index files only) to Lua-specific CHMs as well.

johnd0e:
And macroapi_manual.ru.chm is broken: `charset=utf-8` is in header, but in fact all the text is in 1251.

johnd0e:
> I found one place where this doesn't work:

I have looked inside chm file and found `#STRINGS` file. The problem is that it contains both strings in `cp1251` and `utf-8`.
It seems that during compilation the titles of pages are picked "as is" (as byte stream).

I doubt that there is any way to change `hhc` behavior.
It should be possible to patch [#STRINGS](https://www.nongnu.org/chmspec/latest/Internal.html#STRINGS) directly, but not sure that it's worth it.

johnd0e:
Another possible (though ugly) workaround could be patching *.html files, adding copy of `<title>...</title>` using cp1251.

alabuzhev:
Reverted for now.

BestiaPL:
I looked through the sources of the encyclopedia - at the moment I see that it is generated in Russian, and lacking in English.

The scripts that generate the CHM file are set to generate one version of the encyclopedia, it would be useful to unify them - to generate multiple versions.
I also reviewed that the main code page is WIN-1251. In Poland, we use the code page WIN-1250. Ideally, all language versions should use UTF-8.

johnd0e:
@BestiaPL 
Actually the build of different languages is already unified, as pteviously we did support English version as well.

The reason why we do not build English now - it was never updated for Far 3.

alabuzhev:
> Ideally, all language versions should use UTF-8.

They are, in git and online.
Unfortunately, as discussed above, CHM format doesn't work with UTF-8 properly, so we have to convert everything to a compatible ansi codepage before compiling it.

PR #738, opened 2023-09-27
[+] фильтр исключений для arclite

qlost:


w17:
Why exclude masks only?
Masks syntax allow both include and exclude filters. 
And some patch description is definitely required.

PR #745, opened 2023-11-06
Changed quotation marks

BestiaPL:
Unified quotation marks - all in this same schema.

## Checklist


alabuzhev:
@BestiaPL thanks. However, are you sure that all these languages use the same quotation marks?

If you want to make it consistent, it probably should be consistent with other quotation marks in the same file, not consistent in one particular paragraph across 7 different files.

BestiaPL:
My translation is based on English (and the missing fragments are based on Russian).
I noticed that on line 241
https://github.com/FarGroup/FarManager/blob/8891d36fedbbbedb2ee15c82b75bec9943786d90/far/FarEng.hlf.m4#L241
there is a different quotation mark than on line 237:
https://github.com/FarGroup/FarManager/blob/8891d36fedbbbedb2ee15c82b75bec9943786d90/far/FarEng.hlf.m4#L237

It's the same with lines:
https://github.com/FarGroup/FarManager/blob/8891d36fedbbbedb2ee15c82b75bec9943786d90/far/FarEng.hlf.m4#L200
https://github.com/FarGroup/FarManager/blob/8891d36fedbbbedb2ee15c82b75bec9943786d90/far/FarEng.hlf.m4#L519-L520
https://github.com/FarGroup/FarManager/blob/8891d36fedbbbedb2ee15c82b75bec9943786d90/far/FarEng.hlf.m4#L2269
i.e.
English help file:
image

Russian help file - there is no-one single quotation mark.

Ukrainian:
image

Chech:
image

German:
image


You are right - quotation marks should be:
1) consistent with the grammar of a given country
2) uniform throughout the file.

I don't know English or Russian grammar very well and maybe we should use different citation marks depending on the country.
Short article from Wikipedia: https://en.wikipedia.org/wiki/Quotation_mark

alabuzhev:
> Russian help file - there is no-one single quotation mark.

Indeed, that one uses `‹›` since 99c501487a5f18f0c5bf130319287697688279c7.
Probably it would make sense to keep using `‹›` there (@HamRusTal?). 

Some numbers to consider:

|File\QM|'|"|‘|’|‚|‛|“|”|„|‟|‹|›|«|»|
|-|-|-|-|-|-|-|-|-|-|-|-|-|-|-|
|FarCze.hlf.m4|74|427|24|42|0|0|54|54|0|0|0|23|4|3|
|FarEng.hlf.m4|72|373|30|48|0|0|79|79|0|0|0|23|4|3|
|FarGer.hlf.m4|79|378|30|48|0|0|79|79|0|0|0|23|4|3|
|FarHun.hlf.m4|39|498|14|31|0|0|47|47|0|0|0|23|4|3|
|FarPol.hlf.m4|52|415|31|31|0|0|62|62|0|0|0|23|2|1|
|FarRus.hlf.m4|52|401|0|15|0|0|10|10|0|0|27|50|87|86|
|FarSky.hlf.m4|78|386|27|45|0|0|78|78|0|0|0|23|4|3|
|FarUkr.hlf.m4|285|393|25|52|0|0|69|69|0|0|0|23|4|3|

As a heatmap:

image

From a maintainer's perspective I'd be happy to drop all the fancy quotes in favor of `"`: it dominates everywhere already and help isn't a piece of literature anyway.

MKadaner:
According to my (limited) knowledge of Russian typographical conventions, `‹›` is not a thing at all, and `«»` («ёлочки») should be used throughout.

For English, my observations suggest that there are no hard-and-fast rules to choose between `‘’` and `“”`, but the latter («лапки») are prevalent at the top nesting level.

> From a maintainer's perspective I'd be happy to drop all the fancy quotes in favor of `"`...

Boy! Do I agree.

Px-x64:
Moving to one type of quotes seems good, but one thing to consider here is nested quotes (I didn't check if they are present in help). Text like "and then press "Enter"" is not really good, and using different quotes for "external" quotes adds readability: 'and then press "Enter"'. Thus needs to check if there are nested quotes, and select secondary quotes for such cases if they are present :)

MKadaner:
I think that we should utilize Far help syntax highlight (`#Enter#`). Then the question of nested quotes should be mostly moot.

HamRusTal:
> it would make sense to keep using `‹›` there (@HamRusTal?)

I have no strong feelings on this.

> help isn't a piece of literature anyway

It's still a text in a human language, subject to cultural differences. Quotation marks are part of punctuation to the same extent as e.g. prefix `¿` and `¡` in Spanish. Also "straight quotes" require more complex context to highlight the quoted text than “these” or «those» paired quotation marks.

PR #762, opened 2023-12-20
Use `std::variant` for deferred value

mishasychev:
## Summary
Use `std::variant` to reduce memory consumption for any `lazy` object.

## Checklist
- [x] I have discussed this with project maintainers:  <br/>

## Details
Let's not store the data and initializer separately, because we only need one of them at the moment of time. This is especially important for the _MSVC_ compiler, in which `std::function` has size of 64 bytes on x64 systems.


far/common/lazy.hpp
```diff
@@ -33,25 +33,27 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+		m_Data(std::move(Initialiser))
 	{
 	}
 
 	T& operator*()
 	{
-		return m_Data?
-			*m_Data :
-			*(m_Data = m_Initialiser());
+		if (initializer_type* compute = std::get_if<initializer_type>(&m_Data))
```
alabuzhev, line 53: Let's use auto, no point in specifying the same type twice on the same line.

far/common/lazy.hpp
```diff
@@ -69,22 +71,15 @@ class lazy
...
 	{
-		m_Data = FWD(rhs);
+		m_Data.template emplace<T>(FWD(rhs));
 		return *this;
 	}
 
 private:
-	std::function<T()> m_Initialiser;
-	std::optional<T> mutable m_Data;
+	std::variant<T, initializer_type> mutable m_Data;
```
alabuzhev, line 82: Neat. Indeed, we only need one.

alabuzhev (approved):

alabuzhev:
Squash please.

mishasychev:
> Squash please.

Done.

alabuzhev:
Thanks

PR #767, opened 2023-12-27
Refactored `VMenu::ShowMenu` and its vicinity.

MKadaner:
## Summary
Refactoring in preparation for #711.

## References
- #711

## Checklist

## Details
Sorry, it is large, but I believe overall the code is cleaner and easier to understand.

CodeFactor is complaining, and probably rightfully so. On the positive side, one issue is fixed.


MKadaner:
CodeFactor is complaining, and probably rightfully so. On the positive side, one issue is fixed.

far/vmenu.hpp
```diff
@@ -53,25 +53,25 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //----------------------------------------------------------------------------
 
 // Цветовые атрибуты - индексы в массиве цветов
-enum vmenu_colors
+enum class vmenu_colors
```
alabuzhev, line 56: Do we actually need this in .hpp?

far/vmenu.cpp
```diff
@@ -1957,32 +2054,34 @@ void VMenu::DisplayObject()
...
+		if (!CheckFlags(VMENU_DISABLEDRAWBACKGROUND) && !(BoxType==SHORT_DOUBLE_BOX || BoxType==SHORT_SINGLE_BOX))
 			SaveScr = std::make_unique<SaveScreen>(rectangle{ m_Where.left - 2, m_Where.top - 1, m_Where.right + 4, m_Where.bottom + 2 });
 		else
 			SaveScr = std::make_unique<SaveScreen>(rectangle{ m_Where.left, m_Where.top, m_Where.right + 2, m_Where.bottom + 1 });
 	}
 
 	if (!CheckFlags(VMENU_DISABLEDRAWBACKGROUND) && !CheckFlags(VMENU_LISTBOX))
 	{
-		// BUGBUG, dead code
+		// BUGBUG, dead code -- 2023-07-08 MZK: Is it though? I've got here when pressed hotkey of "Select search area" combobox on "find file" dialog.
```
alabuzhev, line 2069: TBH I don't remember the reason of the comment.

Maybe it applies only to the `!CheckFlags(VMENU_DISABLEDRAWBACKGROUND)` part of the condition, which is always true, since nothing sets the flag anywhere (perhaps we should remove it).
Or maybe I didn't notice the `!` and thought that the whole branch is unreachable.

far/vmenu.cpp
```diff
@@ -1834,15 +1877,69 @@ bool VMenu::ShiftAllItemsShowPos(int Shift)
...
+	const auto MaxLineWidth{ CalculateMaxLineWidth() };
+
 	for (const auto I: std::views::iota(size_t{}, Items.size()))
-		NeedRedraw |= ShiftItemShowPos(static_cast<int>(I), Shift);
+		NeedRedraw |= ShiftItemShowPos(static_cast<int>(I), Shift, MaxLineWidth);
 
 	return NeedRedraw;
 }
 
+struct item_layout
```
alabuzhev, line 1888: Wow, it's finally structured.
Thank you.

far/vmenu.cpp
```diff
@@ -2129,253 +2331,139 @@ void VMenu::ShowMenu(bool IsParent)
...
+	auto NamePos{ (separator.size() - NameWidth) / 2 };
 
-				// fit menu string into available space
-				bool ShowRightScroller = false;
-				if (MenuItemForDisplay.size() > MaxLineWidth)
-				{
-					MenuItemForDisplay.resize(MaxLineWidth);
-					ShowRightScroller = true;
-				}
+	separator.at(NamePos - 1) = L' ';
```
alabuzhev, line 2393: Why `at`, do we expect exceptions?

far/vmenu.cpp
```diff
@@ -2129,253 +2331,139 @@ void VMenu::ShowMenu(bool IsParent)
...
+	separator.replace(NamePos, NameWidth, fit_to_left(CurItem.Name, NameWidth));
+	separator.at(NamePos + NameWidth) = L' ';
+}
 
-				strMenuLine += MenuItemForDisplay;
+void VMenu::DrawRegularItem(const MenuItemEx& CurItem, const item_layout& Layout, const int Y, std::vector<int>& HighlightMarkup, const string& BlankLine) const
```
alabuzhev, line 2398: Any reason BlankLine can't be const string_view?

far/vmenu.cpp
```diff
@@ -2129,253 +2331,139 @@ void VMenu::ShowMenu(bool IsParent)
...
+		CurPos = SliceEnd;
+	}
 
-				if (Items[I].ShowPos)
-				{
-					GotoXY(m_Where.left + (m_BoxType == NO_BOX? 0 : 1) + 1, Y);
-					Text(L'«');
-				}
+	SetColor(ColorIndices.Normal);
+	Text(string_view{ BlankLine.begin(), BlankLine.begin() + (TextBegin + TextWidth - WhereX()) });
```
alabuzhev, line 2447: BlankLine.substr(...) maybe?
(which should be cheap if BlankLine is string_view)

far/vmenu.cpp
```diff
@@ -3048,4 +3129,72 @@ TEST_CASE("find.nearest.selectable.item")
...
+	for (const auto& TestDataPoint : TestDataPoints)
+	{
+		REQUIRE(TestDataPoint.Intersection == Intersect(TestDataPoint.A, TestDataPoint.B));
+		REQUIRE(TestDataPoint.Intersection == Intersect(TestDataPoint.B, TestDataPoint.A));
+	}
+}
+
+TEST_CASE("markup.slice.boundaries")
+{
+	static struct test_data
```
alabuzhev, line 3162: Can it be const?

far/vmenu.cpp
```diff
@@ -3048,4 +3129,72 @@ TEST_CASE("find.nearest.selectable.item")
...
+	{
+		std::pair<int, int> Segment;
+		std::vector<std::pair<int, int>> Slices;
```
alabuzhev, line 3165: Do we need vectors here?
The data is static, so spans, initializer_lists or arrays should probably do.

alabuzhev (commented):
It looks way too complex for a thorough review :)
I trust you spent quite some time debugging it and know what you're doing, so the comments will likely be only about technical details & style.

far/vmenu.hpp
MKadaner: Down below, we have

```C++
FarColor Colors[std::to_underlying(vmenu_colors::COUNT)]{};
```

Is there a way to get the number of Far colors (15) without exposing this definition?

far/vmenu.hpp
```diff
@@ -318,6 +331,8 @@ class VMenu final: public Modal
 	void UpdateSelectPos();
 	void EnableFilter(bool Enable);
 
+	[[nodiscard]] FarColor GetColor(vmenu_colors ColorIndex) const noexcept { return Colors[std::to_underlying(ColorIndex)]; }
```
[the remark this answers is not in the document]
MKadaner: Nope! Thanks for pointing out.

far/vmenu.cpp
```diff
@@ -172,6 +171,47 @@ int find_nearest(std::ranges::contiguous_range auto const& Range, const int Pos,
 			: FindPos(drop(Pos), take(Pos) | reverse));
 }
 
+template <std::integral T>
```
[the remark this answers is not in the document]
MKadaner: Well, we can specialize them for the `std::pair<int, int>` which is the only one required, and for `std::vector<int>& Markup`, though I am not sure it makes things really easier to understand. What we cannot do is to avoid generic `std::ranges::input_range auto const& Slices` because we actually call this function with different view types. I will apply the changes.

far/vmenu.cpp
```diff
@@ -172,6 +171,47 @@ int find_nearest(std::ranges::contiguous_range auto const& Range, const int Pos,
...
+std::pair<T, T> Intersect(std::pair<T, T> A, std::pair<T, T> B)
+{
+	assert(A.first < A.second);
+	assert(B.first < B.second);
+
+	if (B.first < A.first)
+		std::swap(A, B);
+
+	if (A.second <= B.first)
+		return { {}, {} };
```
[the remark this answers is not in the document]
MKadaner: Good point! Thanks.

far/vmenu.cpp
```diff
@@ -172,6 +171,47 @@ int find_nearest(std::ranges::contiguous_range auto const& Range, const int Pos,
...
+
+	return { B.first, std::min(A.second, B.second) };
+}
+
+template <std::integral T>
+void MarkupSliceBoundaries(std::pair<T, T> Segment, std::ranges::input_range auto const& Slices, std::weakly_incrementable auto Markup)
+{
+	assert(Segment.first < Segment.second);
+
+	for (const auto Slice : Slices)
```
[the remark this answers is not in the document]
MKadaner: `Slices` is a collection of `std::pair<int, int>`, so there is not much use in the extra indirection.

I tried to restrict the type of the input range element here, but all my attempts look quite ugly. They could be acceptable in a general-purpose library but not in this one-off context. If you can suggest something you would accept here, I'll be glad to use it and learn.

far/vmenu.cpp
MKadaner: Interesting. I adjusted the comment.

If this this flag is indeed effectively unused, can we repurpose it? I need a new flag for enabling vertical alignment of annotations feature. It should be enabled only for specific lists, not for regular vertical menus, for example. Unfortunately, `VMENU_FLAGS` utilize all 32 bits, so I was not sure how to fit another one.

far/vmenu.cpp
MKadaner: Preexisting code, but of course an exception is not expected. Fixed.

far/vmenu.cpp
MKadaner: No reasons. Fixed. Thank you. Omitted `const` because `string_view` is effectively const anyway. Let me know if you agree.

far/vmenu.cpp
MKadaner: Sure!

far/vmenu.cpp
MKadaner: Absolutely!

far/vmenu.cpp
MKadaner: Ah, `initializer_list` works with `const`, but not with `constexpr` here. I originally played with `constexpr`, could not make it work, and gave up. Went with `vector`s and lost `const`. :( Thank you for calling this out. fixed.

MKadaner:
> [T]he comments will likely be only about technical details & style.

I addressed all comments so far. Could you please check them and resolve whatever looks good?

> I trust you spent quite some time debugging it and know what you're doing...

Well, yes, but... Thorough testing is difficult and time consuming.

@rohitab, if by any chance you can spend some time testing this change, I would hugely appreciate it. The change affects all kinds of vertical lists: menus, histories, Find All, comboboxes... The bugs are expected with vertical layout (irregular movement of selection or items, unexpected empty lines, etc.) and with drawing list borders.

I will of course test more during the long weekend. Just now, I noticed one glitch. Find File -> Code Page Combobox may appear something like this (note the mangled left border; if there is no scrollbar, the right border should look similar):

```
Using code page:    
Automatic detection 
┌───────────────────
├  1201 │ UTF-16 (Bi
├───────┼───────────
├  37   │ IBM EBCDIC
├  500  │ IBM EBCDIC
├  708  │ Arabic - A
├  720  │ Arabic - T
├  737  │ OEM - Gree
│  775  │ OEM - Balt
└───────────────────
```

I guess I know how to fix this one (even two ways). Will do during the weekend as well.

far/vmenu.hpp
alabuzhev: vector maybe?

far/vmenu.cpp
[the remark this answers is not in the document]
alabuzhev: I mentioned it mostly because static analyzers don't like by-value-for-loops at all and strongly suggest it's an error, without paying much attention to the underlying type.
So to see fewer bogus warnings it might be better to go with the flow.

far/vmenu.cpp
alabuzhev: Can't see why not (with a new name obviously).

far/vmenu.cpp
alabuzhev: If you're about `VMenu::at` - that's different.
VMenu is usually accessed via a (smart) pointer and writing `menu->at(idx)` is way more natural than `(*menu)[idx]`.

far/vmenu.cpp
alabuzhev: const is meaningful even on string_view: it means that you can't (accidentally) change the view itself, e.g. by calling remove_suffix or reassigning it altogether.

far/vmenu.cpp
MKadaner: Ah, sorry, it was I who introduced it here. I actually merged my changed from a different branch where I could not be so sure that the `separator` string is of correct size. It should not happen in this context.

far/vmenu.cpp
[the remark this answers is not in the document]
MKadaner: OK. Hopefully, compiler will optimize it out anyway.

far/vmenu.cpp
MKadaner: OK, I should've known your tastes better. 🙂

far/vmenu.hpp
MKadaner: Did it another way. Please see if you like it.

far/vmenu.cpp
MKadaner: Thanks. I thought about plugin interface which might be affected. I am always uncertain about plugin API.

Leaving it for now, and will get back to it when implementing vertical alignment of annotations.

MKadaner:
Fixed everything, including the bug with vertical borders.

Not squashing yet to make verification of the comments easier. Also, I want to test a bit more.

Please let me know if everything is OK.

far/vmenu.cpp
```diff
@@ -172,6 +171,45 @@ int find_nearest(std::ranges::contiguous_range auto const& Range, const int Pos,
 			: FindPos(drop(Pos), take(Pos) | reverse));
 }
 
+std::pair<int, int> Intersect(std::pair<int, int> A, std::pair<int, int> B)
+{
+	assert(A.first < A.second);
+	assert(B.first < B.second);
+
+	if (B.first < A.first)
+		std::swap(A, B);
```
alabuzhev, line 180: Let's use ranges::swap for consistency.
Same below.

far/vmenu.cpp
```diff
@@ -1834,15 +1875,69 @@ bool VMenu::ShiftAllItemsShowPos(int Shift)
...
+		if (NeedScrollbar(Menu))     Scrollbar = Right;
+		if (RightBox || Scrollbar)   Right--;
+		if (NeedSubMenu(Menu))       SubMenu = Right--;
+		if (NeedRightHScroll())      RightHScroll = Right--;
+
+		if (Left <= Right)
+			Text = { Left, Right + 1 - Left };
+	}
+
+	[[nodiscard]] static bool NeedBox(int BoxType) noexcept { return BoxType != NO_BOX; }
```
alabuzhev, line 1915: Should these static methods be in the struct?
If yes (e.g. to and highlight that they're about layout) it's fine, but it would be better to keep the same case style: `item_layout::GetServiceAreaSize` looks a bit eclectic.

far/vmenu.cpp
```diff
@@ -2129,253 +2365,139 @@ void VMenu::ShowMenu(bool IsParent)
...
+	{
+		HighlightMarkup.emplace_back(EndPos);
+	}
 
-				SetColor(Colors[(Items[I].Flags&LIF_DISABLE)?VMenuColorArrowsDisabled:(Items[I].Flags&LIF_SELECTED?VMenuColorArrowsSelect:VMenuColorArrows)]);
+	GotoXY(TextBegin, Y);
+	for (const auto SliceEnd : HighlightMarkup)
+	{
+		SetColor(Colors, CurColorIndex);
+		Text(string_view{ ItemTextToDisplay.begin() + CurPos, ItemTextToDisplay.begin() + SliceEnd });
```
alabuzhev, line 2475: `string_view(ItemTextToDisplay).substr(CurPos, SliceEnd - CurPos)` maybe?

far/vmenu.hpp
alabuzhev: It's ok, as long as we assert the equality.

alabuzhev:
The change looks good, although I haven't tested it.
 
> Thorough testing is difficult and time consuming.

Indeed. We're not in a hurry, so if you'd like to have another look or play more with it, please take your time, we can get back to it next year :)

far/vmenu.hpp
MKadaner: We do. Resolving this one.

far/vmenu.cpp
MKadaner: OK. Sorry, below where?

alabuzhev: Yes

far/vmenu.cpp
MKadaner: Two of them, `NeedSubMenu` and `NeedScrollbar`, need access to `VMenu` privates, as well as `item_layout` constructor. For this purpose, `item_layout` is made a friend. I'll rename them to snake_case.

far/vmenu.cpp
MKadaner: I thought about it. No real preference. Let's do it your way.

alabuzhev (approved):

MKadaner:
Thank you for the approval. I have been playing with it today for an hour or so and could not find any new issues. Let's call it a day.

I rebased and squashed. Please merge.

alabuzhev:
Ok. Thank you.

rohitab:
@MKadaner sorry I wasn't able to respond sooner. I am going to upgrade to this version and will let you know if I come across any issues.

MKadaner:
@rohitab, no problem at all. Your feedback is welcome at any time. When you upgrade, please pick the latest version. I pushed one more refactoring in the same area, drawing of the vertical lists to be more specific. The behavior should not be affected.

Issue #786, opened 2024-01-30
Enable autocompletion in edit controls after initialization


rohitab:
### Description of the new feature or improvement

Autocompletion in dialog edit controls works for History, File system and Environment. However, this initialization must be done when the control is initially created. If the control has already been created, setting flags like `DIF_EDITPATH` and `DIF_HISTORY` using `DM_SETDLGITEM` message do not work.

Even though History can be enabled after initialization, using `DM_SETHISTORY` message, the Autocompletion feature for History does not work.

I believe this feature would be useful, since it would allow enabling (or disabling) of History and Autocompletion features for dialogs created by plugins or by Far itself.

### Proposed technical implementation details (optional)

Code to implement this feature is already present, however, it is currently called in the [DlgEdit::DlgEdit](https://github.com/FarGroup/FarManager/blob/bd2ec802fc91be8e3ce5c608dec5d9cde0ac262c/far/dlgedit.cpp#L56-L105) constructor. I made some modifications to the code as shown in the patch below. It's not the ideal patch, but with these changes, both History and Autocompletion can be enabled _after_ the dialog has been created. These features can also be enabled from Lua `DialogEvent` handlers in response to `DN_INITDIALOG` messages, or enabled at any time using `DM_SETDLGITEM` message. Both these features are demonstrated in the attached script.

### Patch
This patch has been tested with Far Manager, version `3.0.6262.0 x64`, which is the most recent version. The patch is also attached as [`Autocompletion.patch`](https://github.com/FarGroup/FarManager/files/14093275/Autocompletion.patch)

```diff
diff --git a/far/dialog.cpp b/far/dialog.cpp
index bc854e41f..f35d51da8 100644
--- a/far/dialog.cpp
+++ b/far/dialog.cpp
@@ -637,7 +637,7 @@ std::any* Dialog::GetListItemComplexUserData(size_t ListId, size_t ItemId) const
   TODO: Необходимо применить ProcessRadioButton для исправления
         кривых рук некоторых плагинописателей (а надо?)
 */
-void Dialog::InitDialogObjects(size_t ID)
+void Dialog::InitDialogObjects(size_t ID, bool ReInit)
 {
 	size_t InitItemCount;
 	bool AllElements = false;
@@ -802,6 +802,10 @@ void Dialog::InitDialogObjects(size_t ID)
 					Item.ListPtr->SetVDialogItemID(I);
 				}
 			}
+			else if (ReInit)
+			{
[...]

PR #787, opened 2024-01-31
Enable autocompletion in edit controls after initialization

rohitab:
## Summary
Allows you to enable autocompletion in edit controls after they have been initialized. This can be done by sending a `DM_SETDLGITEM` message with the appropriate flags like `DIF_EDITPATH`, `DIF_HISTORY` etc. In Lua, you can also use the  `far.SetDlgItem` function.

 
## References
https://github.com/FarGroup/FarManager/issues/786

## Checklist
- [x] I have discussed this with project maintainers: https://github.com/FarGroup/FarManager/issues/786 <br/>

## Details

- Moved the initialization of autocompletion from the [`DlgEdit::DlgEdit`](https://github.com/FarGroup/FarManager/blob/445298e261499fc5e014c2431cbe190ea83a900d/far/dlgedit.cpp#L56-L105) constructor, to a separate function so that it can be called in response to a `DM_SETDLGITEM` message.
- Note that even through you can use `DM_SETHISTORY` to enable History for edit controls, autocompletion will _not_ be enabled in response to this message. If you want autocompletion of History, you must use `DM_SETDLGITEM` instead, and set the `History` and `Flags` fields. It's possible to enable autocompletion here as well, by calling `Dialog::InitDialogObjects` in response to this message, but to keep changes to a minimum, I decided not to add it in.


alabuzhev:
Thank you

rohitab:
@alabuzhev There is a bug in the code. If you send a `DM_SETHISTORY` message after autocompletion has been enabled, Far will crash. See [Glitches and bugs that appeared in “nightly builds”](https://forum.farmanager.com/viewtopic.php?p=177131#p177131)

### Analysis

The issue appears to be due to the following code, where `iHistory` is changed.

https://github.com/FarGroup/FarManager/blob/00e0e1f800d46d6bd3b3ee0964a507c8673d4eb6/far/dlgedit.cpp#L120-L123

Note that `iHistory` is passed to `lineEdit` in `DlgEdit::Init()`, as seen here:

https://github.com/FarGroup/FarManager/blob/00e0e1f800d46d6bd3b3ee0964a507c8673d4eb6/far/dlgedit.cpp#L106

### Patch
The following patch fixes the issue, although I'm not sure if this is the right way to code it. Should there be `EditControl::SetHistory` method that updates the `pHistory` member, or is this OK?

```diff
diff --git a/far/dlgedit.cpp b/far/dlgedit.cpp
index bfe168a72..51afefc8f 100644
--- a/far/dlgedit.cpp
+++ b/far/dlgedit.cpp
@@ -120,6 +120,10 @@ DlgEdit::~DlgEdit()
 void DlgEdit::SetHistory(string_view const Name)
 {
 	iHistory = std::make_unique<History>(HISTORYTYPE_DIALOG, Name, Global->Opt->Dialogs.EditHistory, false);
+	if (lineEdit)
+	{
+		lineEdit.get()->pHistory = iHistory.get();
+	}
 }
 
 bool DlgEdit::ProcessKey(const Manager::Key& Key)
```

alabuzhev:
> The following patch fixes the issue, although I'm not sure if this is the right way to code it

Neither am I, but it's better than crashing anyway.

Thanks, 6266.

Issue #790, opened 2024-02-03
DI_EDIT changes text after DM_SETDLGITEMSHORT


nightroman:
### Far Manager version

3.0.6266

### OS version

10.0.22631.3007

### Other software

To reproduce, the plugin `FarNet` is needed and its module `FarNet.PowerShellFar`, the latest versions recommended,
see <https://github.com/nightroman/FarNet#readme> for how to install.

Perhaps, the steps are doable using Lua but I cannot provide such code, I do not really master Lua in Far.

### Steps to reproduce

(1) In a panel, navigate to any file, open the attributes dialog `CtrlA`.

(2) Click the edit box `Owner:`, delete all test (`BUILTIN\Administrators` in my case).

(3) Open the PowerShellFar command box: `F11` / `PowerShellFar` / `Invoke commands`

(4) Type/paste `$far.Dialog.Focused.IsPath = $true` and press `Enter`.

As a result, the original text `BUILTIN\Administrators` is unexpectedly restored.

### What plugin does

The plugin changes `DI_EDIT` flags like `DIF_EDITPATH` using `DM_GETDLGITEMSHORT` for getting original data and then `DM_SETDLGITEMSHORT` for setting changed. Per Far enc., `DM_SETDLGITEMSHORT` should not affect text.

Here is the used code:
https://github.com/nightroman/FarNet/blob/275581a8402e8ed43cda2963bf5e1cdd8242e952/FarNet/FarNetMan/DialogControls.cpp#L99-L114

```cpp
void FarControl::SetFlag(int flag, bool value)
{
    if (_dialog->_hDlg != INVALID_HANDLE_VALUE)
    {
        FarDialogItem di;
        Info.SendDlgMessage(_dialog->_hDlg, DM_GETDLGITEMSHORT, Id, &di);
        if (value == ((di.Flags & flag) != 0))
            return;
        SET_FLAG(di.Flags, flag, value);
        Info.SendDlgMessage(_dialog->_hDlg, DM_SETDLGITEMSHORT, Id, &di);
    }
    else
    {
        SET_FLAG(_flags, flag, value);
    }
}
```

### Expected behavior

`DM_SETDLGITEMSHORT` should not change text.

### Actual behavior

`DM_SETDLGITEMSHORT` changes text.

rohitab:
@nightroman Thank you for reporting the issue and providing the instructions to test.

[...]

PR #791, opened 2024-02-04
gh-790 and issues with DM_SETMAXTEXTLENGTH

rohitab:
## Summary
Fixes the following issues:

- If you send `DM_SETDLGITEMSHORT` or `DM_SETMAXTEXTLENGTH`, the text in the edit control is restored to the previous text.
- If you use `DM_SETMAXTEXTLENGTH` to set the maximum length, this value is not returned by `DM_GETDLGITEM` or `DM_GETDLGITEMSHORT`, or by `far.GetDlgItem` in Lua.

 
## References
#790

## Checklist

## Details

- `DM_SETDLGITEMSHORT` calls [`InitDialogObjects`](https://github.com/FarGroup/FarManager/blob/5b00dc1611117b4a6c9cb3def29a20db4aeae8f8/far/dialog.cpp#L5777) which calls [`DlgEdit::SetString`](https://github.com/FarGroup/FarManager/blob/5b00dc1611117b4a6c9cb3def29a20db4aeae8f8/far/dialog.cpp#L947). However, it uses the existing text `Item.strData` of the edit control. I modified the code so that this value is initialized with the current text, before the call to `InitDialogObjects`. This is only done for edit controls.
- `DM_SETMAXTEXTLENGTH` was calling [`InitDialogObjects`](https://github.com/FarGroup/FarManager/blob/5b00dc1611117b4a6c9cb3def29a20db4aeae8f8/far/dialog.cpp#L5734), which, just like the case above, was causing the text to be restored to the previous one. In this case, I removed the call to `InitDialogObjects`, since it doesn't appear to be required. I tested a number of different cases, with `DI_EDIT`, `DI_FIXEDIT`, `DI_PSWEDIT` and `DI_COMBOBOX`, and I couldn't find any issues. However, I'm not aware of why this code was put here originally. I looked at previous commits, but could not find a reason. Please double-check this.
- After `DM_SETMAXTEXTLENGTH` was used, a call to `DM_GETDLGITEM` or `DM_GETDLGITEMSHORT` was not returning the value that was set. To fix this, I updated the value of `CurItem.MaxLength`. Please note that I did not change any of the original casts.


nightroman:
Note to maintainers, I have tried the fix. It fixes known issues in FarNet, all tests pass, unwanted effects are not found yet.

alabuzhev:
Thank you

PR #806, opened 2024-03-02
Update build

bitraid:
## Summary
Some improvements to the build system and github workflow.

Added all the [environments of msys2](https://www.msys2.org/docs/environments/), except MSYS (doesn't target native Windows programs) and CLANGARM64 (cross compilation from x86 is not supported yet).

Also added cross compile support from linux using llvm-mingw, for all supported targets, including arm32/64 (tested in qemu).

Some other small improvements are included, each in its own commit for easier review.

I didn't increase the Far version, since the changes don't affect the actual program.
 
## References
https://github.com/FarGroup/FarManager/discussions/749#discussioncomment-8577527

## Checklist
- [x] I have discussed this with project maintainers:  <br/>


far/main.cpp
```diff
@@ -476,7 +476,7 @@ static void handle_exception(function_ref<bool()> const Handler)
 	throw;
 }
 
-#ifndef _WIN64
+#if !defined(_WIN64) && !defined(_ARM_)
```
alabuzhev, line 479: Might be simpler to `#ifdef _M_IX86` (same in other similar places).

plugins/luamacro/luasdk/ARM32/lua51.dll
alabuzhev, line 1: I see that lua5.1.dll and lpeg.dll depend on all those api-ms-win-crt-whatever, while the ARM64 dlls we already have in the repo don't.
Perhaps it's better to link these statically too for consistency.

plugins/newarc.ex/Framework/Src/farlng2.cpp
alabuzhev, line 1: This plugin is a deep legacy and is not part of the build, it's ok to ignore it.

plugins/proclist/Pclass.cpp
```diff
@@ -30,7 +30,11 @@ class StrTok
 		tok(token)
 	{
 		buf.reset(wcsdup(str));
+#if defined(_UCRT)
```
alabuzhev, line 33: Is there any reason to not use `std::wcstok(buf.get(), token, &pt);` unconditionally?

.github/workflows/build.yml
```diff
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ master ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
alabuzhev, line 11: Nice.

bitraid:
Did I somehow break appveyor?

alabuzhev (commented):
Nice, thank you.

I just hope that will all these new build configurations we are still within github limits, whatever they are.

alabuzhev:
> Did I somehow break appveyor?

Looking at the generated artefacts, all the hlf files are damaged.
Likely related to "Update gawk, m4 binaries" or "mkhlf.awk to utf-8".

bitraid:
> I just hope that will all these new build configurations we are still within github limits, whatever they are.

Nah, we 're good. There are no limits for public projects, only for private ones.

bitraid:
> > Did I somehow break appveyor?
> 
> Looking at the generated artefacts, all the hlf files are damaged. Likely related to "Update gawk, m4 binaries" or "mkhlf.awk to utf-8".

Ok, I'll revert the binaries first to see if it gets fixed.

far/exception_handler_test.cpp
```diff
@@ -502,7 +502,11 @@ WARNING_POP()
 	static void seh_assertion_failure()
 	{
 		if ([[maybe_unused]] volatile const auto Condition = true)
+#if !defined(_MSC_VER) && (defined(_ARM_) || defined(_ARM64_))
+			DebugBreak();
```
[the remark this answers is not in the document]
bitraid: Yes, it's not defined. I'll remote it then.

far/farversion.m4
```diff
@@ -32,6 +32,8 @@ M4_MACRO_DEFINE(BUILD_PRERELEASE, m4_ifelse(
 M4_MACRO_DEFINE(BUILD_PLATFORM, m4_ifelse(
 	FARBIT, 64, x64,
 	FARBIT, 32, x86,
+	FARBIT, ARM64, arm64,
```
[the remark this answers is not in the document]
bitraid: I was going to have it capitalized first, but then I saw some programs had it lower case. I'll change it.

plugins/proclist/Pclass.cpp
bitraid: I won't work for mingw-msvcrt:
https://github.com/mingw-w64/mingw-w64/blob/e0d90587c9825614fe59eda9066ad12fe72ff9bf/mingw-w64-headers/crt/string.h#L147-L151

plugins/luamacro/luasdk/ARM32/lua51.dll
bitraid: It's linked against ucrt, which is part of Windows 10 and later (where arm is supported). This happens for all other binaries of both arm32/arm64 from llvm-mingw (in contrast win vc). Are you sure?

alabuzhev: Not absolutely sure - I've seen quite a few issues with ucrt on older OSes, but not on 10, so I'm not saying it won't work.
My only argument is consistency: our prod binaries are vc, what happens with llvm-mingw is not that important.

---
By the way, do we need to "officially" publish ARM32 builds?
I think everything should work already, but no one ever asked.

bitraid: That makes sense. Even more so if you're thinking of providing an arm32 release.

alabuzhev:
Looks like all is good now.

Thanks, merged.

bitraid:
Awesome. Thanks.

PR #807, opened 2024-03-06
Fix lto build for clang

bitraid:
## Summary
I forgot to add this check (clang has no `lto-odr-type-merging`).
 
## Checklist

### DIFF
--- a/far/makefile_gcc_common
+++ b/far/makefile_gcc_common
@@ -223,9 +223,11 @@ CFLAGS += \
 	-O3 \
 
 ifeq ($(USE_LTO),1)
-CFLAGS += \
-	-flto \
-	-flto-odr-type-merging \
+CFLAGS += -flto
+
+ifndef CLANG
+CFLAGS += -flto-odr-type-merging
+endif
 
 endif # USE_LTO
 


bitraid:
Should we add an lto build to the workflow?

alabuzhev:
Thanks.

> Should we add an lto build to the workflow?

I'd rather just turn it on for release builds.
It's not enabled currently only because it was [broken](https://sourceforge.net/p/mingw-w64/bugs/557/) years ago.
Maybe it's fixed already, I haven't checked.

bitraid:
It is fixed now. I can make the change if you want. Also, I could tweak the msvc build, so it doesn't require two separate jobs, and avoid code duplication.

alabuzhev:
That would be great, thank you.
What do you mean by two separate jobs?

bitraid:
I meant to say two separate build definitions.

bitraid:
You know, one for msbuild and another for nmake. Could be added on one matrix.

alabuzhev:
Ah. Yes, can't see why not.

bitraid:
OK, great.

PR #825, opened 2024-04-09
Fixed Hex mode width calculations in Viewer.

MKadaner:
## Summary
Fixed a few places where the line width in the Viewer Hex mode was assumed to be 80 even though the number of bytes per line can now be changed by the user.

Also, some renaming and refactoring.

## References
0ecc671

## Checklist


far/viewer.cpp
```diff
@@ -4089,3 +4159,80 @@ void Viewer::vgetc_cache::compact()
...
+		{
+			for (const auto& LineWidth : std::views::iota(hex_mode_line_width_from_bytes(Bytes, CharSize), hex_mode_line_width_from_bytes(Bytes + 1, CharSize)))
+			{
+				REQUIRE(Bytes == hex_mode_bytes_from_line_width(LineWidth, CharSize));
+			}
+		}
+	}
+}
+
+int HexModeByteFromColumn16BytesFixed(const int Column, const int CharSize)
```
alabuzhev, line 4212: Please mark as static to avoid "[missing declaration](https://github.com/FarGroup/FarManager/actions/runs/8609968009/job/23594740225?pr=825#step:7:249)" warnings

far/viewer.cpp
```diff
@@ -4089,3 +4159,80 @@ void Viewer::vgetc_cache::compact()
...
+		{ 33, 107 }, // > 2D 2D 2D 2D 2D 2D 2D 2D │ 2D 2D 2D 2D 2D 2D 2D 2D │ 2D 2D 2D 2D 2D 2D 2D 2D │ 2D 2D 2D 2D 2D 2D 2D 2D │ 2D<
+	};
+
+	for (const auto& TestDataPoint : TestDataPoints)
+	{
+		REQUIRE(TestDataPoint.Width == hex_mode_bytes_area_wdth(TestDataPoint.Bytes));
+	}
+}
+
+TEST_CASE("hex.mode.bytes.from.line.width")
```
alabuzhev, line 4198: Please don't use `.` as a word separator: UI tools often group tests by these tokens, supposedly assuming something like a namespace.class.method structure, which doesn't make any sense in this case:

image

Better something like "area.function_being_tested" e.g. "viewer.hex_mode_bytes_from_line_width"

MKadaner:
> Could you please give an example of an issue fixed by this change?

Here is a problematic scenario:
- In the Viewer, open a text file containing a relatively long line, say 350 characters.
- Press `F4` to switch to Hex mode.
- Press `Ctrl+Alt+Left` to reduce the width of Hex view to 8 bytes per line.
- Press `F4` again to switch to Text mode and `F2` to disable line wrap.
- Resize Far window width to 30 columns.
- Press `Ctrl+Shift+Right` to position the screen to the end of the longest line.
- Press `F4` to switch to Hex mode.

Observe that the Far window is empty, and the Hex dump completely disappears. In fact, the dump now is far beyond the left edge of the window.

One can argue that this issue is "by design," because the Viewer tries to preserve the horizontal position of the content while switching between view modes. However, the ostentatious intent of the existing code is to make sure the Hex dump is always visible. I think that the latter behavior is more important and should override "preserving horizontal position."

Another concern is `Viewer::XYfilepos`, `case VMT_HEX`. The code here translates the X mouse click coordinate to the byte offset. The code obviously assumes that Hex dump is always 16 bytes per line. With user-customizable number of bytes per line in Hex dump, it is simply not true. Unfortunately, I cannot describe the scenario when this issue becomes visible.

More importantly, I think that any particular visualization issue notwithstanding, it is plain wrong to have the width of Hex view hard coded to 80 columns in several places while the actual view width is variable.

Last but not the least. I stumbled upon this issue while I was about to introduce a `far:config` to reverse the horizontal scrolling direction. I absolutely could not understand what was going on in two or three places in Viewer implementation, so I could not easily generalize the code to handle the swap direction option. I spent substantial time to reverse engineer the code. This PR is the outcome of this effort.

far/viewer.cpp
MKadaner: Of course! Fixed.

far/viewer.cpp
MKadaner: LOL. I've never given a thought to this naming. Makes sense, of course. Fixed.

far/viewer.cpp
```diff
@@ -1291,28 +1362,23 @@
...
 	long long pos = -1;
 
-	const auto csz = getChSize(m_Codepage);
+	const auto csz = get_char_size(m_Codepage);
 	switch (m_DisplayMode)
 	{
 	case VMT_DUMP:
-		pos = FilePos + csz*(Width*row + col);
+	{
+		pos = FilePos + csz * (ScrollbarAdjustedWidth * row + col);
```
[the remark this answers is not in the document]
MKadaner: Fixed.

far/viewer.cpp
```diff
@@ -1291,28 +1362,23 @@
...
-		else if (col < 35) col = (col-11)/3;
-		else if (col < 37) col = s_BytesPerStripe;
-		else if (col < 61) col = s_BytesPerStripe + (col-37)/3;
-		else if (col < 63) col = 0;
-		else if (col < 63 + 16 / csz) col = (col-63) * csz;
-		else               col = 16;
-		pos = FilePos + 16*row + col / csz * csz;
+	{
+		const auto byte{ hex_mode_byte_from_column(col, m_HexModeBytesPerLine, csz) };
+		pos = FilePos + 16 * row + byte / csz * csz;
```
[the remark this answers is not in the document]
MKadaner: Fixed.

MKadaner:
I looked through sonarcloud's warnings. All of them are preexisting issues except for:
- [Move this #include directive to the top of the file.](https://sonarcloud.io/project/issues?resolved=false&pullRequest=825&id=FarGroup_FarManager&open=AY7A9WC8Ez_cHOjpzsfU) - Expected?
- [Use "std::array" or "std::vector" instead of a C-style array.](https://sonarcloud.io/project/issues?resolved=false&pullRequest=825&id=FarGroup_FarManager&open=AY7A9WC8Ez_cHOjpzsfe) - It looks like `std::array` here requires either specifying the array size explicitly or putting `test_data` on each line. Neither seems to be a better alternative to C-style array.

Let me know, if you want me to fix any of these issues, even preexisting. I'll do your bid.

alabuzhev:
Please ignore those warnings.

alabuzhev:
> I absolutely could not understand what was going on in two or three places in Viewer implementation

Ah yes, it's... tangled, to put it mildly.
Thanks.

MKadaner:
Thank you.

rohitab:
> Another concern is `Viewer::XYfilepos`, `case VMT_HEX`. The code here translates the X mouse click coordinate to the byte offset. The code obviously assumes that Hex dump is always 16 bytes per line. With user-customizable number of bytes per line in Hex dump, it is simply not true. Unfortunately, I cannot describe the scenario when this issue becomes visible.

The issue becomes visible when you use the mouse to select bytes in the Hex dump view. If the bytes-per-line are anything other than `16`, the selection starts and ends at the wrong position. For example:

1. Open a file in Hex view
2. Press `Alt+Left` to reduce the bytes-per-line to `7`
3. Press `Shift+Left_Click` on the byte `1` in the second line. Note that the byte `2` gets selected instead.
4. Press `Shift+Left_Click` on the byte `8` in the second line.

You'll see that bytes `2-9` get selected, instead of bytes `1-8`.

Unfortunately, this issue is still present in version `3.0.6304.0 x64`

MKadaner:
Ouch! Silly me. Fixed (hopefully) in #826.

Thank you for reporting, @rohitab.

PR #835, opened 2024-04-30
Enhance the function mf.flock to enable switching the Lock-state even when a key is held

johnd0e:
When a key is held we just need the opposite order of key presses (release, then press).

Sample macro to test the feature:
```lua
Macro {
  description="flock test";
  area="Editor"; key="CtrlCapsLock";
  priority=100;
  action=function()
    mf.flock(1,2)
  end;
}
```

Fix Mantis #3557


far/keyboard.cpp
```diff
@@ -382,8 +382,16 @@ int SetFLockState(unsigned const vkKey, int const State)
 	{
 		if (State == 2 || (State==1 && !oldState) || (!State && oldState))
 		{
-			keybd_event(vkKey, 0, ExKey, 0);
-			keybd_event(vkKey, 0, ExKey | KEYEVENTF_KEYUP, 0);
+			if (oldState&0x8000 == 0x8000) // key is down
```
alabuzhev, line 385: Just `if (oldState & 0x8000)`  will do.

johnd0e: Done

alabuzhev (approved):
Please update changelog & build and squash commits.

johnd0e:
Done

PR #836, opened 2024-05-03
Apply ENHANCED_KEY for more relevant keys in TranslateKeyToVK

johnd0e:
Fix issue with `FarNameToInputRecord` reported here: https://forum.farmanager.com/viewtopic.php?p=177899#p177899

todo: https://bugs.farmanager.com/view.php?id=4044


far/keyboard.cpp
```diff
@@ -1730,10 +1730,18 @@ int TranslateKeyToVK(int Key, INPUT_RECORD* Rec)
...
+						KEY_RWIN,
+						KEY_APPS,
+						KEY_PRNTSCRN,
+						KEY_BREAK,
+						KEY_DIVIDE,
+						KEY_NUMLOCK
 					};
 
-					if (contains(ExtKey, FKey))
+					if (contains(ExtKey, FKey) || VirtKey==VK_RCONTROL || VirtKey==VK_RMENU)
```
alabuzhev, line 1744: VirtKey deduction can be ambiguous:
https://github.com/FarGroup/FarManager/blob/13b99b1d19a279de6bce350e0c8d945677f5323c/far/keyboard.cpp#L1605-L1618

Maybe check for `Key == KEY_RCTRL || Key == KEY_RALT`?

johnd0e: > Maybe check for Key == KEY_RCTRL || Key == KEY_RALT?

Hm, that actually was meant to also work with combinations like <kbd>Ctrl</kbd>+<kbd>RAlt</kbd>
But I see now that using VirtKey does not help as well...

alabuzhev: If you want to support CtrlAlt etc. the last one should be treated as a key (`FKey` above) and the rest as modifiers (`FShift` above), which is currently not the case.

johnd0e: Fixed for now.

PR #853, opened 2024-07-08
Set of macros to swap horizontal scroll direction in Viewer, Menus, and Shell.

MKadaner:
When enabled, these macros change horizontal scrolling direction in the Internal Viewer, VMenu / Lists, and File Panels.

Left is Right and Right is Left.

## Summary
For a long time, I was considering the behavior of the Left and Right Arrow keys in the Viewer, `Find All` menu, and Panels literally counterintuitive. The keys move viewport, while I normally want to move the items. Even after many years, I still almost always start scrolling with the wrong key.

Let's have these macros available out of the box.

## References
The issue was tangentially discussed [here](https://github.com/FarGroup/FarManager/pull/789#issuecomment-1925385503).

## Checklist

### DIFF
--- a/extra/Addons/Macros/Descript.ion
+++ b/extra/Addons/Macros/Descript.ion
@@ -37,6 +37,7 @@ Panel.SpaceToSelect.lua Panel: Use Space to select files
 Search.CtrlIns.lua Close Fast find and send key to panel
 ShiftBS.lua Converts the word before cursor using XLat function.
 ShiftTab.lua Shift+Tab = Ctrl+Shift+Tab
+SwapHorizontalScrollDirection.lua Panel, Viewer and Menu: Swap horizontal scroll direction
 Symlink.lua Alternative way to create symbolic links
 UppercaseFolders.lua Restore the option "Create folders in uppercase" that was removed in Far 3.0.4184
 View.CtrlUpDown.lua Viewer: Scroll screen (using keys like in editor)
--- a/extra/Addons/Macros/SwapHorizontalScrollDirection.lua
+++ b/extra/Addons/Macros/SwapHorizontalScrollDirection.lua
@@ -1,0 +1,65 @@
+local function LeftToRight()
+    Keys(mf.akey(1,1):gsub("Left", "Right"):gsub("Num4", "Num6"))
+end
+
+local function RightToLeft()
+    Keys(mf.akey(1,1):gsub("Right", "Left"):gsub("Num6", "Num4"))
+end
+
+Macro {
+  area="Viewer";
+  description="Swap horizontal scroll direction";
+  key=
+    "Left Num4 ShiftNum4"               -- Horizontal scroll in Text and Hex modes
+    .. " CtrlLeft CtrlNum4"             -- Horizontal scroll by 20 in Text mode; roll the contents in Dump and Hex modes
+    .. " MsWheelLeft AltMsWheelLeft";   -- Translates to "Left" repeated as necessary
+  action = LeftToRight;
+}
+
+Macro {
+  area="Viewer";
+  description="Swap horizontal scroll direction";
+  key=
+    "Right Num6 ShiftNum6"              -- Horizontal scroll in Text and Hex modes
+    .. " CtrlRight CtrlNum6"            -- Horizontal scroll by 20 in Text mode; roll the contents in Dump and Hex modes
+    .. " MsWheelRight AltMsWheelRight"; -- Translates to "Right" repeated as necessary
+  action = RightToLeft;
+}
+
+Macro {
+  area="Menu";
+  description="Swap horizontal scroll direction";
+  key=
+    "AltLeft AltNum4 MsWheelLeft"                   -- Horizontal scroll all items by 1 position
+    .. " CtrlAltLeft CtrlAltNum4 CtrlMsWheelLeft"   -- Horizontal scroll all items by 20 positions
+    .. " AltShiftLeft AltShiftNum4"                 -- Horizontal scroll the current item by 1 position
+    .. " CtrlShiftLeft CtrlShiftNum4";              -- Horizontal scroll the current item by 20 positions
+  action = LeftToRight;
+}
+
+Macro {
+  area="Menu";
+  description="Swap horizontal scroll direction";
+  key=
+    "AltRight AltNum6 MsWheelRight"                 -- Horizontal scroll all items by 1 position
+    .. " CtrlAltRight CtrlAltNum6 CtrlMsWheelRight" -- Horizontal scroll all items by 20 positions
+    .. " AltShiftRight AltShiftNum6"                -- Horizontal scroll the current item by 1 position
+    .. " CtrlShiftRight CtrlShiftNum6";             -- Horizontal scroll the current item by 20 positions
+  action = RightToLeft;
+}
+
+Macro {
+  area="Shell";
+  description="Swap horizontal scroll direction";
+  flags="EmptyCommandLine";
+  key= "AltLeft";   -- Horizontal scroll of file names and descriptions
+  action = LeftToRight;
+}
+
+Macro {
+  area="Shell";
+  description="Swap horizontal scroll direction";
+  flags="EmptyCommandLine";
+  key= "AltRight";  -- Horizontal scroll of file names and descriptions
+  action = RightToLeft;
+}
--- a/misc/msi-installer/features.wxs
+++ b/misc/msi-installer/features.wxs
@@ -103,6 +103,7 @@
           <ComponentRef Id="Search.CtrlIns.lua" />
           <ComponentRef Id="ShiftBS.lua" />
           <ComponentRef Id="ShiftTab.lua" />
+          <ComponentRef Id="SwapHorizontalScrollDirection.lua" />
           <ComponentRef Id="Symlink.lua" />
           <ComponentRef Id="UppercaseFolders.lua" />
           <ComponentRef Id="View.CtrlUpDown.lua" />
--- a/misc/msi-installer/files.wxs
+++ b/misc/msi-installer/files.wxs
@@ -376,6 +376,9 @@
           <Component Id="ShiftTab.lua" Guid="$(var.Guid.ShiftTab.lua)" Win64="$(var.Win64)">
             <File Id="ShiftTab.lua" KeyPath="yes" Source="$(var.SourceDir)\Addons\Macros\ShiftTab.lua" />
           </Component>
+          <Component Id="SwapHorizontalScrollDirection.lua" Guid="$(var.Guid.SwapHorizontalScrollDirection.lua)" Win64="$(var.Win64)">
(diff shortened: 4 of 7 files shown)


HamRusTal:
Does the change affect only keyboard arrows specifically? What about horizontal mouse wheel (I know there are mice/controllers with both vertical and horizontal wheels but I don't have any unfortunately to check myself.)
I wonder if the direction of horizontal wheel scrolling should be bound to the direction of horizontal arrows scrolling…

johnd0e:
> I believe the issue was discussed somewhere, but I could not find the original topic.

I bet it was not discuscussed previously.
Looks noone else has similar Problem ;-)

Btw, why not macros?

alabuzhev (commented):
Please set the commit author properly, I have nothing to do with it.

HamRusTal:
> I bet it was not discuscussed previously.

[It was](https://github.com/FarGroup/FarManager/pull/789#issuecomment-1925385503), yet in a slightly smaller scope.

far/filelist.cpp
```diff
@@ -634,6 +634,20 @@ void FileList::CorrectPosition()
 		m_CurTopFile=m_CurFile-m_Stripes*m_Height+1;
 }
 
+void FileList::HorizontalScroll(const bool LeftKeyPressed)
+{
+	if (LeftKeyPressed == Global->Opt->VMenu.HorizontalArrowsMoveContents)
```
alabuzhev, line 639: Should FileList behaviour depend on menu settings?

MKadaner:
> ... the direction of horizontal wheel scrolling should be bound to the direction of horizontal arrows scrolling

Yes, it should. My bad. Will fix it. I keep forgetting about mouse.

MKadaner:
> why not macros?

The short and honest answer is because it's much easier to me to change C++ code than to write a macro. That said, what advantage the macro would have over `far:config` options?

MKadaner:
> Please set the commit author properly,

I apologize. Do not know what happened. Will fix it of course.

If somebody can tell me what I did wrong, I will appreciate.

MKadaner:
> [It was](https://github.com/FarGroup/FarManager/pull/789#issuecomment-1925385503), yet in a slightly smaller scope.

Thank you for finding it, @HamRusTal. I will update the description of this PR accordingly.

far/filelist.cpp
MKadaner: Of course, it should not. I will add another `far:config` for FileList. Unfortunately, I could not get back to this task for so long that I forgot what I did at the beginning. 🙁

MKadaner:
> other lefty/righty navigational keys like Home & End

I do not think the behavior of `Home` and `End` keys should change. There is a fundamental difference between arrow keys and `Home` / `End` keys. Arrows move either the viewport or the contents. Depending on the user's perception of what is being moved, the behavior should be different.

Now, the "Home" key is not about moving anything left, or right, not Turnwise, nor even Widdershins. It means "bring me home" or "set everything in the initial position." The initial position is when the beginnings of the lines are visible in the viewport, it's lines' _beginnings,_ not lines' _ends._

Thus, the behavior of the "Home" key should not be affected by the setting controlling what is moving, viewport or contents. Nor should it be affected by any other setting. It is a little more difficult to apply similar line of arguments to the "End" key, but since it is traditionally just the opposite of the "Home," we should conclude that the "End" should not be affected either.

HamRusTal:
> > ... the direction of horizontal wheel scrolling should be bound to the direction of horizontal arrows scrolling
> 
> Yes, it should. My bad. Will fix it. I keep forgetting about mouse.

Then ideally the setting names should not mention arrows specifically. I suggest that they mention “horizontal navigation” or the like.

HamRusTal:
To be honest, I'm with others who consider that keyboard macros are the right mechanism to implement features like this. Swapping left/right in certain contexts, even with regard to keyboard modifiers, can be accomplished with simple macros, much simpler IMHO than integrate and maintain indefinitely a new bunch of settings (even as simple as these).

alabuzhev:
> If somebody can tell me what I did wrong, I will appreciate.

Looking at the timestamp, you probably branched from 5246488f7e69012a5f747be8d60d5d5f6c3dbb8c, but amended it at some point (`git commit --amend`) instead of making a new commit.

> what advantage the macro would have over `far:config` options?

- The user can write the macro themselves and the project maintainers won't have to maintain the code indefinitely.
- The user can do it _exactly_ how they like it, even if it's against the overall design, inconsistent or half-baked in the eye of the maintainer.
- The user doesn't have to persuade the maintainers that the change is reasonable, justified and worthy to support indefinitely.
- The maintainers won't break it accidentally.
- It is consistent with the project's philosophy:
  There are quite a few other macros that could've been (or were) in the code. Canonical examples are Esc and Del in panels. If we say "use a macro if you want to do something as obvious as delete files with Del", how can we justify baking something way, way more niche into the exe?

In general, I agree with other commenters here - it should be a macro. The arguments are above.
Happy to review your change, but I probably won't merge it, unless you manage to persuade others.

Having said that, if you decide to implement it as a macro, I don't see why we can't add it to the distribution, and installing something from Addons\Macros is not much harder than changing a config option.

> it's much easier to me to change C++ code than to write a macro

I can relate, but it's not as hard as it sounds. You can start with recording a simple one with `Ctrl.` and generalise and improve it afterwards. If you stuck or run out of ideas, I'm sure @johnd0e or @shmuz won't mind suggesting what to do next.

MKadaner:
> Looking at the timestamp, you probably branched from 5246488f,..

Thanks. Understand, but that would be extremely unlikely. I thought about the whole thing a bit more. This issue may have something to do with my local repo having `remote "alabuzhev"` config from the time when we discussed splitting `farlang.templ.m4`. It is plausible that at certain point I sprouted my branch from one in your repo. Anyway, I fixed it locally and will push if we agree on the below.

> I agree with other commenters here - it should be a macro.

Thank you for your patient explanations. I'll go with a set of macros, and let's add it to the distribution. Next PR.

For this PR, I am going to repurpose it to fix horizontal scrolling Viewer. Besides making the code a little cleaner, it will fix horizontal scrolling in Hex mode. Currently, it is possible, with switching modes and window resizing, to get Hex view shifted beyond the left edge of the window, and without a good way of scrolling it horizontally. New commit follows.

HamRusTal:
> going to repurpose it to fix horizontal scrolling Viewer

Then ideally the branch has to be renamed too, because the GitHub history records the name of the source branch, and (near) future history readers would be misguided by the current name.

alabuzhev:
> that would be extremely unlikely

Another way to achieve the same result is rebasing with squashes/fixups.

MKadaner:
> Then ideally the branch has to be renamed too...

Yeah! In fact, repurposing PR is not a good idea. It looks like GH does not allow to change source branch in PR. More importantly, the discussion in the PR will become irrelevant and confusing. I think I'll implement macros here and Hex Viewer improvements in a new one.

MKadaner:
> Another way to achieve the same result is rebasing with squashes/fixups.

This is one of my favorite operations. 😀 I'll be more careful in the future.

MKadaner:
@johnd0e, @shmuz, could you please have a look? Do these macros make sense to you? Would you suggest any improvements? I am aware of the use of regexes for matching keys, but I think that in this case explicit list is more appropriate.

```Lua
Macro {
  area="Viewer"; key="Left CtrlLeft RCtrlLeft Num4 ShiftNum4 CtrlNum4 RCtrlNum4 MsWheelLeft AltMsWheelLeft RAltMsWheelLeft";
  action = function()
    Keys(mf.akey(1,1):gsub("Left", "Right"):gsub("Num4", "Num6"))
  end;
}

Macro {
  area="Viewer"; key="Right CtrlRight RCtrlRight Num6 ShiftNum6 CtrlNum6 RCtrlNum6 MsWheelRight AltMsWheelRight RAltMsWheelRight";
  action = function()
    Keys(mf.akey(1,1):gsub("Right", "Left"):gsub("Num6", "Num4"))
  end;
}
```

shmuz:
@MKadaner 
These macros look good for me.

johnd0e:
> I am aware of the use of regexes for matching keys, but I think that in this case explicit list is more appropriate.

Too explicit is also not good, as for me.
I would use at least `R?` shortcut:
```lua
Macro {
  area="Viewer";
  key="Left R?CtrlLeft Num4 ShiftNum4 R?CtrlNum4 MsWheelLeft R?AltMsWheelLeft";
  action=function()
    Keys(mf.akey(1,1):gsub("Left", "Right"):gsub("Num4", "Num6"))
  end;
}

Macro {
  area="Viewer";
  key="Right R?CtrlRight Num6 ShiftNum6 R?CtrlNum6 MsWheelRight R?AltMsWheelRight";
  action=function()
    Keys(mf.akey(1,1):gsub("Right", "Left"):gsub("Num6", "Num4"))
  end;
}
```

shmuz:
> I would use at least `R?` shortcut:

It won't work with standard Far distribution.

johnd0e:
Indeed.
Anyway we do not need explicit "R" here. 
```lua
Macro {
  area="Viewer";
  key="Left CtrlLeft Num4 ShiftNum4 CtrlNum4 MsWheelLeft AltMsWheelLeft";
  action=function()
    Keys(mf.akey(1,1):gsub("Left", "Right"):gsub("Num4", "Num6"))
  end;
}

Macro {
  area="Viewer";
  key="Right CtrlRight Num6 ShiftNum6 CtrlNum6 MsWheelRight AltMsWheelRight";
  action=function()
    Keys(mf.akey(1,1):gsub("Right", "Left"):gsub("Num6", "Num4"))
  end;
}
```

MKadaner:
> Anyway we do not need explicit "R" here.

Ah, I noticed that it worked without `R-` variants but was not sure whether it's accidental or fully supported. Thank you for the confirmation.

BTW, what is the recommended formatting if I want to attach comments to individual keys? Will something like this do?

```Lua
Macro {
  area="Viewer";
  key=
    "Left Num4 ShiftNum4" -- Horizontal scroll in Text and Hex modes
    + " CtrlLeft CtrlNum4" -- Horizontal scroll by 20 in Text mode; roll the contents in Dump and Hex modes
    + " MsWheelLeft AltMsWheelLeft"; -- Translates to "Left" repeated as necessary
  action=function()
    Keys(mf.akey(1,1):gsub("Left", "Right"):gsub("Num4", "Num6"))
  end;
}
```

shmuz:
> BTW, what is the recommended formatting if I want to attach comments to individual keys? Will something like this do?

Yes, only replace `+` with `..` (Lua string concatenation operator).

MKadaner:
> only replace + with .. (Lua string concatenation operator).

Ouch! I forget about this difference every time I am writing something in Lua. Not that I write in Lua much.

Thank you for the reminder.

MKadaner:
> Anyway we do not need explicit "R" here.

Well, it works with Viewer but not with FindAll in the Editor. And I don't want even to start thinking why. That's the main reason I want this functionality to be implemented natively. When I change C++ code, I know exactly what I am doing. With macros, there is at least one (maybe more) layer of translation between me and the actual action.

Just ranting.

alabuzhev (approved):

alabuzhev:
Thank you

MKadaner:
Thank you!

PR #855, opened 2024-07-14
Improvements to horizontal scrolling in Hex Mode of Internal Viewer.

MKadaner:
## Summary
1. If Hex View is wider than the window, it can be properly scrolled horizontally with arrow keys. (The width of the Hex View depends only on the number of bytes per line and does not change with the Far window width.)
2. `Ctrl+Alt+Arrow` keys snap the number of bytes per line in Hex View to the multiples of 16 (instead of changing the number by 16).
3. Help files updated accordingly.
4. Refactoring

## Checklist

### 1. Fixed issue with horizontal scrolling of Hex View

Currently, Hex View behaves strangely in the scenario below. It is not exactly a bug but close.

1. Open `changelog` in the Internal Viewer.
2. Switch to Hex Mode (`F4`).
3. Increase Far window width to 120 columns.
4. Increase the number of bytes per line to the maximum (`Ctrl+Alt+Right` repeatedly).
5. Switch to Text Mode (`F4`).
6. Shrink Far window to 50 columns.
7. Unwrap the lines (`F2` until its legend reads `Wrap`).
8. Scroll right to column 200 (`Ctrl+Right`; watch `Col` indicator in the status bar).
9. Switch to Hex Mode (`F4`).
10. Observe that the window is somewhere in the middle of the Hex View horizontally.
11. Observe that Hex View can be scrolled left (`Left`).
12. Observe that Hex View _cannot_ be scrolled right (`Right`).

This PR fixes this issue and allows freely scrolling Hex View horizontally.

### 2. Snapping the number of bytes per line to the multiples of 16

Currently, `Ctrl+Alt+Arrow` keys simply change the number of bytes per line in Hex Mode by 16 bytes. While viewing Hex dump chopped into odd-width lines sometimes may be useful, when pressing `Ctrl+Alt+Arrow` the user is more likely wants to round the number of bytes to a 16-bytes boundary.

This PR changes the behavior of `Ctrl+Alt+Arrow` keys in Hex Mode to snap the number of bytes per line to the nearest multiple of 16 in the corresponding direction.


far/viewer.cpp
```diff
@@ -296,13 +296,29 @@ wchar_t Viewer::ZeroChar() const
...
+			if (Amount > 0) return RoundUp(m_HexModeBytesPerLine, Amount);
+
+			const auto Pivot{ RoundUp(m_HexModeBytesPerLine, -Amount) };
+			// Do the math in the inverted space X' = Pivot - X, then return to the original space.
+			return Pivot - RoundUp(Pivot - m_HexModeBytesPerLine, -Amount);
+		}();
 
-	m_HexModeBytesPerLine = NewValue;
+	const auto NewBytesPerLine =
+		std::clamp(RoundedBytesPerLine, s_HexModeBytesPerStripe, hex_mode_bytes_from_line_width(ScrollbarAdjustedWidth, get_char_size(m_Codepage)));
```
alabuzhev, line 317: > This PR changes the behavior of `Ctrl+Alt+Arrow` keys in Hex Mode to snap the number of bytes per line to the nearest multiple of 16 in the corresponding direction.

Does it?
Ctrl+Alt+Left eventually goes from 16 to 8
Ctrl+Alt+Right eventually goes from 16*N to whatever fits into the window.

This behaviour is inherited from the previous implementation of course and comes from this clamp, but I'm not sure it makes sense anymore: Ctrl+Alt+Keys sometimes snapping and sometimes doing something else is confusing.
Maybe just do nothing if we can't make it a multiple of 16?

Also, maybe reflect the snapping in help?
For some reason it's in at least two places:
https://github.com/FarGroup/FarManager/blob/8d1fc4d68092498130f02772df328a1aca279326/far/FarEng.hlf.m4#L2859-L2862
https://github.com/FarGroup/FarManager/blob/8d1fc4d68092498130f02772df328a1aca279326/far/FarEng.hlf.m4#L3080-L3085

far/viewer.cpp
```diff
@@ -296,13 +296,29 @@ wchar_t Viewer::ZeroChar() const
...
 void Viewer::ChangeHexModeBytesPerLine(int const Amount)
 {
-	const auto NewValue =
-		std::clamp(m_HexModeBytesPerLine + Amount, s_HexModeBytesPerStripe, hex_mode_bytes_from_line_width(ScrollbarAdjustedWidth, get_char_size(m_Codepage)));
+	if (Amount == 0) return;
 
-	if (NewValue == m_HexModeBytesPerLine)
-		return;
+	// Round up to the nearest multiple of AbsAmount.
+	const auto RoundUp = [](const auto BytesPerLine, const auto AbsAmount)
```
alabuzhev, line 302: RoundUp(32, 16) is 48, so this isn't actually "round up to the nearest multiple" (32 already is), more like "the next multiple". Maybe rename the function to better reflect what it does and avoid confusions with the math "round up" notation?

far/viewer.cpp
```diff
@@ -296,13 +296,29 @@ wchar_t Viewer::ZeroChar() const
 
 void Viewer::ChangeHexModeBytesPerLine(int const Amount)
 {
-	const auto NewValue =
-		std::clamp(m_HexModeBytesPerLine + Amount, s_HexModeBytesPerStripe, hex_mode_bytes_from_line_width(ScrollbarAdjustedWidth, get_char_size(m_Codepage)));
+	if (Amount == 0) return;
```
alabuzhev, line 299: I think this function is supposed to be called only with ±(1, 16), maybe just assert?

far/viewer.cpp
```diff
@@ -319,6 +335,43 @@
...
+{
+	assert(m_DisplayMode == VMT_HEX || m_DisplayMode == VMT_DUMP);
+
+	if (!ViewFile) return;
+
+	const auto CharSize = GetModeDependentCharSize();
+	const auto AlignToChar{ [&](const auto Pos) { return Pos / CharSize * CharSize; } };
+
+	const auto NewFilePos =
+		AlignToChar(std::max(std::min(FilePos + Shift * CharSize, FileSize - 1), 0LL));
```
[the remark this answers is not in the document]
alabuzhev: False positive, but since Shift can only be ±1, maybe reword it to something like
`Shift < 0? FilePos - CharSize : FilePos + CharSize`?

far/viewer.cpp
MKadaner: It works for any (reasonable) `Amount` without sacrificing readability (I hope) and performance (almost). We may want to add snapping 8-byte grid or reuse the code elsewhere. Let's keep it generic.

far/viewer.cpp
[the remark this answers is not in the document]
MKadaner: In a sense CodeQL is right. Here we are talking about (byte or character) offset in a file which should be 64 bits wide. I changed argument type (and name) accordingly.

far/viewer.cpp
MKadaner: I thought about it as well. Decided to keep the old behavior, but happy to change it.

Help: ugh! I updated both `@Viewer` and `@ViewerMode` articles. Please read carefully.

far/viewer.cpp
MKadaner: Yeah! It makes sense. I renamed it to `GetNextGridPoint`. Will be happy to rename to something better if you suggest.

alabuzhev:
Thank you

MKadaner:
Thank you!

PR #857, opened 2024-07-21
Do not change the number of bytes per line in Hex View when not in Hex mode.

MKadaner:
## Summary
It's uncouth to change appearance of something that is not visible at the moment.

## Checklist


far/viewer.cpp
```diff
@@ -1964,22 +1966,30 @@ bool Viewer::process_key(const Manager::Key& Key)
 
 		case KEY_ALTLEFT:
 		case KEY_RALTLEFT:
-			ChangeHexModeBytesPerLine(-1);
+			if (m_DisplayMode == VMT_HEX)
```
alabuzhev, line 1969: Maybe just move it into the function instead of repeating the same thing 4 times?

MKadaner: It's similar to how `HorizontalScroll` and `RollContents` work. The functions `assert` their context, while the caller is responsible to check the mode. Granted, with these two we do not have much of a choice because the key handler has to call different functions depending on the mode. With `ChangeHexModeBytesPerLine`, this is not necessary, but I wanted to keep symmetry.

alabuzhev: From the maintainer's perspective less code is better than more code, and it's not like the function is pure or something - it reads class members, it modifies class members, one more won't make it worse.

I won't insist if you prefer to keep it as is though.

MKadaner: Let's keep it as is. Kinda separation of concerns, the key handlers deal with view mode, while the functions do what they a called to do.

alabuzhev:
Thank you

PR #862, opened 2024-08-03
Add soft failover to fast search

nyq:
## Summary
Modified filelist.cpp so that if standard fast search in file panel yields no results, a second search attempt is made by searching in the middle of filename strings.
## Details
Expanded file panel fast search functionality to also search for matches in the middle of the filename strings if search in the beginning of the filename strings was unsuccessful. This is very helpful in two scenarios:

1. When no exact match in the beginning of the filename string was found, but user would like to also find matches in the middle of the filenames.
2. When directory contains files/folders that start with characters that can't be easily entered into fast search box like '[' and '{', yet the user would like to search beyond these characters.

**Specific use case that justifies this change**
A directory containing very large number sub-directories in the following format: {12345} where 12345 is a serial number. Such directlry can't be easily searched with fast search in current version of FAR because:

1. Holding Alt key and pressing "{" key results in no response from FAR (pressing "{" key without Shift generates "[" character which is filtered out)
3. Holding Alt key and pressing Shift-"{" to truly generate "{" character also results in no response from FAR
4. So the only way to search in this situation is to hold Alt key, press any other character key to bring up the fast search window, then press Shift-"{" to enter "{" into the fast search window, and then type the rest of the characters, which is very tedious.

This pull request fixes this problem by allowing the user to directly enter characters that follow the "{" character. At the same time it does not break the standard fast search functionality because if there are strict matches in the beginning of the filename strings as per classic FAR behavior, they will be found first. New code will only resort to searching in the middle of the filename strings if the classic search was unsuccessful.

### DIFF
--- a/far/filelist.cpp
+++ b/far/filelist.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
 filelist.cpp
 
 Файловая панель
@@ -3690,7 +3690,7 @@ bool FileList::FindPartName(string_view const Name,int Next,int Direct)
 		NameView.remove_suffix(1);
 	}
 
-	const auto strMask = exclude_sets(NameView + L'*');
+	auto strMask = exclude_sets(NameView + L'*');
 
 	const auto Match = [&](int const I)
 	{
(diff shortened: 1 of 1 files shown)


yegor-mialyk:
I think this behavior has to be definitely configurable and not turned on "by default".

And why not macros?

Here is a good example I personally use (by @johnd0e):
https://forum.farmanager.com/viewtopic.php?f=15&t=9246

alabuzhev:
> **Specific use case that justifies this change**
> A directory containing very large number sub-directories in the following format: {12345} where 12345 is a serial number. Such directlry can't be easily searched with fast search in current version of FAR

But it can - press `Alt+Shift+8`, then enter your serial number.

nyq:
> > **Specific use case that justifies this change**
> > A directory containing very large number sub-directories in the following format: {12345} where 12345 is a serial number. Such directlry can't be easily searched with fast search in current version of FAR
> 
> But it can - press `Alt+Shift+8`, then enter your serial number.

This is pretty burdensome when I have to do this several hundreds of times each day in the course of my work. Pressing Alt and typing a serial is much more ergonomic under heavy workload.

PR #869, opened 2024-08-27
Arclite: Updated 7z.dll to v24.08

yegor-mialyk:
## Checklist

## Details

Updated the following binaries to v24.08:

plugins/arclite/7z/dll/32/7z.dll
plugins/arclite/7z/dll/64/7z.dll
plugins/arclite/7z/dll/ARM64/7z.dll
plugins/arclite/7z/sfx/7z.sfx
plugins/arclite/7z/sfx/7zCon.sfx
plugins/arclite/7z/sfx/7zS2.sfx
plugins/arclite/7z/sfx/7zS2con.sfx
plugins/arclite/7z/sfx/7zSD.sfx

There are no changes to the source code (including 7zip) besides the version update.


alabuzhev:
I'm not sure that adding ~5 MB of binaries into git every month is healthy in the long run.
Would it work for y'all if we keep them in a separate repo instead (e.g. [here](https://github.com/FarGroup/thirdparty)) and connect it as a git submodule?

yegor-mialyk:
I'm fine with moving binaries to a separate repository (I just followed the current approach here). As for the git submodules I like a general idea behind them, but I had tons of headache supporting submodules in git, so I personally prefer to avoid using submodules and, in this particular case, I would rather pull the binaries from a separate repository during so called "packaging" step (creation of a distributive or installation files) and not during the build process.

alabuzhev:
> pull the binaries from a separate repository during so called "packaging" step and not during the build process

We still need to be able to build and test arclite locally, i.e. without proper packaging.

yegor-mialyk:
Then submodules are the only option, unless we want to have two repositories cloned/pulled side by side and have cross references.

alabuzhev:
Or something like `powershell -c "Invoke-WebRequest -Uri 'https://github.com/.../file.zip' -OutFile 'file.zip'"` in a build step / event.

alabuzhev:
#### For future upgrades:

Binaries are now here: https://github.com/FarGroup/thirdparty

Version is defined in these files:

plugins\arclite\arclite.vcxproj
plugins\arclite\makefile_gcc
plugins\arclite\makefile_vc
misc\nightly\plugins.sh

We might do something later to define it only once somewhere, but for now it will do.

yegor-mialyk:
@MKadaner, I may be exaggerating a bit, but using Git submodules is not just about referencing a different repository; it's about referencing a specific commit within that repository. Everything works fine until you need to update your reference to a more recent version. Technically, it’s not a big deal either. However, if you are modifying both repositories as part of your feature implementation and using feature branches in each, complications can arise. If you want to update the reference in your feature branch of repository 1 to match the latest commit in the feature branch of repository 2, and later merge both into the master/main branch, that's where the headache starts.

MKadaner:
Ah, yes, I see. Having a feature branch in the target repo is confusing indeed. I did it once. Fortunately, it should not be needed for the case being discussed here.

Issue #763, opened 2023-12-21
incorrect redrawing of the history window


AVBL:
### Far Manager version

3.0.6226

### OS version

win 10

### Steps to reproduce

1. Run Far (I run without plugins - "far /p").
2. Create temporary file in editor (Shift-F4 TEST Enter F2 F10).
3. Delete just created file.
4. Open files history.
![image](https://github.com/FarGroup/FarManager/assets/49921898/ae968b88-eccc-4326-adff-37249f1fee9d)
5. IMPORTANT: turn on filtering (Ctrl-Alt-F)
6. Clear history (Ctrl-R)...

### Actual behavior

History border disappear.
![image](https://github.com/FarGroup/FarManager/assets/49921898/71bedac7-ba28-4c5e-89c9-f27ad4420f02)
Press any key (for example, Up) and border appears again.

Firebie:
On latest Far 3.0.6245.0 (Private) x64 with Consolas as font after AltHistory forms:
![image](https://github.com/FarGroup/FarManager/assets/231808/024b27cc-c287-4df9-8e93-47fd437bc7ef)

MKadaner:
@alabuzhev, assign it to me please. I do not know when I will be able to get to it, but I guess I've been the last who touched this place.

MKadaner:
I cannot reproduce it. I tried official build 6353 and local debug build from the same sources. Could it have been a drawing artifact, which @alabuzhev addressed some time ago?

@AVBL, @Firebie, may I ask you to test the issue again on the latest version and, if it is still repro, update repro steps?

AVBL:
Только что проверил - ничего не изменилось. Напоминаю, перед Ctrl-R (шаг 6) нужно включать фильтра (шаг 5).

MKadaner:
Tried harder, could repro. It looks like it depends on the size of the history list window relative to that of the Far window. I'll look further; hopefully will have time during the weekend.

MKadaner:
Fixed in 6403.

MKadaner:
@AVBL, could you please test and close this issue?

AVBL:
Looks like fixed, thank you.

PR #892, opened 2024-12-19
gh-763: Fixed drawing of History menu border after refresh (`Ctrl+R`) and clear (`Del`).

MKadaner:
## Summary
Now History menu border and shadow does not disappear when the history is refreshed (`Ctrl+R`) or cleared (`Del`) with the menu filter turned on.

## References
- gh-763

## Checklist

## Details
Even after half a year of debugging (although it was very busy half a year, and I could not spend much time on it), I still cannot understand how History menu works (for that matter, I do not understand how Far windowing system works). Thus, the fix is done somewhat blindly. In my limited testing, everything seems to work and there was no perceptible performance regression. However, I will not be surprised if I broke some other subtle scenario.


alabuzhev:
Thank you

MKadaner:
Thank you!

PR #924, opened 2025-03-05
[Suggestion] Added 'Used' space and memory values to the Panel Info

yegor-mialyk:
## Summary

Added 'Used' space and memory values to the Panel Info.

 
## References

https://forum.farmanager.com/viewtopic.php?t=13600

## Checklist


far/infolist.cpp
```diff
@@ -429,9 +429,16 @@ void InfoList::DisplayObject()
...
+
 		GotoXY(m_Where.left + 2, CurY++);
 		PrintMetricText(Kind, lng::MInfoMetricAvailable);
 		PrintInfo(far::format(L"{}%, {}"sv, ToPercent(Available, Total), size2str(Available)));
+
+		const auto Used = Total - Available;
+
+		GotoXY(m_Where.left + 2, CurY++);
+		PrintMetricText(Kind, lng::MInfoMetricUsed);
+		PrintInfo(far::format(L"{}%, {}"sv, ToPercent(Used, Total), size2str(Used)));
```
MKadaner, line 441: Will `Available` and `Used` always add up to 100%? See https://godbolt.org/z/TMs6heWEh.

Maybe calculate `Used` as `100 - ToPercent(Available, Total)`? Just asking.

yegor-mialyk: Yep, I think it makes sense as it is 99% most of time.

alabuzhev:
Let's say I plan to download a huge file or run a huge program.

"Available" tells me how much I can use right now. This is helpful.
"Total" tells me how much I can use in principle if I free some resources. This is also helpful.

What's the point of "used" metrics?

yegor-mialyk:
The explorer and some other tools shows all three (maybe for illustrative purposes) and I'm fine if the PR is dropped (I added `[Suggestion]` to a title for this reason).

image

alabuzhev:
Yeah I know, I am asking mostly because I wanted to add exactly this 10 or so years ago, but could not figure out why would I want to see it, so maybe I am missing out something useful.

codemonkey493:
> could not figure out why would I want to see it

For instance when I want to backup my drive to somewhere else, I need to know how much data is in that drive.

far/infolist.cpp
alabuzhev: This basically means rounding 'used' up, which is not necessarily perfect:

image

But I have an idea.

Zeroes1:
@yegor-mialyk @alabuzhev 
I think bad idea show symbols ">" and "<" ?
ps. Also Shmuel also versus...

image

w17:
Совершенно согласен. Отвратный мультик нарисовали.
Вместо этого 1 знак после запятой в процентах был бы гораздо белее уместен и полезен.

pasha-zzz:
> Вместо этого 1 знак после запятой в процентах был бы гораздо белее уместен и полезен.

И я про то же - https://forum.ru-board.com/topic.cgi?forum=5&topic=51024&start=2660#15

MKadaner:
> This basically means rounding 'used' up, which is not necessarily perfect

@alabuzhev, will proper rounding alleviate your concern? See #924.

alabuzhev:
@MKadaner the concern was not about about precision, but about corner cases.
E.g. showing 1 byte used out of 1 TB as 1% doesn't make much sense.

I'm not sure that rounding percentages to the closest integer would make much sense either, again because of corner cases: rounding 99.5% would yield 100%, but 100% means " literally _everything_ is free / used / completed / whatever", which is not the case.

For me flooring is the most reasonable approach, even if the free/used numbers won't add up to 100, but I guess these corner cases are too rare to be concerned about them anyway.

MKadaner:
@alabuzhev, then maybe round both "Used" and "Free" down and let them add up to 99% sometimes? The greater / less signs look strange indeed.

Yet another approach is to never allow 0% or 100% unless the numerator is actually zero or equals to denominator. It will address the one byte of one terabyte concern.

Issue #918, opened 2025-02-17
Keep line numbers in Find All dialog always visible


MKadaner:
Folling up the side discussion in #789:

@HamRusTal said:

> @MKadaner Do you have any idea how feasible it is to make the line numbers and their vertical separator line stay put on the left when scrolling the menu items? As they were scrolling all in accord, that looked OK but now when the lines are aligned by the found text, these prefix fragments are scattered all around and look quite confusing IMHO.
At least some menus have non-scrollable portions of their items already: e.g. check marks in history menus, A (ANSI) marks in plugin configuration menu, etc. Maybe this can be generalized and reused in the menu at hand?

@MKadaner said:

> @HamRusTal, this behavior is a damn nuisance I have been thinking about for years. Actually, even before my change, one could align all items to the right (`Alt+End`) and line numbers would become scattered all over the place. I was enjoying this sight pretty often when I used `Alt+End` as a poor man's substitute for aligned annotations. As such, I hold that I did not break anything. However, I admit that I did not realize it could become more of an issue for somebody else.
>
> For me personally, aligning annotations has much higher priority than scattered line numbers, so it went first. Now aligning is done, I am warming up for line numbers. I can see at least two solutions:
>
> 1. Let each list item have more than one content string and draw these strings within their own vertical boundaries set up for the entire list.
> 2. Create a new dialog called `Grid` which would be similar to `VMenu2` but would have more than one list. Vertical scrolling should affect all lists synchronously, while horizontal scrolling should be applied to the focused list only. Maybe just extend `VMenu2` to contain more than one `VMenu`.
>
> The resulting UX will be more or less the same. As it usually happens, each option has its own pros and cons. So, stay tuned.

_Originally posted by @MKadaner in https://github.com/FarGroup/FarManager/issues/789#issuecomment-1943122080_

MKadaner:
Question to everyone who is interested in this feature. @HamRusTal.

If line numbers stay put at the left side of the Find All menu, (while the text still can be scrolled horizontally), it will be very natural to collapse (and restore) the column with line numbers. It is even imaginable to resize this column, because why not.

1. Which accelerator key should be used to collapse / restore?
2. If resize is supported, how to resize from the keyboard? I considered an accelerator which puts the menu in the special mode (similar to turning on filtering with `RAlt`), then use arrow key to resize. What should this accelerator be?

HamRusTal:
[...]

PR #927, opened 2025-03-16
gh-918: Introduced type `segment`

MKadaner:
## Summary
Using `segment` instead of `std::pair<int, int>` (or similar) because there is no way to remember what the `second` item represents: length, end, or last.

## References
- gh-918

## Checklist

## Details
I started working on #918 and soon realized that I am totally confused by the meaning of various `std::pair`s I introduced in #789. Hence this PR. I am going to use `segment` even more.


far/common/2d/segment.hpp
```diff
@@ -0,0 +1,93 @@
...
+template<typename T>
+struct segment_length_t
+{
+	T length;
+
+	explicit operator T() const noexcept { return length; }
+};
+
+template<typename T>
+struct segment_t
```
alabuzhev, line 47: This looks suspiciously similar to `std::ranges::subrange`.
Can we just use that?

MKadaner: I looked at `std::ranges::subrange`. While it is probably possible to bend it to the purpose, it has a few properties that make it less suitable.

1. Subrange is, well, a subrange. It strongly suggests that somewhere there is an underlying range. It has `data`, `front`, `back`, and `operator[]`. In the contexts where I am going to use `segment`, it is not always easy to tell what the sequence the `segment` is a slice of. The type I need is a mere arithmetic abstraction supporting convenient operations like `intersect`, rather than a representation of an iterable sequence of real elements.
2. The type of the `size()` method is, of course, `size_t`, which will play poorly with all other signed arithmetic where `segment` is be used. (And the arithmetic there should be signed because the strings of text we manipulate can extend beyond the left boundary of the areas where they rendered.)
3. The `std::ranges::subrange` maintains an iterator, ["The subrange class template combines together an iterator and a sentinel into a single view."](https://en.cppreference.com/w/cpp/ranges/subrange) It is tricky to persuade the compiler to accept an integral type as an iterator, ["you cannot dereference an operand of type 'int'."](https://godbolt.org/z/sEjvMvf65)

alabuzhev: Ah right, it's not really a range but a pair of offsets.
In this case yes, it's fine.

For signed arithmetic there's std::ssize now, maybe it can be helpful in some places.

far/common/2d/segment.hpp
```diff
@@ -0,0 +1,93 @@
+﻿#ifndef SEGMENT_HPP_DC38F7F7_0E1B_486E_8368_2ACE6415D234
+#define SEGMENT_HPP_DC38F7F7_0E1B_486E_8368_2ACE6415D234
+#pragma once
+
+/*
+segment.hpp
+
+*/
+/*
+Copyright © 2018 Far Group
```
alabuzhev, line 10: 2018 -> 2025.

Why is it in 2d? Isn't it 1-dimentional conceptually?

far/common/2d/segment.hpp
```diff
@@ -0,0 +1,93 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+//----------------------------------------------------------------------------
+
+template<typename T>
+struct segment_length_t
```
alabuzhev, line 39: Supposedly this is to resolve ambiguous ctors (Begin, End) vs (Begin, Length)?
Do we actually need both?
Looks like (Begin, Length) is used only in one place now.

far/common/2d/segment.hpp
MKadaner: > For signed arithmetic there's std::ssize now, maybe it can be helpful in some places.

Thank you for the reminder. I'll see where it can be helpful. It's probably a good idea to switch everything consistently to `std::ssize`, but:

1. It's a completely different story.
2. The return type is still very wide. If we go this way, we should switch everything to 64-bit integers (on x64), otherwise we have the same plague of `static_casts`.

far/common/2d/segment.hpp
MKadaner: Correct. So far, yes, it's used only in the `Editor`. But it makes a good concept (pun unintended). A `segment` can be considered as either (start, end) or (start, length), and one should never be limited to one particular representation, neither in accessors, nor in constructors. If you think it proper, I can add `begin()` and `end()` methods to make it fully symmetrical.

Let's keep it.

far/common/2d/segment.hpp
MKadaner: > 2018 -> 2025.

Ough! Sure, will fix it.

Yes, it is technically 1d. I could not think of a good place for it. Does it make sense to move it up to "common"?

far/common/2d/algorithm.hpp
```diff
@@ -66,4 +67,19 @@ static consteval auto column_major_iota()
 	return Result;
 }
 
+template<typename TA, typename TB>
+segment_t<std::common_type_t<TA, TB>> intersect(segment_t<TA> const A, segment_t<TB> const B)
+{
+	if (A.empty() || B.empty())
```
alabuzhev, line 73: I know this is not new but moved from vmenu, but just in case:
- segment_t::operator== is defaulted, so it's a piecewise comparison that takes into account both begin and end.
- here we return {} (i.e. {0, 0} when there is no intersection.
- is {0, 0} a special sentinel segment to indicate no intersection or is it conceptually equal to, say, {42, 42}?
- I'm not a set theory expert, but it feels like empty segments can intersect differently too? E.g. the intersection of {1, 1} and {1, 1} is *probably* {1, 1} (since they overlap, even if empty), but the intersection of {1, 1} and {2, 2} is neither of them, e.g. {0, 0}?

far/common/2d/segment.hpp
alabuzhev: Yes, why not.

far/common/2d/algorithm.hpp
MKadaner: I am not a mathematician either, but I think the following axioms would not cause any contradictions down the road:

1. All empty segments are equivalent.
2. An empty segment has no ends.

On the practical level, it means that all accessors should have a precondition (`assert(!empty()`), and `operator==` should either reject empty segments as well or, according to Axiom 1, compare all of them equal. I think the latter is more correct.

Even more practically, what else can be returned if segments do not intersect? And if two empty segments happen to have the same `begin`, they are still empty, the intersection is empty, all empties are the same, so why not?

I do not think it worth cramming in `std::optional` to represent empty segments (though it would be more correct), mostly because of implementation and mental overhead. Also, please think of `std::optional<small_segment> menu_layout::TextArea`. There is nothing wrong with `std::optional<std::optional<small_segment>>`, but... Ahem...

If you do not mind, let me add missing accessors and `operator==` and `assert(!empty()` as appropriate.

alabuzhev: Thanks.
Just to clarify, I'm not trying to be picky - it's all mostly theoretical and probably irrelevant for all the current usages, but we're moving it to common where it might get used by something entirely different (also, do we actually expect it to be used elsewhere?), so it should hold some water.

MKadaner: No worries, I share the concern, and I am working on some generalization and clean up, while trying to avoid overengineering. Stay tuned.

MKadaner: Please look at the updated implementation. Hopefully, I hit the right balance.

The code in vmenu.cpp already became cleaner, and I look forward to using `segment` with fixed left columns.

far/common/segment.hpp
```diff
@@ -0,0 +1,113 @@
...
+public:
+	segment_t() noexcept = default;
+
+	segment_t(T const Start, segment_sentinel_t<T> const End) noexcept
+		: segment_t{ Start, static_cast<T>(End) }
+	{}
+
+	segment_t(T const Start, segment_length_t<T> const Length) noexcept
+		: segment_t{ Start, Start + static_cast<T>(Length) }
+	{}
```
MKadaner, line 66: I think it's a good idea to force the user to be explicit when creating `segment`s. It's a bit mouthful but eliminates guesswork when you see or use the `(int, int)` constructor.

far/common/segment.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+	segment_t() noexcept = default;
+
+	segment_t(T const Start, sentinel_tag const End) noexcept
+		: segment_t{ Start, End.m_end }
+	{}
+
+	segment_t(T const Start, length_tag const Length) noexcept
+		: segment_t{ Start, static_cast<T>(Start + Length.m_length) }
+	{}
```
MKadaner, line 53: I think it's a good idea to force the user to be explicit when creating segments. It's a bit mouthful but eliminates guesswork associated with the (int, int) constructor.

far/vmenu.cpp
```diff
@@ -71,6 +71,7 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "common.hpp"
 #include "common/scope_exit.hpp"
 #include "common/uuid.hpp"
+#include "common/2d/algorithm.hpp"
```
alabuzhev, line 74: Probably without 2d now?

MKadaner: Good catch, thank you.

alabuzhev:
Thank you

MKadaner:
Thank you!

PR #932, opened 2025-03-25
luafar & luamacro: apply __tostring to error object

johnd0e:
The initial goal of the proposed changes is to correctly display error messages for objects that have a `__tostring` metamethod, which needs to be called safely to obtain the final message.
All errors that might occur during this process are handled and outputted consistently with the standard interpreter's behavior (the error text was checked against Lua 5.4).

The changes affected `pcall_msg` (in `luafar/lf_exported.c`) and `FixReturn` (in `luamacro/luamacro.lua`). Additionally, common helper functions were implemented (in `luafar/lf_string.c`):

*   `luaL_tolstring` - A copy of the function which has been [standard](https://www.lua.org/manual/5.2/manual.html#luaL_tolstring) since Lua 
5.2.
    It is used in: `far.MakeMenuItems`, `far.Message`, `win.OutputDebugString`, and `far.MakeMenuItems` (as a helper for `far.Show`). This cleaned up "copy-pasted" code in these areas and ensured consistent handling.
*   `safe__tostring_meta`: Used for error output in `pcall_msg` and is therefore protected (i.e., handles potential errors during the `__tostring` call safely).
    <details>

    ```c
    /**
     * @brief Safely calls the __tostring metamethod for a value on the Lua stack.
     *
     * Attempts to retrieve and call the __tostring metamethod for the Lua value
     * at the given stack index `idx`.
     *
     * Stack effects:
     * - If no __tostring metamethod exists: Returns 0, stack unchanged.
     * - If __tostring call succeeds and returns a string: Returns 1, leaves the
     *   resulting string on top of the stack.
     * - If __tostring call fails (errors): Returns -1, leaves an error message
     *   string on top of the stack. The original error object is replaced with
     *   a formatted string if it wasn't a string itself.
     * - If __tostring call succeeds but returns a non-string: Returns -1, leaves
     *   the error message "'__tostring' must return a string" on top of the stack.
     *
     * The original value at `idx` remains untouched on the stack below the result/error.
     *
     * @param L The Lua state.
     * @param idx The stack index of the value whose __tostring metamethod should be called.
     * @return 1 on success (string pushed),
     *         0 if no __tostring metamethod exists (stack unchanged),
     *        -1 on error (__tostring errored or returned non-string; error string pushed).
     */
    ```
    </details>

*   `safe_luaL_tolstring`: the same as the `luaL_tolstring`, but protected (uses `safe__tostring_meta` under hood).
    On error:  error message is pushed onto the stack, and NULL is returned.
    Used in `win.WriteConsole`, which allows obtaining meaningful error messages. This is used, in particular, in `mf.printconsole`: the function now no longer ignores errors during string conversion.

### DIFF
--- a/plugins/luamacro/luafar/lf_string.c
+++ b/plugins/luamacro/luafar/lf_string.c
@@ -1,42 +1,83 @@
 ﻿#include <shlobj.h>
 #include "lf_string.h"
 
-const char *safe_luaL_tolstring(lua_State *L, int idx, size_t *len)
+// initially from: https://www.lua.org/source/5.2/lauxlib.c.html#luaL_tolstring,
+// but additionally throws on invalid __tostring return values, like in Lua 5.3
+const char *luaL_tolstring(lua_State *L, int idx, size_t *len)
 {
-	if (luaL_getmetafield(L, idx, "__tostring")) {
+	// partly from: https://www.lua.org/source/5.3/lauxlib.c.html#luaL_tolstring
+	if (luaL_callmeta(L, idx, "__tostring"))  /* metafield? */
+	{
+		if (!lua_isstring(L, -1))
+			luaL_error(L, "'__tostring' must return a string");
+	}
+	else
+	{
+		switch (lua_type(L, idx))
+		{
+			case LUA_TNUMBER:
+			case LUA_TSTRING:
+				lua_pushvalue(L, idx);
+				break;
+			case LUA_TBOOLEAN:
+				lua_pushstring(L, (lua_toboolean(L, idx) ? "true" : "false"));
+				break;
+			case LUA_TNIL:
+				lua_pushliteral(L, "nil");
+				break;
+			default:
+				lua_pushfstring(L, "%s: %p", luaL_typename(L, idx), lua_topointer(L, idx));
+				break;
+		}
+	}
+	return lua_tolstring(L, -1, len);
+}
+
+// noop if the value has no __tostring metamethod
+// otherwise it is called and the result string (or error msg) is pushed onto the stack
+ToStringResult safe__tostring_meta(lua_State *L, int idx)
+{
+	if (luaL_getmetafield(L, idx, "__tostring"))
+	{
 		lua_pushvalue(L, idx);
-		if (lua_pcall(L, 1, 1, 0) != 0) {
-			if (!lua_isstring(L, -1)) {
+		if (lua_pcall(L, 1, 1, 0) != 0)
+		{
+			if (!lua_isstring(L, -1))
+			{
 				const char* tname = luaL_typename(L, -1);
 				lua_pop(L, 1);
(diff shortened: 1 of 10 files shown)


johnd0e:
`luaL_tolstring` implementation is from Lua 5.2: https://www.lua.org/source/5.2/lauxlib.c.html#luaL_tolstring

I have no idea why but there is no protection from wrong (non-function) `__tostring` value, as well as from runtime errors.

johnd0e:
I've implemented relevant error handling, but it's not clear how this will behave with original `luaL_tolstring` under Lua 5.2+.
So perhaps it's better to just rename our function and do not rely on build-in `luaL_tolstring` at all.

johnd0e:
@shmuz

johnd0e:
Related issues:
- `tostring` usage is not consistent across sources - in other places lua native function is used (`lua_getglobal(L, "tostring");`)
  In order to handle errors in consistent manner it would be better to have some common handling.
- `far.Show` crashes the Far if `tostring` pcall failed with non-string error object.
  To reproduce just run `luas:=setmetatable({},{__tostring=error})`
  https://github.com/FarGroup/FarManager/blob/183c32abb2a25d0b9015ebd1630d891e3eabedb2/plugins/luamacro/luafar/lf_service.c#L5348-L5349
- `win.WriteConsole` ignores such cases, returning 0 without additional details.
  It would be more helpful if it would throw, or return `nil, "error message" instead.
  https://github.com/FarGroup/FarManager/blob/183c32abb2a25d0b9015ebd1630d891e3eabedb2/plugins/luamacro/luafar/lf_win.c#L805-L806
- mf.printconsole relies on `win.WriteConsole`, but not check it's return value, so user is not even getting aware of errors.
  https://github.com/FarGroup/FarManager/blob/183c32abb2a25d0b9015ebd1630d891e3eabedb2/plugins/luamacro/api.lua#L627-L629
- possibly more

shmuz:
@johnd0e, please add the usual version increment and changelog records and then it can be merged.

plugins/luamacro/luafar/lf_string.c
```diff
@@ -1,6 +1,84 @@
...
+// on error: error message is pushed onto the stack, and NULL is returned.
+const char *safe_luaL_tolstring(lua_State *L, int idx, size_t *len)
+{
+	switch (safe__tostring_meta(L, idx))
+	{
+		case TOSTRING_NOMETA:
+			return luaL_tolstring(L, idx, len);
+		case TOSTRING_SUCCESS:
+			return lua_tolstring(L, -1, len);
+		case TOSTRING_ERROR:
```
alabuzhev, line 77: Please either add the default case or return something after the switch.
Right now this [breaks](https://github.com/FarGroup/FarManager/actions/runs/15795030577/job/44525952885?pr=932#step:8:1119) the build.

johnd0e: Done

PR #935, opened 2025-04-06
gh-918: Keep found position in Find All dialog always visible

MKadaner:
## Summary
- Now it's possible to add fixed columns at the left side of Lists.
- Used fixed columns in Find All dialog to keep found position always visible.
- Fixed hot key auto-assignment (because it happened to be on the way).

## References
- gh-918

## Checklist

## Details
@alabuzhev, @HamRusTal, please review / try it. It's ready.

The basic idea is to keep the contents of the fixed columns in the item strings, (almost) as it was before. This approach saves memory allocations for the contents of each fixed column in each item. The user specifies the positions of the columns using the function

```C++
void SetFixedColumns(std::vector<vmenu_fixed_column_t>&& FixedColumns, small_segment ItemTextSegment);
```

Where `vmenu_fixed_column_t` is defined as

```C++
struct vmenu_fixed_column_t
{
	small_segment TextSegment;
	short CurrentWidth;
	wchar_t Separator;
};
```

Each fixed column is described by the segment of the item string where the column content is located, the separator which should be displayed to the right of the column content, and the current column width on the screen. The fixed columns may be displayed (truncated as necessary) within any number of positions from 1 to the full width of the column segment. If the current width is zero, the column is hidden (along with its separator). Currently, the column can be either hidden or displayed in the full width.

The second parameter, `ItemTextSegment`, defines the position of the main item text and is stored in `m_ItemTextSegment`. See the example of calling `SetFixedColumns` in the `find_all_list::make_ready()` function in the `editor.cpp`.

`VMenu` maintains the current state of the fixed columns in `std::vector<vmenu_fixed_column_t> m_FixedColumns`. Only `CurrentWidth` of each column descriptor can be changed. See `bool ToggleFixedColumns()` function which changes visibility of the fixed columns upon pressing `Shift+F5`.

New field `std::optional<small_segment> FixedColumnsArea` is added to `menu_layout`. It is recalculated before each draw based on `m_FixedColumns` and describes the actual screen position where all fixed columns are to be displayed.

Two new functions, `DrawFixedColumns` and `DrawItemText`, were introduced to keep `DrawRegularItem` more manageable. The algorithm of drawing item text with highlights did not change. However, now highlights (both accelerator keys and annotations) are applicable only to the main item text and their positions are relative to `m_ItemTextSegment`. Highlights in the fixed columns are not supported.

Most remaining changes are required to properly account for the fact that item text, along with highlights, is now located within `m_ItemTextSegment` instead of occupying the entire item string as it was before. Notably, careful adjustments were necessary in the `AssignHighlights` function, so I refactored it to make the changes easier. I also slightly changed the behavior of this function to work per the original intent (as I understand it).

Hope this description will help with the reviewing this substantial change.


far/headers.hpp
```diff
@@ -82,6 +82,7 @@ WARNING_POP()
 #include <forward_list>
 #include <fstream>
 #include <functional>
+#include <generator>
```
MKadaner, line 85: Som compilers seem to be unhappy with `generator`. Can we upgrade them? If not, I'll fix it.

far/editor.cpp
```diff
@@ -3395,6 +3395,100 @@ int Editor::CalculateSearchNextPositionInTheLine(const bool Backward, const bool
...
+			const short LineNumColumnWidth{ radix10_formatted_width(m_MaxLineNum) };
+			const short FoundPosColumnWidth{ radix10_formatted_width(m_MaxFoundPos) };
+			const short LineNumColumnStart{ m_LineNumColumnMaxWidth - LineNumColumnWidth };
+			const short FoundPosColumnStart{ m_LineNumColumnMaxWidth + m_FoundPosColumnMaxWidth - FoundPosColumnWidth };
+			const short ItemTextStart{ m_LineNumColumnMaxWidth + m_FoundPosColumnMaxWidth };
+			m_Menu->SetFixedColumns(
+				{
+					{
+						.TextSegment{ LineNumColumnStart, small_segment::length_tag{ LineNumColumnWidth } },
+						.CurrentWidth{ LineNumColumnWidth },
```
MKadaner, line 3451: Clang seems to be unhappy with curly braces around scalar initializer. Is it possible to enable this syntax [-Wbraced-scalar-init]. If not, I'll fix the code.

MKadaner: Actually, replaced curly braces with `=`.

far/headers.hpp
MKadaner: OK. Let's do without.

MKadaner:
Since the change is large, I expect many comments, in more than one wave. To allow for incremental code review, I will be adding commits instead of squashing with every round of fixes. When everything looks good, I will finally squash-rebase.

far/dialog.cpp
```diff
@@ -2561,6 +2555,7 @@ bool Dialog::ProcessKey(const Manager::Key& Key)
 			case KEY_NUMENTER:
 			case KEY_ENTER:
 			case KEY_NUMPAD5:
+			case KEY_F5:
```
alabuzhev, line 2558: Why F5?
The PR mentions Shift+F5, so seems to be unrelated to columns.

far/editor.cpp
```diff
@@ -3395,6 +3395,100 @@ int Editor::CalculateSearchNextPositionInTheLine(const bool Backward, const bool
...
+	}
+
+	class find_all_list
+	{
+	public:
+		explicit find_all_list(int MaxLinesCount)
+			: m_LineNumColumnMaxWidth{ radix10_formatted_width(MaxLinesCount) }
+		{}
+
+		void add_item(FindCoord FoundCoords, const string& ItemText)
```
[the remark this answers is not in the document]
MKadaner: Yes, it can. I changed it.

I am not a fan of `std::string_view`. I think it is overused, and sometimes even misused way too often. The main problem with it, is that it's too easy to get a dangling `std::string_view`. One of the questions I often ask at interviews is a rather simple and naturally looking one-liner which ends up in a dangling `std::string_view`. And do not let me start on the code where `std::string_view::data()` is assumed to return a null-terminated string because somebody was overzealous when he was switching the codebase to passing around shiny new `std::string_view` instead of rusty old `const std::string&`.

I think that most legitimate uses of `std::string_view` should be confined to various library APIs where it provides greater flexibility to callers (which was the main intent behind `std::string_view`). I further think that in application code `std::string_view` should be used sparingly and with great consideration.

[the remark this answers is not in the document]
alabuzhev: >  it's too easy to get a dangling `std::string_view`

Indeed it is, especially in cases like `string_view s = something_that_returns_string_by_value();`
However, it's usually perfectly fine as a function parameter (like here).

far/common.tests.cpp
```diff
@@ -1200,7 +1200,7 @@ TEST_CASE("segment.iota")
 		std::initializer_list<int> Expected;
 	};
 
-	static const test_data TestDataPoints[] =
+	static const auto TestDataPoints = std::to_array<test_data>(
```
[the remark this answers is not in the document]
MKadaner: Happy to. I never know.

far/dialog.cpp
```diff
@@ -766,9 +766,6 @@ void Dialog::InitDialogObjects(size_t ID, bool ReInit)
 				ListPtr->ChangeFlags(VMENU_AUTOHIGHLIGHT, (Item.Flags & DIF_LISTAUTOHIGHLIGHT) != 0);
 				ListPtr->ChangeFlags(VMENU_NOMERGEBORDER, (Item.Flags & DIF_LISTNOMERGEBORDER) != 0);
 
-				if (Item.Flags & DIF_LISTAUTOHIGHLIGHT)
```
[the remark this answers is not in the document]
alabuzhev: Ok, sounds reasonable.

far/vmenu.cpp
```diff
@@ -229,7 +230,7 @@ struct menu_layout
 	{
 		// ╚═ Ctrl+Enter F5 Gray + Ctrl+Up Ctrl+Down ══╝
 		// ?^^                                      ^ ^?
-		return NeedBox + 1 + 1 + 1 + 1 + NeedBox;
+		return NeedBox ? 6 : 4;
```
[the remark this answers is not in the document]
MKadaner: Sonar's complaint was about inappropriate use of `bool` in an integer expression. There was one more place. Let me revert it as well. Please let me know if that is OK also.

far/FarEng.hlf.m4
```diff
@@ -340,6 +340,10 @@ $ #Menu control commands#
  #Alt+Shift+End#
  Align the selected item to the right.
 
+ #Shift+F5#
+ Some menus, e.g., ~Internal Editor~@Editor@'s ~All matching entries menu~@FindAllMenu@, have one or more
```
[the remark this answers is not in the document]
MKadaner: Sure! Fixed.

far/vmenu.cpp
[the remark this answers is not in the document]
alabuzhev: It's ok.

far/dialog.cpp
MKadaner: None of the `case`s above include any modifiers. I originally thought (without looking anywhere) that `LocalKey()` somehow strips off the modifiers, but apparently it is not.

The code below seems to be about ensuring menu redraw when the selected item was changed by the key. I am not sure why it should be done from the dialog and not in the menu itself. If indeed the intent was to handle redraw on selection change, neither `KEY_NUMPAD5` nor `KEY_(SHIFT)F5` should be here because they cannot change the selection. And when I comment out these two `case`s, things _seem_ to work fine.

I did not dig deeper and changed `KEY_F5` to `KEY_SHIFTF5`. This place should probably be revisited, but not in this PR if possible. Please let me know.

alabuzhev: Looks like KEY_NUMPAD5 [was added by you](https://github.com/FarGroup/FarManager/commit/bf0d80386e38c19fb0000d2e97d991587c241797#diff-6e956c0db925004801c1509ab2884997c52d3bc9f5f44c9af810b204d6a08a55R2551) as well :) for purposes Ctrl+Numpad5.

The key is definitely not stripped of any modifiers here.
Since these keys alone have nothing to do with the intended actions and, as you say, things seem to work fine without them, may I suggest removing both KEY_NUMPAD5 and KEY_SHIFTF5 in this PR?
We probably have way too much dead code already.

MKadaner: Done!

alabuzhev:
@MKadaner looks like column widths are not always calculated correctly:

image
Adding another line fixes it, probably an off-by-one error somewhere?

far/FarCze.hlf.m4
```diff
@@ -341,6 +341,10 @@ $ #Menu control commands#
  #Alt+Shift+End#
  Align the selected item to the right.
 
+ #Shift+F5#
+ Some menus, e.g., ~all matching entries~@FindAllMenu@ in ~internal editor~@Editor@, have one or more
+fixed columns at the left. The #Shift+F5# key combination toggles visibility of these columns.
```
alabuzhev, line 346: Looks like Shift+F5 toggles it in a more granular way, at least in find all: 0 - 1 - 2 - 0...
If this is intentional and the logic is "each press shows one more column" - maybe describe it here briefly?

alabuzhev:
@MKadaner what are the benefits of turning off the columns?
It doesn't free any extra screen space for menu text and it's not persisted, so, as a user, why would I want to press Shift+F5?

far/FarCze.hlf.m4
MKadaner: I thought about it... It's somewhat tedious to describe, while very easy to see and understand. I intentionally did not go into these details. Technically, my wording is correct. It toggles visibility of these columns. Nobody said which columns. 😀

If you think it is worth it, I'll add something. LMK please.

MKadaner:
> It doesn't free any extra screen space...

But it does. Toggling the columns does not shift the content horizontally, but, e.g., `Alt+Home` will move all items to the left so they touch the left edge. I tried to move the items automatically as the fixed columns become wider or narrower, but it looks ugly in all cases except for when all items are already aligned to the left. Let me see if I can handle this scenario specifically, without breaking everything else.

far/FarCze.hlf.m4
alabuzhev: Ok, fair enough.

alabuzhev:
> `Alt+Home` will move all items to the left

Ah, I totally forgot that they are movable, thanks.

MKadaner:
> [L]ooks like column widths are not always calculated correctly

Argh! Not only lines / columns are counted from 1 on screen, but also the maximums should be incremented.

Thank you for noticing it! Fixed.

alabuzhev (approved):

MKadaner:
> ...except for when all items are already aligned to the left...

I added special treatment for the all-items-are-at-home case. There is still one subtle scenario which looks a bit surprising. If people complain (or I am annoyed), I'll address it separately. The implementation is a bit involved.

alabuzhev:
Thanks

MKadaner:
Thank you!

P.S. Could you please apply this tiny fix?

```diff
diff --git a/far/editor.cpp b/far/editor.cpp
index d6821d0b2..beb5d85a1 100644
--- a/far/editor.cpp
+++ b/far/editor.cpp
@@ -3424,8 +3424,8 @@ namespace
 				m_LastSeenLine = FoundCoords.Line;
 				m_UniqueLineCount++;
 			}
-			m_MaxLineNum = std::max(m_MaxLineNum, FoundCoords.Line + 1);
-			m_MaxFoundPos = std::max(m_MaxFoundPos, FoundCoords.Pos + 1);
+			m_MaxLineNum = std::max(m_MaxLineNum, FoundCoords.Line);
+			m_MaxFoundPos = std::max(m_MaxFoundPos, FoundCoords.Pos);
 		}
 
 		void make_ready()
@@ -3439,8 +3439,8 @@ namespace
 			m_Menu->SetHelp(L"FindAllMenu"sv);
 			m_Menu->SetId(EditorFindAllListId);
 
-			const short LineNumColumnWidth{ radix10_formatted_width(m_MaxLineNum) };
-			const short FoundPosColumnWidth{ radix10_formatted_width(m_MaxFoundPos) };
+			const short LineNumColumnWidth{ radix10_formatted_width(m_MaxLineNum + 1) };
+			const short FoundPosColumnWidth{ radix10_formatted_width(m_MaxFoundPos + 1) };
 			const short LineNumColumnStart{ static_cast<short>(m_LineNumColumnMaxWidth - LineNumColumnWidth) };
 			const short FoundPosColumnStart{ static_cast<short>(m_LineNumColumnMaxWidth + m_FoundPosColumnMaxWidth - FoundPosColumnWidth) };
 			const short ItemTextStart{ static_cast<short>(m_LineNumColumnMaxWidth + m_FoundPosColumnMaxWidth) };
```

alabuzhev:
Applied.

MKadaner:
Thank you. Appreciated.

HamRusTal:
[Looks like a fallout from these changes…](https://forum.farmanager.com/viewtopic.php?p=180301#p180301)

MKadaner:
Replied on the forum.

PR #942, opened 2025-04-29
Fix incorrect item selection when opening combobox

johnd0e:
**Combobox items:**
```
test1
test2
test
```
If the value of the combobox itself is `test`, upon opening it, I expect the corresponding element to be active.
In reality, the first item is active.

The problem lies in the function `int VMenu::FindItem` (in the API, this corresponds to `DM_LISTFINDSTRING`):

The proposed fix should solve the problem with the combobox and not affect anything else in Far, as `LIFIND_EXACTMATCH` is used in only one place internally.

As for the plugin API, the flag name itself implies an exact match.
Nevertheless, in the [Encyclopedia](https://api.farmanager.com/ru/structures/farlistfind.html), it is stated verbatim about the `LIFIND_EXACTMATCH` flag:

> `LIFIND_EXACTMATCH`: Pattern is not a pattern - treat as a regular search **substring**.

This also doesn't match the current behavior; currently, substring matching is strictly done "from the beginning" (`starts_with_icase`).

Given the inconsistency, I propose applying the correction to use `equal_icase`. If necessary, flags for defining other behaviors can be added later.

Forum discussion: https://forum.farmanager.com/viewtopic.php?p=180275#p180275


far/vmenu.cpp
```diff
@@ -3296,7 +3296,7 @@ int VMenu::FindItem(int StartIndex, string_view const Pattern, unsigned long lon
 
 			if (Flags&LIFIND_EXACTMATCH)
 			{
-				if (starts_with_icase(strTmpBuf, Pattern))
+				if (equal_icase(strTmpBuf, Pattern))
```
alabuzhev, line 3299: Let's just do `if (strTmpBuf == Pattern)` here.
"Exact match" should mean exact.
We can add case-insensitive flags later if needed.

johnd0e: Done

alabuzhev: Looks good, pls update build and changelog.

johnd0e: Updated

alabuzhev:
Thanks

PR #946, opened 2025-05-04
gh-945: Edit box histories should not be saved on cancelled DN_CLOSE

### LINKED ISSUE #945: Edit box histories should not be saved on cancelled DN_CLOSE
### Far Manager version

3.0.6471

### OS version

10

### Other software

_No response_

### Steps to reproduce

```lua
local F = far.Flags
local width = 78
local items = {
  {F.DI_SINGLEBOX, 0,0,77,3,0,0,0,0, "test"},
  {F.DI_EDIT,      2,1,68,1,0,"test",0,F.DIF_HISTORY, "test"},
  {F.DI_BUTTON,   70,1, 0,1,0,0,0,0, "OK"},
}
far.Dialog(id, -1, -1, 78, 3, nil, items, 0, function (hDlg, Msg, idx, Param2)
  if Msg==F.DN_CLOSE then
    return idx<0 and true or false
  end
end)
```

1. Input some value
2. Press Enter
3. The value is now in the history

### Expected behavior

I expect that histories should not be saved before actual dialog close.

### Actual behavior

Instead, the histories are saved regardless whether is dialog actually closed or the action is cancelled.

johnd0e:
Fix #945


far/dialog.cpp
```diff
@@ -4229,12 +4229,12 @@ void Dialog::Process()
 
 intptr_t Dialog::CloseDialog()
 {
-	GetDialogObjectsData();
-
 	const auto result = DlgProc(DN_CLOSE, m_ExitCode, nullptr);
 	if (!result)
 		return 0;
 
+	GetDialogObjectsData();
```
alabuzhev, line 4236: This function does some other stuff, e.g. propagates some data from EditPtr and ListPtr back to dialog items.
DN_CLOSE handlers might expect that, so we can't just move the whole thing.

Could you please try the attached? (and update the PR if it works as expected)
[946.diff.txt](https://github.com/user-attachments/files/20040334/946.diff.txt)

johnd0e: Works like a charm

alabuzhev:
Thanks

PR #951, opened 2025-05-08
fix: Update editable combobox behavior in dialogs

johnd0e:
## Summary

Only update combobox edit text automatically when DIF_DROPDOWNLIST flag is set.

Editable comboboxes now maintain their text value when list items are modified programmatically, which prevents unwanted text changes in the edit field.

## References

Original bug report: https://forum.farmanager.com/viewtopic.php?t=13631

## Test script

```lua
local F = far.Flags
--local width = 78
local list = {
  {Text="test1"},
  {Text="test2"},
}
local items = {
  {F.DI_SINGLEBOX, 0,0,77,5,0,   0,0,0, "test"},
  {F.DI_COMBOBOX,  2,1,68,1,list,0,0,0, "test"},
  {F.DI_BUTTON,   70,2, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTADD"},
  {F.DI_BUTTON,   70,2, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTADDSTR"},
  {F.DI_BUTTON,   70,2, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTDELETE"},
  {F.DI_BUTTON,   70,2, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTINSERT"},

  {F.DI_BUTTON,   70,3, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTSET"},
  {F.DI_BUTTON,   70,3, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTSORT"},
  {F.DI_BUTTON,   70,3, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTUPDATE"},
  
  {F.DI_BUTTON,   70,3, 0,1,0,   0,0,F.DIF_BTNNOCLOSE+F.DIF_CENTERGROUP, "DM_LISTSETCURPOS"},
}
local ret = far.Dialog(id, -1, -1, 78, 5, nil, items, 0, function (hDlg, Msg, idx, Param2)
  if Msg==F.DN_BTNCLICK then
    local dm = F[items[idx][10]]
    if dm==F.DM_LISTADD then
      hDlg:ListAdd(2, {{Text="test3"}})
    elseif dm==F.DM_LISTADDSTR then
      hDlg:ListAddStr(2, "test3")
    elseif dm==F.DM_LISTDELETE then
      hDlg:ListDelete(2, {StartIndex=1,Count=1})
    elseif dm==F.DM_LISTINSERT then
      hDlg:ListInsert(2, {Index=3, Text="test3"})
    elseif dm==F.DM_LISTSET then
      hDlg:ListSet(2, {{Text="test3"}})
    elseif dm==F.DM_LISTSORT then
      hDlg:ListSort(2,0)
    elseif dm==F.DM_LISTUPDATE then
      hDlg:ListUpdate(2,{Index=2,Text="test3"})
    elseif dm==F.DM_LISTSETCURPOS then
      hDlg:ListSetCurPos(2, {SelectPos=2})
    end
  end
end)
```

Before the patch every DM_ message in the list caused unwanted changes of the text.
After: none of them.

The only msg where I am uncertain is DM_LISTSETCURPOS: may be it should still affect the edited text.


yegor-mialyk (approved):

far/dialog.cpp
```diff
@@ -5022,11 +5022,11 @@ intptr_t Dialog::SendMessage(intptr_t Msg,intptr_t Param1,void* Param2)
 					// уточнение для DI_COMBOBOX - здесь еще и DlgEdit нужно корректно заполнить
 					if (!CurItem.IFlags.Check(DLGIIF_COMBOBOXNOREDRAWEDIT) && Type==DI_COMBOBOX && CurItem.ObjPtr)
 					{
-						if (ListBox->HasVisible())
+						if (ListBox->HasVisible() && (CurItem.Flags & DIF_DROPDOWNLIST || Msg==DM_LISTSETCURPOS))
 						{
 							const auto& ListMenuItem = ListBox->at(ListBox->GetSelectPos());
 							const auto Edit = static_cast<DlgEdit*>(CurItem.ObjPtr);
-							if (CurItem.Flags & (DIF_DROPDOWNLIST|DIF_LISTNOAMPERSAND))
+							if (CurItem.Flags & DIF_LISTNOAMPERSAND)
```
alabuzhev, line 5029: Why this change? It seems to be unrelated to the issue.

alabuzhev: > That condition in initial form would be ever true

Maybe, I can't see it yet. Why do you think so?

> I suppose that was erroneous from the start.

It seems to be a [deliberate bugfix](https://github.com/FarGroup/FarManager/commit/1818aa02b3b72bbc9ac31befc7057b9ec5c2fe6c#diff-6e956c0db925004801c1509ab2884997c52d3bc9f5f44c9af810b204d6a08a55R5643-R5646).

johnd0e: I mean that this (2nd) condition is true

```cpp
if (CurItem.Flags & DIF_DROPDOWNLIST)
{
	if (CurItem.Flags & (DIF_DROPDOWNLIST|DIF_LISTNOAMPERSAND))
	{
```

And in it's initial form it also looked suspicious to me:
https://github.com/FarGroup/FarManager/blob/5f71f81412163821955049811aba544be5cbdcae/far/dialog.cpp#L5029-L5032

In case of DIF_DROPDOWNLIST we use SetHiString. Is that really for purpose?

alabuzhev: > I mean that this (2nd) condition is true

In the code the first condition looks different and we can enter the block even when DIF_DROPDOWNLIST is not set.

> In case of DIF_DROPDOWNLIST we use SetHiString. Is that really for purpose?

Looking at the changelog for builds 431 and 436, the idea apparently was to filter out extra '&' when hotkeys are enabled (DIF_LISTNOAMPERSAND is set).
However, it also happens for *any* dropdown, even with hotkeys disabled, which doesn't make much sense. Perhaps you're right and it's a bug.

There are 3 more similar occurrences:
image

Could you please remove DIF_DROPDOWNLIST from all of them?
Ideally in 2 commits: one with changing the first condition, the other  with ~~DIF_DROPDOWNLIST~~.

johnd0e: Done

far/dialog.cpp
```diff
@@ -936,7 +936,7 @@ void Dialog::InitDialogObjects(size_t ID, bool ReInit)
 				{
 					const auto Text = NullToEmpty(ItemIterator->Text);
 
-					if (Item.Flags & (DIF_DROPDOWNLIST | DIF_LISTNOAMPERSAND))
+					if (Item.Flags & DIF_DROPDOWNLIST)
```
alabuzhev, line 939: Wrong flag removed?

johnd0e: Oops. My bad.

alabuzhev (approved):
Looks good.
Please prepend the commit message and the changelog entry with "gh-951" and we can merge.

scriptforfivem:
The performance optimizations here are quite clever. Nice engineering work!

alabuzhev:
Thanks

Issue #948, opened 2025-05-07
Build 6473 (fixed columns in Find All menu) broke format of menu item


MKadaner:
### Far Manager version

3.0.6473.0

### OS version

N/A

### Other software

See [Ошибки при получении текста пункта МЕНЮ в результате "Искать все"](https://forum.farmanager.com/viewtopic.php?t=13636) on Far Manager Forum.

### Steps to reproduce

See script from [Синхронизация редактора с результатами поиска (Show All)](https://forum.farmanager.com/viewtopic.php?t=11977).

### Expected behavior

Line and column number can be parsed from the retrieved menu item string using the regex.

### Actual behavior

Parsing is broken.

MKadaner:
@alabuzhev, please assign this issue to me.

MKadaner:
Fixed in 6548.

PR #952, opened 2025-05-09
gh-948: Fixed format of Find All menu item

MKadaner:
## Summary
The format was modified in 6473 and this change broke some macros, see [Ошибки при получении текста пункта МЕНЮ в результате "Искать все"](https://forum.farmanager.com/viewtopic.php?t=13636).

Starting from this version, the item can be parsed with Lua regex `"^%s*(%d+)%s*%D%s*(%d+)%D"`, which also works for the versions before 6473.

## References
- gh-948

## Checklist


alabuzhev:
image

Why `  4│10│         <setting...` on the screen, but `4:    10│        <setting...` in GetValue?
It should either be the same as shown (if we consider columns to be part of the item) or just `        <setting...` (if we ignore them).
"*Suddenly a wild colon appears*" can be confusing, especially to new users, not burdened with historic baggage.

> which also works for the versions before 6473

Why do we care? All menus and dialogs are subject to change without any notice, we never promised any compatibility here.

MKadaner:
> All menus and dialogs are subject to change without any notice, we never promised any compatibility here.

I like this statement. And it looks like the author of the affected macro managed to solve the problem. So, I will drop this PR.

> It should either be the same as shown (if we consider columns to be part of the item) or just `        <setting...` (if we ignore them).

With the existing implementation, it is neither. What I did is somewhat of a hack. I do not exactly like it, but I consider it a step in the right direction, which is the latter of your alternatives. I have an idea, which I am going to implement in due course.

The contents of the fixed columns should be obtained from a callback to the menu owner (with trivial default for all menus but Find All). Then the row/column is not part of the item anymore. (We will need to expose them to Lua directly, otherwise the unfortunate macro is broken for good.)

The next step after this one is to avoid copying the Editor line into menu item. The item text also can be obtained from the owner on the fly.

In fact, the last two nights I tried to decouple `Item.Name` from the string returned to the API, and generate the latter on demand, but it proved to be way too involved.

PR #966, opened 2025-06-07
bugfix: fix handling of special symbols from user input dialog

michaellukashov:
Special symbols from user input dialog not handled properly.

how to reproduce:
  create menu item with user input dialog:
git !?$GitCommand$Command:?! !?$GitCommandAddOptions$Parameters? ! !?$GitCommandAddOptions2$&2 Additional options?!

call this menu item, enter input string with special symbols, for example: checkout !&

resulting string is: git checkout !&

This patch fixes it.


alabuzhev:
The current behavior is intentional, for obvious reasons.
If we start expanding special symbols in _user inputs_, the users will have to manually escape every `!` there, or funny things will happen.

michaellukashov:
> The current behavior is intentional, for obvious reasons. If we start expanding special symbols in _user inputs_, the users will have to manually escape every `!` there, or funny things will happen.

No. With this patch, `diff !&` expands to e.g. `git diff "panel.cpp" "infolist.cpp"`
Without this patch `diff !&` is executed as is:
```
git.cmd diff !&
fatal: ambiguous argument '!': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

alabuzhev:
> With this patch, diff !& expands

Exactly, but not _only_ `!&`.

Guess what will happen when you paste names like `Wham! - Last Christmas.mp3` or `S01E04 - M. Night Shaym-Aliens!.mkv` or `Readme!.txt` into the input box.

michaellukashov:
> Guess what will happen when you paste names like `Wham! - Last Christmas.mp3` or `S01E04 - M. Night Shaym-Aliens!.mkv` or `Readme!.txt` into the input box.

ok. created file Readme!.txt, selected it, called menu item: `git diff !&`
```
git diff "Readme!.txt" "panel.cpp" "infolist.cpp"
```
It works as intended, nothing wrong is happened.

alabuzhev:
Good to know, but it's not what I asked.

alabuzhev:
Just in case if it's still not obvious what is wrong with this approach:
1. Extract [gh-966.zip](https://github.com/user-attachments/files/22567107/gh-966.zip) somewhere.
2. Put the cursor on `_readme!.txt`
3. F2, Enter, Enter

I'm not against expanding special symbols in user inputs in principle, but we cannot break the processing of random files that just happen to have `!` somewhere in the name.
Yes, perhaps picking `!` for `Long file name without extension` was a horrible design mistake in the first place, but that ship has sailed.
It has to be either some new, non-conflicting syntax or opt-in, e.g. a checkbox "Expand special symbols" in the dialog or something.

PR #980, opened 2025-07-12
minor fixups for history

vyache76:
## Summary
avoid adding to history if  `cmd' starts with space

 
## Checklist

### DIFF
--- a/far/cmdline.cpp
+++ b/far/cmdline.cpp
@@ -490,7 +490,7 @@ bool CommandLine::ProcessKey(const Manager::Key& Key)
 			{
 				ActivePanel->SetCurPath();
 
-				if (!(Global->Opt->ExcludeCmdHistory&EXCLUDECMDHISTORY_NOTCMDLINE))
+				if (strStr.at(0) != TEXT(' ') && !(Global->Opt->ExcludeCmdHistory&EXCLUDECMDHISTORY_NOTCMDLINE))
 					Global->CtrlObject->CmdHistory->AddToHistory(strStr, HR_DEFAULT, nullptr, {}, m_CurDir);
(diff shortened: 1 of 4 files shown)


vyache76:
Should I close this request or what...?

yegor-mialyk:
It is likely one of the following reasons:

- Nobody needs this
- This can be achieved using a macro

vyache76:
**ok**, [but inserting `space` is easier than disabling history or if you forgot then use `vacuum`], anyway 'histview' needs some (cosmetic) fixes, most things have already been fixed (less than six years have passed)

yegor-mialyk:
Here is a simple straightforward macro that does the job.

```lua
Macro {
  description = "Do not add a command to history if it starts with a space";
  area = "Shell";
  key = "Enter";
  flags = "NotEmptyCommandLine";
  
  condition = function()
    local cmd = panel.GetCmdLine()
    return cmd:len() >= 2 and cmd:sub(1, 1) == " ";
  end;

  action = function()
    Keys("Enter AltF8 ShiftDel Esc")
  end;
}
```

There are many advanced Lua writers here who can come up with a better solution. This macro does not disable history and it just takes four simple keystrokes.

vyache76:
and what about Lif-separator? and `hitory` when usr start with `space`?
((off top))

yegor-mialyk:
I do not think there is any problem with LIF_SEPARATORs, a date separator in the History dialog will be automatically removed by Far.

But maybe I just did not get your question.

vyache76:
it's just logical - what's in history is on the screen (without the past :) )
They will be removed but after redrawing... and what about history when we start with `[[space]]`
(кроме того звать плагин при каждом нажатии `клавиши` ..... Зачем?, would be good to add this to the `settings` of course..)

Issue #973, opened 2025-07-04
New Editor with All Found lines: filename extension should be configurable


MKadaner:
### This is follow-up on #955. Copying here relevant discussion from the PR

_By @nightroman in https://github.com/FarGroup/FarManager/issues/955#issuecomment-3021420441_

Hello All, thank you for the promising feature.

A flaw? As far as I understand this pattern is used for new file names with results: `{original-base-name}.Filtered.{original-extension}`. With `FarColorer` plugin, in some cases this leads to lots of Colorer errors in result files. E.g. found pieces of original XML file `Test.xml` are no longer valid XML in `Test.Filtered.xml` (no closing tags, etc.).

Maybe it would be better if the result file is named like `*.txt` (neutral txt Colorer scheme) or even `{original-base-name}.{original-extension}.Filtered` hoping that this does not find any Colorer scheme for `.Filtered`. // IMHO, `*.txt` way is better.

NB Technically, it's not even about `FarColorer`. Other editor tools bound to file extensions may be confused, too.

_By @MKadaner in https://github.com/FarGroup/FarManager/issues/955#issuecomment-3024087559_

Ouch! When I implemented this behavior, I had `.log` files in mind, but of course, all structured file formats are going to be damaged. This is totally unacceptable. Corrupted `.xml` files look awful with Colorer.

Appending `.Filtered.txt` to the original file name is a simple and safe solution. The only disadvantage, at least from my perspective, is that the original extension is lost even for non-structured files, which means coloring may change dramatically between the original and the filtered file.

Unless somebody (@HamRusTal?) suggests a more user-friendly, even if more complex to implement, solution, I'll change the behavior to `.Filtered.txt`. Will try to do it during the coming weekend.

_By @HamRusTal in https://github.com/FarGroup/FarManager/issues/955#issuecomment-3024670401_

.., I suggest that it is the entire resulting file name template that is put to the language file, with the original base name (as `{1}`) and original extension (as `{2}`) separated, like this: `{1}.{2}.Filtered.txt`. This way, those few users who are unhappy with the default will be able to customize (via `FarEng.lng.custom`) the template to their liking, e.g. `{1}_grep.{2}`.

**P.S.** An example of how these `{1}` and `{2}` should be substituted can be seen e.g. with the `MListFileSize` and `MListFileSizeStatus` localized strings.

_By @MKadaner in https://github.com/FarGroup/FarManager/issues/955#issuecomment-3025386993_

[...]

PR #981, opened 2025-07-13
gh-973: Smarter filename for the Find All Editor.

MKadaner:
## Summary
Two UI format strings and a `far:config` to choose the right one.

## References
- #973

## Checklist
- [x] I have discussed this with project maintainers: #973<br/>


far/farlang.templ.m4
```diff
@@ -7681,20 +7681,35 @@ upd:"Occurrences: {0}, lines: {1}"
...
-upd:".Filtered"
-".Odfiltrowane"
-upd:".Filtered"
-upd:".Filtered"
-upd:".Filtered"
-upd:".Filtered"
-upd:".Filtered"
-upd:".Filtered"
-upd:".Filtered"
+MEditSearchAllFileNameFormat
```
alabuzhev, line 7684: Do we really need this in lng?
Maybe keep the "Filtered" suffix as is and form the names in the code?

MKadaner:
Well, this approach was suggested by @HamRusTal. I think it's a good idea. It is more flexible than hard-coded formatting, and what's wrong with it?

I am not really strong on this solution, so please let me know.

far/farlang.templ.m4
MKadaner: This approach was suggested by @HamRusTal. I think it's a good idea. It is more flexible than hard-coded formatting, and I do not see anything wrong with it.

That said, I am not really strong on this solution, so please let me know.

HamRusTal: Here's the [original suggestion](https://github.com/FarGroup/FarManager/pull/955#issuecomment-3024670401):

> I did not look into the source code, but since the current approach involves the English word `Filtered` I assume the implementation uses FAR's l10n mechanism with `.lng` files already (and if not, it definitely should). With that in mind, I suggest that it is the entire resulting file name template that is put to the language file, with the original base name (as `{1}`) and original extension (as `{2}`) separated, like this: `{1}.{2}.Filtered.txt`. This way, those few users who are unhappy with the default will be able to customize (via `FarEng.lng.custom`) the template to their liking, e.g. `{1}_grep.{2}`.
> 
> **P.S.** An example of how these `{1}` and `{2}` should be substituted can be seen e.g. with the `MListFileSize` and `MListFileSizeStatus` localized strings.

MKadaner: And this is exactly what I did, plus a new `far:config` to choose between two different `.lng` format strings from the follow-up discussion.

nightroman: Just IMHO. This `Filtered` case and some others like `New_File...` file patterns should not be localized, perhaps. Known file patterns may be used by various tools / plugins for deciding how to process a file. Localization makes this more difficult.

HamRusTal: IMHO forcing all users who prefer texts in their native language to the hard-coded English-language names for the sake of hypothetic tools that these users will likely never need is too unfriendly. Not to say about inflexibility of being unable to change the file name format (see my [earlier comment](https://github.com/FarGroup/FarManager/pull/981#discussion_r2203534742)).

alabuzhev: I tend to (partially) agree with @nightroman.
`New_File...` is not really special, it's just a new file without any meaning attached to it.
But this `Filtered` thing is semantically different from other files, so it could be useful to make it recognizable by plugins. E.g. maybe someone will come up with a colorer schema to add zebra background to it, or highlight the search string etc.

Also, it's a _temporary_ file. Who cares? The name is only there because some name has to be there.
When you run `edit:<git diff`, does seeing "FCQ9A7.tmp" in the title also feel forced?

MKadaner: > Also, it's a _temporary_ file. Who cares?

If Colorer depends (at least partially) on file extension for choosing its schema, I do care. When I searched a `.log` file, I really want to see the filtered result in the same color schema as the original file. However, with `.xml` it's a totally different matter, Then, If I decide to save the filtered file, I usually want it in the same directory with the original file, sorted next to it. Of course, the `Filtered` part is more or less irrelevant.

> ... it could be useful to make it recognizable by plugins ...

Can plugins access Far localized strings? Just asking. Also, if such a plugin is ever written, with customizable format, one can always redefine the format string to be equal to the predefined (possibly English) version. However, with a hard-coded format, such plugins / macros will have to hard-code this string as well, and the next thing will be, "why did you break my beloved macro by changing this filthy "Filtered" to shiny "Grepped?"

That said, I can live with any solution if it will preserve extension of certain files. If all filtered files will become `.txt` (or `.tmp`, or whatever)... well, I can probably live with that too.

Please let me know.

HamRusTal: > Also, it's a _temporary_ file. Who cares? The name is only there because some name has to be there.
> When you run `edit:<git diff`, does seeing "FCQ9A7.tmp" in the title also feel forced?

The example with `git diff` is not representative enough because e.g. Colorer (fortunately) recognizes the “diff/patch” format by content (rather than by file name) and provides highlighting (and outliner) based on that. It's not always the case, file name (not only its extension) often matters for plugins (like Colorer, EUM, [ESC], etc) to provide their services based on file name pattern.

So, as long as I don't really care about the .tmp file name in the window title (which in this case just tells me that the file is temporary), I'd much prefer if plugins (like FarCmds) could specify more meaningful and integration-friendly file names for temporary files.

Back to the functionality at hand, Colorer-friendly file name is the primary requirement for me (with the “LOG vs XML” distinction made of course). Personally, I don't care too much about l10n of these patterns as I always use FAR in English anyway. But I can easily imagine people (e.g. from France, Iceland, Turkey, or China to name a few) who will perhaps dislike being unable to translate this particular piece of the otherwise well localizable UI.

And, as I mentioned earlier, the added benefit is that having the file name format in the .lng file is a convenient means of customizing it to one's liking.

johnd0e: > If Colorer depends (at least partially) on file extension for choosing its schema, I do care.

You could add new filename patterns to Colorer, even if it will be "name.ext.Filtered".

HamRusTal: > You could add new filename patterns to Colorer, even if it will be "name.ext.Filtered".

It's true, but why make every user of the feature perform the same additional tweaking rather than design things so that they work automagically?

johnd0e: > they work automagically?

I see it differently. Result file almost never can be syntax-valid with original Colorer schema.

HamRusTal: > Result file almost never can be syntax-valid with original Colorer schema.

If the format is line-oriented (like TXT, LOG, CSV, INI, ASM, …) and the filtered file preserves whole lines (which I believe is true) then the original Colorer schema will still apply. Whether it will be “almost never” or “almost always” depends exclusively on the user's individual distribution of file types they handle. E.g. @MKadaner uses this mostly for logs if I understand him correctly.

MKadaner: @HamRusTal, thank you for having spoken on my behalf while I could not get here being busy staring at LOG files.

Yes, LOGs and TXTs. ASM yesterday. XMLs and JSONs occasionally. It will be very vexing to me if Colorer picks a different schema after I filtered a LOG.

Anyway, last attempt to merge it before I give up.

MKadaner: So, basically, we are discussing three options:

1. No customization. Append hardcoded `.Filtered.txt` to the end of the file name.
2. Keep `.Filtered.txt` in the LNG file and append it to the end of the file name.
3. Current proposal: two customizable filename formats; select based on `far:config` list of extensions.

Would it suite everyone here if in the current proposal I make the list of special extensions empty by default? The behavior out of the box will be similar to bullet 2, except for having more customization points to those who wants them (myself)?

alabuzhev: @MKadaner I don't have a strong opinion on this.
From a maintainer's perspective the less code and knobs the better.

MKadaner: @alabuzhev, I can live with either of the four variants. However, I strongly prefer the current proposal.

Just curious, from maintainer's perspective, how much expensive extra 10 lines of code and a config value are?

far/editor.cpp
```diff
@@ -4047,16 +4047,26 @@ void Editor::SaveFoundItemsToNewEditor(const VMenu& ListBox, const bool Matching
...
 {
-	auto SearchAllFileName{ msg(lng::MEditSearchAllFileNameAppend) };
+	string MyFilename;
+	string_view Stem;
+	string_view Ext;
+	auto MsgId{ lng::MEditSearchAllFileNameFormat };
 
-	const auto HostFileEditor = GetHostFileEditor();
-	if (!HostFileEditor) return SearchAllFileName;
+	if (const auto MyFileEditor = GetHostFileEditor())
```
alabuzhev: Please avoid MyEditor, MyClass etc.
There's no reason to not give it a proper name.

far/FarCze.hlf.m4
```diff
@@ -6622,6 +6622,25 @@ by the Editor.
...
+
+ The new filename is created from the stem filename and the extension
+of the file currently opened in the Editor. Depending on the original
+file extension, the new filename is formatted using one of the two
+alternative UI format strings defined in ~.lng files~@CustomizingUI@.
+The string with the ID #MEditSearchAllFileNameFormat# is used if the
+original file extension is not found in the value of this parameter;
+otherwise, #MEditSearchAllFileNameFormatAlt# is used.
+
+ Default value: #.txt;.log;.md;.csv;.ini;.cmd;.map#
```
alabuzhev: Do we really need dots here?

far/editor.cpp
```diff
@@ -4047,16 +4047,26 @@ void Editor::SaveFoundItemsToNewEditor(const VMenu& ListBox, const bool Matching
...
+		const auto [MyStem, MyExt]{ name_ext(MyFilename) };
+
+		Stem = MyStem;
+		if (MyExt.size() > 1) // More than empty or "."
+			Ext = MyExt;
+	}
 
-	string HostType, HostName;
-	HostFileEditor->GetTypeAndName(HostType, HostName);
+	if (!Ext.empty() && contains(EdOpt.SearchAllUseAltFileNameFormat.Get(), Ext))
```
alabuzhev: 1. EdOpt.SearchAllUseAltFileNameFormat = `".txt;.csv"`
2. Ext = `".cs"`
3. ?

far/FarCze.hlf.m4
MKadaner: Not absolutely. I think it is more natural for the user, since we are talking file extensions here, and it resembles the format of `PATHEXT` environment variable.

I am not strong on it. Please let me know.

far/editor.cpp
MKadaner: Sure. Done.

far/editor.cpp
MKadaner: Right! Let me fix it. Thanks.

alabuzhev: You can use `enum_tokens` here.

MKadaner: Thank you for the hint. I am experimenting with ranges. If it works and not obscenely expensive, I'll go with it, if you do not mind. Likely later today my time (I have to step away right now).

far/FarCze.hlf.m4
alabuzhev: pathext is ancient legacy and the user is Far user, not DOS user, so technically "more natural" would've been having a proper list of file masks here, like in highlight for example, with all its luxury like groups, regexes etc.

far/editor.cpp
alabuzhev: If we use [proper masks here](https://github.com/FarGroup/FarManager/pull/981/files#r2217456834), it should be as simple as
```C++
filemasks Mask;
const auto MsgId = Mask.assign(EdOpt.SearchAllUseAltFileNameFormat) && Mask.check(OriginalFilename)?
    lng::MEditSearchAllFileNameFormatAlt :
    lng::MEditSearchAllFileNameFormat;
```

MKadaner: Cool! Thank you. This is how it should be.

Done.

far/FarCze.hlf.m4
MKadaner: Fixed it all. Thank you so much!

alabuzhev:
Thank you

MKadaner:
Thank you for your help!

far/farlang.templ.m4
alabuzhev: > how much expensive extra 10 lines of code and a config value are?

A quote from someone:
> Every configuration option in a program is a place where the program is too stupid to figure out for itself what the user really wants, and should be considered a failure of both the program and the programmer who implemented it.

It applies in this case too.
What we ultimately wanted here is to higlight the filtered data *properly*.
And "properly" here is defined by the source data.
It can be structured across multiple lines, e.g. it's a source file.
It can be structured within one line, e.g. it's a log or a csv file.
It can be unstructured, e.g. it's a piece of literature.

The highlighter (e.g. colorer) already has this information, so a proper solution would've been communicating to it *somehow* that the new editor has filtered data of a certain type.
Maybe via some new API, or maybe even via file names like now, but by establishing and setting in stone a protocol like _"`name.filtered.ext` means filtered data of type `ext`"_, to let colorer figure out _automatically_ how to highlight it, i.e. whether to show/validate the structure or ignore it to certain degree or entirely. Yes, it would've required more work and changes in colorer, but it would've worked automatically, always and for everyone.
Instead we came up with a config option to either keep the original type (same extension) for unstructured or line-structured data, or fallback to txt for multiline-structured data. It's faster, but ultimately it means that the program failed to figure it out for itself, even though it had all the required information, and dropped the burden on the user.

On top of that, every config option doubles (or triples, or more) the number of possible paths and the ways things could go wrong. I can't count how many times there was a bug report I couldn't reproduce because the reporter's configuration was different from mine.
I'm not saying we can't have them at all, or shouldn't add them anymore, or that I haven't done that myself, just something to keep in mind.

MKadaner: Thank you for the thorough explanation, @alabuzhev. I agree that coordinating with Colorer (via a new protocol) would be the right solution here. I considered it momentarily but dropped the idea because implementing such end-to-end solution would be impossible for me given the amount of time I can devote to this hobby of mine. So, I pushed for the poor man's substitute.

> I can't count how many times there was a bug report I couldn't reproduce because the reporter's configuration was different from mine.

Fair enough. What do you think of routinely requiring accompanying bug reports with the exported Far configuration (maybe provide a tool to redact PII)?

> it would've worked automatically, always and for everyone.

Except for when it would not. And when it happens, it would be a maintenance nightmare. It so happened that for many years now I work in organizations using so called DEVOPS model (not that I like it). Thus, periodically, instead of writing new bugs, I am investigating issues reported by the customers (this is where my obsession with logs comes from.) In my experience, the toughest issues are those which happen because two components failed to coordinate properly. Hot potato game aside, everything works as expected on both sides, but the system behavior observed by the customer is... not what they expect.

> A quote from someone:
> 
> > Every configuration option in a program is a place where the program is too stupid to figure out for itself what the user really wants, and should be considered a failure of both the program and the programmer who implemented it.

Great maxim. And I like Far (and few other programs) exactly because in many cases it does the right thing without me even thinking of it. Two things to notice:

1. The poor man's solution I pushed through is actually not so bad from this perspective. With the right configuration (we may have to tweak it a bit based on customers' feedback), it will do the right thing out of the box, and nobody will ever remember about the config and customizable formats which work behind the scenes.
2. Trying too hard to live up to this maxim, we run the risk of ending up with AI in its contemporary sense. ~The program which is able to make a correct and ultimate guess about what the user really wants should just tell the user, "Go dance with your girlfriend."~ Kidding aside, it's a story of my life. My scenario is never a target scenario... and there is never this petty configuration option that would make me a happy user.

alabuzhev: > implementing such end-to-end solution would be impossible for me given the amount of time I can devote to this hobby of mine

I realise that and that's exactly why it's merged already.
Thank you for your contributions and your patience.

> What do you think of routinely requiring accompanying bug reports with the exported Far configuration (maybe provide a tool to redact PII)

I think we can even export and redact it automatically.

> we run the risk of ending up with AI in its contemporary sense

AI in its contemporary sense at least considers your remarks and can be persuaded to do your biddings, so one can say it has an infinite number of knobs (albeit unlabeled and shuffled at random). I suspect eventually that might change though.
Still, I do use some software for years without opening its settings even once, so that probably means something.

MKadaner: > I think we can even export and redact it automatically.

On a crash, yes. But how about less severe issues? After all, crash dump is pretty much self-contained. At least all settings should be cached in memory. So, unless you have a trick up your sleeve, it should be a script / macro / command to automatically collect artifacts in support of the report. Let's get thing rolling. What needs to be collected?

- List of active plugins, maybe with some settings.
- Installed macros.
- Sysinfo.
- All Far settings.
- Relevant system settings, like conhost vs. WT, current code page...
- Anything else?

Should it be an external script / executable or something which can be activated from within Far?

> AI in its contemporary sense at least considers your remarks and can be persuaded to do your biddings

I was told (never checked) that, for example, there is no way to make Grok AI generate a picture of the Battle of Hastings without inserting lots of Black and some Asian characters. To the best of my professional, though limited understanding, it will be so until AI instances won't have to undergo mandatory education phase where they are forced (with a thumb-sized stick if necessary) to ingest the known (not confabulated) human history, science, and culture and unquestioningly use this knowledge to ground their utterances. The same pedagogical device may be helpful in educating some human beings. Eventually that might happen.

alabuzhev: > But how about less severe issues?

On crashes we produce a human-readable report along with the dump (use CtrlAltApps to check it out).
In theory it should be possible to reuse most of it to generate similar reports on demand.

PR #984, opened 2025-08-05
gh-948: Lua API to retrieve extended user data from menu items.

MKadaner:
## Summary
Added Lua API `T=Menu.GetItemExtendedData([N])` and a new macro `Edit.SyncEditorAndFindAllMenu.lua` demonstrating usage of this API.

 
## References
- gh-948

## Checklist
- [X] I have discussed this with project maintainers [on the forum](https://forum.farmanager.com/viewtopic.php?p=180360#p180360)<br/>

## Details
The second attempt to provide proper support for the script [Синхронизация редактора с результатами поиска (Show All)](https://forum.farmanager.com/viewtopic.php?t=11977). With this API, the script should become substantially simpler.

When this script is updated to take advantage of this API, I will remove the fixed columns with the line and the position of the found pattern from the Find All menu items. We briefly discussed this idea [on the forum](https://forum.farmanager.com/viewtopic.php?p=180414#p180414).

This PR is ready for review.


MKadaner:
Hi @shmuz, could you please help me here?

I added a new Lua function `Menu.UserData(N)` which retrieves a typed element of menu item's complex user data. The difference from the existing `DM_LISTGETDATA(SIZE)` plugin API is that the call reaches directly the owner of the complex user data (in this case `Editor`) which actually knows what the data is and can return a meaningful data element instead of an opaque byte buffer.

I am experimenting with this new function in the following Lua script (adopted from [Синхронизация редактора с результатами поиска](https://forum.farmanager.com/viewtopic.php?t=11977).

<details>

<summary>EditorFindAllList.lua</summary>

```Lua

local HighlightColor = 0x5f
local FindAllListSyncId = "8564FC78-88C0-4044-9EB2-2FD345B52936"

local F = far.Flags
local EditorFindAllListId = win.Uuid(far.Guids.EditorFindAllListId)

local EditorInfo = nil
local HighlightedText = nil
local FirstDraw = true

local function DelColor()
  if HighlightedText then
    editor.DelColor(EditorInfo.EditorID, HighlightedText.Line, HighlightedText.Pos, FindAllListSyncId)
    HighlightedText = nil
  end
end

local function PositionEditor()
  editor.SetPosition(EditorInfo.EditorID,
                     HighlightedText.Line,
                     HighlightedText.Pos,
                     nil, -- CurTabPos
                     HighlightedText.Line - 1,
                     0) -- LeftPos
end

local function HighlighText()
  DelColor()
  HighlightedText = { Line = Menu.UserData(0), Pos = Menu.UserData(1), Len = Menu.UserData(2) }

--  far.Message(HighlightedText.Line, "HighlighText") -- !!!!

  editor.AddColor(EditorInfo.EditorID,
                  HighlightedText.Line,
                  HighlightedText.Pos,
                  HighlightedText.Pos + HighlightedText.Len - 1,
                  F.ECF_NONE,
                  HighlightColor,
                  nil, -- Priority
                  FindAllListSyncId)
  PositionEditor()
  editor.Redraw(EditorInfo.EditorID);
end

local function OnInitDialog()
  EditorInfo = editor.GetInfo(nil)

--  far.Message(EditorInfo.EditorID, "OnInitDialog") -- !!!!

  HighlighText()
end

local function OnDrawDialogDone()

--  far.Message(EditorInfo.EditorID, "OnDrawDialogDone") -- !!!!

  if Menu.UserData(0) > 0 and FirstDraw then
    FirstDraw = false
    HighlighText()
  end
end

local function OnListChange()
  HighlighText()
end

local function OnCloseDialog(FarDialogEvent)
  if (EditorInfo) then
    DelColor()
    if FarDialogEvent and FarDialogEvent.Param1 ~= 1 then
      editor.SetPosition(EditorInfo.EditorID, EditorInfo)
      editor.Redraw(EditorInfo.EditorID);
    end
  end
end

Event {
  description="Sychronize editor with the Find All menu";
  group="DialogEvent";
  condition = function(Event, FarDialogEvent)
    if Event ~= F.DE_DLGPROCEND then return false end
    local DialogInfo = FarDialogEvent.hDlg:send(F.DM_GETDIALOGINFO)

--    if FarDialogEvent.Msg == F.DN_DRAWDIALOGDONE then
--      far.Message(DialogInfo and DialogInfo.Id, "DialogEvent") -- !!!!
--    end

    return DialogInfo and DialogInfo.Id == EditorFindAllListId
  end;
  action = function(Event, FarDialogEvent)
    if     FarDialogEvent.Msg == F.DN_INITDIALOG     then OnInitDialog()
    elseif FarDialogEvent.Msg == F.DN_DRAWDIALOGDONE then OnDrawDialogDone()
    elseif FarDialogEvent.Msg == F.DN_LISTCHANGE     then OnListChange()
    elseif FarDialogEvent.Msg == F.DN_CLOSE          then OnCloseDialog()
    end
  end;
}
```

</details>

Everything works fine, except for when I call `Menu.UserData(N)` from the `DN_INITDIALOG` handler, the `VMenu2` dialog is not yet on the screen, so `MCODE_F_MENU_USERDATA` is dropped on the floor.

It looks like the right solution is to add a new `DM_LISTXXX` plugin API with the same functionality (given that, for example, `DM_LISTGETITEM` works in `DN_INITDIALOG` handler). However, I am not sure about the signature of this new API.

I want this API to be more or less generic and support data elements of different types. To Lua code, I could return an array of values of appropriate types, integers or strings (by the way, how to return a Lua array?) Plugin API seems to be more rigid, so the interpretation of the data rests with the calling plugin rather than supported by the type system. What is the right way to convey data type of the result of a Plugin API call? Is this question even meaningful?

I would appreciate if you draft the proper implementation or point me to an existing example.

Or maybe this problem should be solved in a completely different way?

shmuz:
@MKadaner please check your personal messages on the forum.

alabuzhev:
@MKadaner, could you please explain the concept a little more?
User data usually makes sense exclusively to its owner, to everyone else it is opaque by definition and is not something to touch or even observe. Peeking into it sounds suspicious.

MKadaner:
@alabuzhev, let me try to explain.

The solution drafted here does not allow peeking into "user data" associated with menu items. This data is indeed an implementation detail and should not be exposed. The central idea in my proposal is an optional callback deliberately registered by the owner of a menu. It should be registered only by such owners who can and want to provide their clients (scripts) with a contract going beyond the standard `VMenu` API.

Unfortunately, the function name used in this draft is a misnomer. As @shmuz and I discussed offline, it should be called `Menu.GetItemExtendedData` or something on these lines. It emphatically returns _not_ the user data stored by the owner along with the menu item. It is something which the owner is willing to provide to whoever may be interested. With this interpretation, the owner implicitly guarantees that (a) the data make some sense to the client in the context of this menu and (b) the contract is binding, i.e., the clients can depend on the types and semantics even if the internal implementation changes.

Please think of the Editor which created Find All menu. Incidentally, for now this is the only `VMenu` owner who uses any interesting information beyond menu item text. The Editor (and Editor only) registers the callback, thus promising to return, for a given menu item, the coordinates (in the file) of the found pattern featured in this menu item. It returns (to a Lua script) the line, position, and the length of the found pattern. These values have nothing to do with the `ComplexUserData` stored in the item. The Editor can change the way it internally stores the found coordinates (for example, it can use `Annotations` array to return position and length) and still maintain the external contract.

On the other hand, these values have very clear meaning in the context of the Find All menu. A script can use them, for example, to synchronize the Editor with the menu as the user switches between menu items. I am going to add such script (which is based on the existing one mentioned in the PR description) to Far installation. With the new API, the script will be reasonably simple and maintainable, and, as we know, it is actually used by some users. (I personally somewhat like the feature, though as usual I would prefer it being implemented natively).

Below is the draft description of this new Lua function. I hope it also helps understanding the concept.

> Эта функция возвращает дополнительные данные, связанные с указанным (или текущим) элементом меню. Количество и типы возвращаемых значений зависят от меню, для которого эта функция вызвана. Например, для меню всех вхождений образца поиска во встроенном редакторе возвращается три числа: номер строки в исходном файле, где найден образец, позиция образца в строке и длина найденного образца.

Did I manage to explain? I'll be happy to answer more questions if needed.

alabuzhev:
@MKadaner thanks, it makes more sense now.

> Unfortunately, the function name used in this draft is a misnomer

Maybe rename it then to extended information or something like that?

MKadaner:
> Maybe rename it then to extended information or something like that?

Doing exactly that. It's a draft. So far, I am getting the functionality right (and fighting some compilation issues.)

BTW, please look at the compiler errors in the latest build. The simplest way to fix them would be to `#include "vmenu.hpp"` into `vmenu2.hpp`. Technically, it will introduce an extra dependency. Practically, `vmenu.hpp` is already included everywhere where `vmenu2.hpp` is needed anyway.

What do you think? If you are OK with this, I'll effect it in a separate PR.

alabuzhev:
> compiler errors in the latest build

does RegisterExtendedDataProvider have to be a template (and thus in the header)?

MKadaner:
> does RegisterExtendedDataProvider have to be a template (and thus in the header)?

I am passing a lambda. Also, the callback type is defined in `vmenu.h`, but where I register it, I want to call a method of `VMenu2`, because it is handy. I can:

- Remove this method from `VMenu2` and register directly with `VMenu`.
- Move `extended_data_provider` definition to `VMenu2`.
- Suppress error with `if constexpr (requres ...)`
- Something else?

far/vmenu2.hpp
```diff
@@ -117,6 +117,12 @@ class VMenu2 final: public Dialog
 		return std::any_cast<T>(GetComplexUserData(Position));
 	}
 
+	template<class F, class M = VMenu>
```
MKadaner: See https://github.com/llvm/llvm-project/issues/115547. Is it related?
Also, https://godbolt.org/z/59ehn1jr9

MKadaner: This issue is moot. I removed the function from `VMenu2`.

MKadaner:
@shmuz, could you please look at the API implementation, the macro, and the related changes?

@alabuzhev, could you please review? It should make much more sense now. It is large. Technically, I could split it into API implementation and the macro, but I think that the API without the macro to test it is difficult to review.

far/editor.cpp
```diff
@@ -3471,6 +3471,16 @@ namespace
 				},
 				small_segment::ray(ItemTextStart)
 			);
+			m_Menu->ListBox().RegisterExtendedDataProvider([](const menu_item_ex& Item)
+				{
+					const auto Coord{ std::any_cast<FindCoord>(Item.ComplexUserData) };
+
+					return VMenu::extended_item_data{
+						{ L"Line", Coord.Line + 1 },
+						{ L"Position", Coord.Pos + 1 },
```
alabuzhev, line 3480: Why do we have `+ 1` here?
Is this about Lua using 1-based indexing / a general convention in legacy macros / both?

enc/enc_rus/meta/macro/macrocmd/prop_func/menus.html
```diff
@@ -207,6 +207,22 @@ <h2>Функции</h2>
        </table>
     </td>
   </tr>
+  <tr>
+    <td>T=<dfn><a name="menu.getitemextendeddata">menu.getitemextendeddata</a></dfn>([Pos])</td>
+    <td><p>Возвращает таблицу с дополнительными данными, связанными с указанным (<var>Pos</var>) или текущим (если <var>Pos</var> не указан) элементом меню.</p>
+      <p>Если текущий "объект" не меню или позиция <var>Pos</var> не валидна (меньше или равно 0 или больше количества пунктов в меню), то возвращается <var>nil</var>.</p>
```
alabuzhev: `равно` -> `равна` probably?
It's about position, which is feminine in this context.

far/editor.cpp
shmuz: > Is this about Lua using 1-based indexing / a general convention in legacy macros / both?

Both. And not only in legacy macros but also in LuaFAR API.

enc/enc_rus/meta/macro/macrocmd/prop_func/menus.html
johnd0e: > (`<= 0` или `>` количества пунктов в меню)

MKadaner: > `равно` -> `равна` probably?

Probably... I checked with a Russian grammar expert and improved both sentences according to the suggestion. Please see if you like it.

MKadaner:
Thank you.

enc/enc_rus/meta/macro/macrocmd/prop_func/menus.html
shmuz: @MKadaner 
The description of this function in the Encyclopedia is incomplete.
1. The hDlg parameter is missing
2. The function returns nil also in the case the menu does not have extended data

P.S.
I see you added the hDlg parameter description in [this](https://github.com/FarGroup/FarManager/pull/1024) pull request.
I would recommend to make a separate pull request related to this function's documentation.

MKadaner: @shmuz

I moved the description of `hDlg` parameter to a separate PR #1045.

Please have a look.

shmuz: @MKadaner 
Merged, thank you.

MKadaner: Thank you, @shmuz, appreciated.

Issue #995, opened 2025-09-01
Optionally hide files and folders starting with dot


mihnita:
### Description of the new feature or improvement

Because in the UNIX / Linux / macOS world the files and folders starting with a `'.'` are not shown, this convention "leaked" into Windows as well.

These files "get in the way" the same way hidden / system files do.

For example in my home folder I have 47 files and 75 folders.
Out of that, there are 12 files and 11 folders hidden, and 22 files / 41 folders starting with a dot.

So, if anything, the problem is worse.

It would be really nice to be able to easily hide / unhide these files / folders, the same way we can do with the ones with hidden / system attributes.

### Proposed technical implementation details (optional)

First: I would be happy to contribute the code for this.

The implementation is trivial, I don't think there are any technical challenges.
Just a bit of care to not miss anything.

alabuzhev:
Closed via #996

PR #996, opened 2025-09-01
Optionally hide files and folders starting with dot

mihnita:
## Summary

Because in the UNIX / Linux / macOS world the files and folders starting with a '.' are not shown, this convention "leaked" into Windows as well.

These files "get in the way" the same way hidden / system files do.

It would be really nice to be able to easily hide / unhide these files / folders, the same way we can do with the ones with hidden / system attributes.

## References

Issue #995

## Checklist
- [x] I have followed the [contributing guidelines]

I did not discuss it, but the change is small enough that if it is rejected I will not mind.
(I might still mind the lack of that feature though ;-)

## Details

The change is quite trivial.

In most places related to hiding/showing hidden and system files I added similar code to hide/show files starting with a dot.

## Note

I was not sure how to add the extra section in the help for other languages, so I only updated `FarEng.hlf.m4`


far/FarEng.hlf.m4
```diff
@@ -2504,6 +2505,9 @@ $ #Settings dialog: panel#
  #Show hidden and system files#
  Enables displaying files with Hidden or System attributes. This option can also be switched by #Ctrl+H#.
```
MKadaner: Please adjust this as well:

```suggestion
 Enables displaying files and folders with Hidden or System attributes. This option can also be switched by #Ctrl+H#.
```

far/FarEng.hlf.m4
```diff
@@ -2504,6 +2505,9 @@ $ #Settings dialog: panel#
...
 
+ #Show dot files#
+ Enables displaying files and folders that start with a dot. This option can also be switched by #Ctrl+D#.
+
```
MKadaner: Please copy the English section to all other `.hlf.m4` files. For Russian you can use:

```
 #Показывать файлы, начинающиеся с точки#
 Разрешает показ файлов и папок с именем, начинающимся с точки. Этот режим также может быть переключён с помощью #Ctrl+D#.

```

far/FarEng.hlf.m4
```diff
@@ -407,6 +407,7 @@ $ #Panel control commands#
  Set alternative full view mode                          #LeftCtrl+0#
 
  Toggle hidden and system files displaying                   #Ctrl+H#
+ Toggle dot files and folders displaying                     #Ctrl+D#
```
MKadaner: Please copy the English section to all other `.hlf.m4` files. For Russian you can use:

```
 Убрать/показать файлы с именем, начинающимся с точки        #Ctrl+D#
```

MKadaner (commented):
Thanks.

> I was not sure how to add the extra section in the help for other languages, so I only updated `FarEng.hlf.m4`

Simply copy-paste the English paragraphs to the corresponding places in other `.hlf.m4` files. Please make sure to keep the exact paragraph structure (number of paragraphs, indentations, empty lines after).

Disclaimer: I am not a maintainer.

alabuzhev:
Hi @mihnita,

Thanks for looking into this.

> this convention "leaked" into Windows as well

It did and it is so frustrating.
The whole idea of dotfiles is so fundamentally broken.

Just a few examples.
No one probably wants or needs to see the .git directory, because it's just some boring metadata, managed by the system.
But at the same time .gitignore and .gitattributes are very real files that are supposed to be in the index and properly managed by you to reflect your project needs.
And on top of that the bloody Github requires you to keep your workflows and issue & pull request templates in... the .github directory. 🤦 
Why "the stuff I don't care about" and "the stuff I have to care about" are handled in the same way?
This does not make any sense.
And in a typical home directory on a typical linux machine it's exactly the same situation, but on the scale of full blown insanity: manual configs, generated configs, histories, caches, some binary rubbish, whatever - a legion of them.

And the worst thing is that the file name is, obviously, part of the contract (unlike the file attributes), so you generally cannot unhide the stuff you do care about and keep the rubbish hidden (like in Windows).
This does not make any sense.
It's as if one approach was deliberately designed, and the other happened to evolve from a bug in ls. Oh wai...

 
But I digress.

Yes, we _probably_ should support it.
The whole concept is cursed, but, as you mentioned, it's leaking all over the place and with the advent of WSL it will only get worse with time.
Even the damn Visual Studio now creates .vs directories for some unholy reasons.
@microsoft, watch out, there's an impostor in VS team! 😄 

However, I do not think that we need to do it separately with all the glory like a config option, a hotkey, an API exposure etc.
If such files, or at least a significant portion of them, are conceptually hidden / system, albeit bastardized, **why don't we just bundle them with real hidden / system files?**
As you said, they get in the way the same way hidden / system files do.
Especially since quite a few apps ported to Windows use both the dot and the hidden attribute simultaneously, so you'll probably have to press Ctrl+D Ctrl+H together more often than not anyway just to make sure.

> I might still mind the lack of that feature though ;-)

Do you know that we have panel filters (Ctrl+I), which are basically Ctrl+H on steroids, and, technically, this particular feature (and much more, like actually separating the wheat from the tares and .git from .gitattributes) has been there all along?

> I was not sure how to add the extra section in the help for other languages

We have a check during the build that the basic structure and the number of paragraphs are the same, to stop them from diverging too much.
If you need to update the help, as @MKadaner mentioned, just copy the English blocks to all the languages you're not familiar with.

far/FarEng.hlf.m4
HamRusTal: Like it or not but `Ctrl+D` in panels is utilized in the command line (if not empty) already: it's equivalent to `Right`. So the suggested change will be breaking for some users (not me personally, I have a macro for `Ctrl+D` to perform another action).

HamRusTal: > Убрать/показать файлы

I suggest «Скрыть» instead of «Убрать» as that will be both more technical and less ambiguous.

yegor-mialyk:
I would personally prefer a setting (e.g. `Panel.TreatDotFilesAsHidden`) in `far:config`, and I’d like to keep using `Ctrl+H` as before.

far/FarEng.hlf.m4
mihnita: Done

far/FarEng.hlf.m4
mihnita: > Like it or not but Ctrl+D in panels is utilized in the command line (if not empty) already

ACK.
I will look for something else.
Or, if you all agree, I can go with the latest 2 suggestions: treating dot files as hidden, and only use one hotkey for both.

mihnita: > I suggest «Скрыть» instead of «Убрать» as that will be both more technical and less ambiguous.

That would be inconsistent

English
```
 Toggle hidden and system files displaying                   #Ctrl+H#
 Toggle dot files and folders displaying                     #Ctrl+D#
 Toggle long/short file names view mode                      #Ctrl+N#
```

Russian
```
 Убрать/показать файлы с атрибутом "Скрытый" и "Системный"   #Ctrl+H#
 Скрыть/показать файлы с именем, начинающимся с точки        #Ctrl+D#
 Переключить отображение длинных/коротких имён файлов        #Ctrl+N#
```

It is already inconsistent, with "Toggle" translated once as "Убрать/показать" and once as "Переключить"

I am happy to update all. Or just the new one.
Whatever you decide, I'll implement.
For now I went with "Убрать/показать" for (some) consistency.

mihnita:
> Do you know that we have panel filters (Ctrl+I), which are basically Ctrl+H on steroids

I know. And I used it for a while.
But because you still need to see files like .gitconfig, .gitignore, .github, and so on, it is a lot nicer to have a hotkey to easily turn things on / off.

---

> However, I do not think that we need to do it separately with all the glory like a config option, a hotkey, an API exposure etc.

> I would personally prefer a setting (e.g. Panel.TreatDotFilesAsHidden) in far:config, and I’d like to keep using Ctrl+H as before.

I'll do that.

In fact, that's how I started, adding the dotted files with hidden and system, with no option.
Just changed the test.
But I thought that maybe some people will want more control.

I think `Panel.TreatDotFilesAsHidden` is a good compromise. No extra hotkey, but there is some control.

mihnita:
Thank you all for the feedback.

I've changed the implementation to use `Panel.TreatDotFilesAsHidden`
I merged all the previous commits to hide all the back-and-forth, as this last change reverts several files (all the Plugin* stuff).

But since this required new wording I will need (again) Russian translations, sorry.

And because the previous string also changed, the Russian (and others :-( ) might also have to change.
Not critical (does not fundamentally change the meaning), but nice to update.

```diff
MConfigHidden
"Показывать скр&ытые и системные файлы"
"-Show &hidden and system files"
"+Show &hidden and system files and folders"
...
```
Should I add `upd:` in front of the other languages (for `MConfigHidden`)?

And is there an equivalent of `upd:` for the help files?

Thank you,
Mihai

yegor-mialyk:
> Do you know that we have panel filters (Ctrl+I)

Oh, I forgot about filters. Technically `Show/Hide hidden files` should have been migrated to filters + key macro :)

far/FarEng.hlf.m4
MKadaner: If you go with the new option, please use "Скрыть/показать" everywhere. There is no good translation for "Toggle" in Russian.

MKadaner:
If you add new `Panel.TreatDotFilesAsHidden` `far:Config`, please add help article (bunch of them at the bottom of the `.hlf.m4` file. When you have English version, I'll translate it to Russian. And you do not need any other language strings, right?

mihnita:
> If you add new `Panel.TreatDotFilesAsHidden` `far:Config`, please add help article (bunch of them at the bottom of the `.hlf.m4` file. When you have English version, I'll translate it to Russian.

Already done it.
Line 2507 in English and 2557 in Russian.
Also added the same small paragraph in all languages. 
```
 #Treat dot files and folders as hidden#
 When hiding/showing Hidden or System files or folders also hide/show the ones starting with a dot.
```

There is also one line in the UI, added it in English, and with `upd` for all other languages. 
`far/farlang.templ.m4` line 2623:
```
MConfigTreatDotFilesAsHidden
upd:"Treat d&ot files and folders as hidden"
"Treat d&ot files and folders as hidden"
...
```

I tested it, looks good, and functions correctly. 

Unless the proposal was just to add a config entry, with no UI at all, only accessible through `far:config`?

Mihai

far/FarEng.hlf.m4
mihnita: Like this?
```
 Скрыть/показать файлы с атрибутом "Скрытый" и "Системный"   #Ctrl+H#
 Скрыть/показать отображение длинных/коротких имён файлов    #Ctrl+N#
```

Technically that region does not change at all now.
But since you identified it as a problem, why not fix it

Note:
- the "new" option (with #Ctrl+D#) is gone
- I don't know Russian. I can read, and I can understand maybe 10%.
  But I can't build a sentence with proper grammar if my life depends on it.
  Don't assume I can fix gender / number / case agreements and so on :-)

alabuzhev:
> Unless the proposal was just to add a config entry, with no UI at all, only accessible through `far:config`?

Yes, just a config entry will do fine.

far/FarEng.hlf.m4
alabuzhev: Guys, please don't clutter the PR with translation peculiarities, especially since the author does not know the language and needs to wait for your input every time.
Please feel free to raise a following PR with whatever wording you prefer.

far/filelist.cpp
```diff
@@ -6882,6 +6882,17 @@ void FileList::UpdateIfRequired()
...
 
+static bool hideHiddenSystemAndDotFiles(DWORD Attributes, const string &FileName)
+{
+	if (Global->Opt->ShowHidden)
+		return false;
+	if (Attributes & (FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM))
+		return true;
+	if (!Global->Opt->TreatDotFilesAsHidden)
+		return false;
+	return !FileName.empty() && FileName.at(0) == '.';
```
alabuzhev: `FileName.starts_with(L'.')`

far/filelist.cpp
```diff
@@ -6882,6 +6882,17 @@ void FileList::UpdateIfRequired()
 	Update((m_KeepSelection? UPDATE_KEEP_SELECTION : 0) | UPDATE_IGNORE_VISIBLE);
 }
 
+static bool hideHiddenSystemAndDotFiles(DWORD Attributes, const string &FileName)
```
yegor-mialyk: Use `PascalCase` (or even `snake_case`) naming convention for methods. `Hide` sounds like an action, maybe something like `IsFileHiddenForDisplay` or `ShouldHideFilesFromView` etc

alabuzhev:
> But because you still need to see files like .gitconfig, .gitignore, .github, and so on, it is a lot nicer to have a hotkey to easily turn things on / off.

Do you know that we have macros and you can have a hotkey to basically anything, including switching certain filters on and off?

alabuzhev:
> Technically `Show/Hide hidden files` should have been migrated to filters + key macro :)

Be careful what you wish for, it might come true 😀

yegor-mialyk:
> I would personally prefer a setting (e.g. `Panel.TreatDotFilesAsHidden`) in `far:config`, and I’d like to keep using `Ctrl+H` as before.

Just to make it clear I meant "As a user I would prefer a setting..." above. However, my hidden files are always "tuned on" and `Ctrl+H` is remapped to a different action :)

far/farlang.templ.m4
```diff
@@ -2607,7 +2607,7 @@ l:
 
 MConfigHidden
 "Показывать скр&ытые и системные файлы"
```
MKadaner, line 2609:
```suggestion
"Показывать скр&ытые и системные файлы и папки"
```

MKadaner:
I removed a comment. Sorry for the mess. Please disregard any translation issues. Just copy-paste English version everywhere. I'll take care about Russian later. I thought it would be simple, but as always, one thing leads to another...

> > Unless the proposal was just to add a config entry, with no UI at all, only accessible through `far:config`?
> 
> Yes, just a config entry will do fine.

And if you do it, please add a help article similar to `@Panel.CtrlFRule` one.

far/filelist.cpp
mihnita: Done (I like `ShouldHideFilesFromView`, and went with it)

far/filelist.cpp
mihnita: Done

mihnita:
New commit, with all the changes requested.
So it should be ready for review.
Thank you all,
Mihai

mihnita:
> And if you do it, please add a help article similar to `@Panel.CtrlFRule` one.

Done.

far/FarEng.hlf.m4
```diff
@@ -6874,6 +6874,23 @@ deleted, or renamed.
...
+ This Boolean parameter controls the filtering of files or folders
+starting with a dot.
+ ^<wrap>It works in combination with the #Show hidden and system files and folders#
+option in the ~panel settings~@PanelSettings@ dialog.
+
+ False - ^<wrap>Dot files and folders will always be shown.
+ True  - Dot files and folders will be hidden / shown together
+with the Hidden and System files.
+
+ Default value: True (hide / show the dot files in sync with the
```
HamRusTal: It's a very minor thing but I'd suggest `hide/show` without spaces (in all language versions of course)

far/FarEng.hlf.m4
```diff
@@ -6874,6 +6874,23 @@ deleted, or renamed.
...
 
+@Panel.TreatDotFilesAsHidden
+$ #far:config Panel.TreatDotFilesAsHidden#
+ This Boolean parameter controls the filtering of files or folders
+starting with a dot.
+ ^<wrap>It works in combination with the #Show hidden and system files and folders#
+option in the ~panel settings~@PanelSettings@ dialog.
+
+ False - ^<wrap>Dot files and folders will always be shown.
+ True  - Dot files and folders will be hidden / shown together
```
HamRusTal: Shouldn't we enclose “Dot files” in (curly) quotes? This is not an official term after all, and it can even be confused with files with the `.dot` extension…
Also, it's a very minor thing but I'd suggest `hidden/shown` without spaces (in all language versions of course).

far/FarEng.hlf.m4
mihnita: Done

far/FarEng.hlf.m4
mihnita: > Also, it's a very minor thing but I'd suggest hidden/shown without spaces (in all language versions of course).

Done

---

> Shouldn't we enclose “Dot files” in (curly) quotes? This is not an official term after all, and it can even be confused with files with the .dot extension…

I tried, but it was weird, because the text was - Dot files and folders...
So technically I would have had to quote  not just
` False - "Dot files and folders" will be hidden/shown ...`
not just
` False - "Dot files" and folders will be hidden/shown ...`

So I did two things:
1. Changed the first paragraph to also kind of "define" what dotfiles are
- This Boolean parameter controls the filtering of files or folders
starting with a dot (“dotfiles”).

2. After searching online, there are many-many mentions of "dotfiles", without space
Many of the first hits also contain a definition of what those are.
So I change the term to "dotfiles" (without spaces), as that seems a lot more established.

WDYT?

---

Thank you,
M

far/farlang.templ.m4
```diff
@@ -1,4 +1,4 @@
-#
+﻿#
```
HamRusTal, line 1: Did something happen to the BOM here, did it dis-/appear? Won't it confuse the build pipeline tools?
Is any change in this file needed at all?

mihnita: I used Beyond Compare 5 at some point to selectively restore some things from a clean repo.
I did not realize BC removes the BOM.
Then I've seen a check (on cpp file) reporting the BOM as missing, and I put it back.
Checking, most files have a BOM (I did not check all).
But files that contain user visible text have it (all the help files, for example)

Since `farlang.temp.m4` contains all the UI text, in all languages, makes sense to also have a BOM.
Probably not technically needed? But safer, cleaner, and more consistent (at least in my opinion).

I can revert it if you want, of course.

M

far/config.cpp
```diff
@@ -1983,6 +1983,7 @@ void Options::InitConfigsData()
 		{FSSF_PRIVATE,           NKeyPanel,                  L"ShellRightLeftArrowsRule"sv,      ShellRightLeftArrowsRule, false},
 		{FSSF_PANEL,             NKeyPanel,                  L"ShowBytes"sv,                     ShowBytes, false},
 		{FSSF_PANEL,             NKeyPanel,                  L"ShowHidden"sv,                    ShowHidden, true},
+		{FSSF_PANEL,             NKeyPanel,                  L"TreatDotFilesAsHidden"sv,         TreatDotFilesAsHidden, true},
```
MKadaner: Is it a good idea to change the default, out of the box behavior? Even though I think I may use this feature in the future, as a Far user, I would be very unhappy if my `.gitignore` files disappear.

far/FarEng.hlf.m4
```diff
@@ -6874,6 +6874,22 @@ deleted, or renamed.
  This parameter can be changed via ~far:config~@FarConfig@ only.
 
 
+@Panel.TreatDotFilesAsHidden
+$ #far:config Panel.TreatDotFilesAsHidden#
+ This Boolean parameter controls the filtering of files or folders
+starting with a dot (“dotfiles”).
+ ^<wrap>It works in combination with the #Show hidden and system files and folders#
```
MKadaner: It is uncommon to use `^<wrap>` to break paragraphs. I am not even sure it works the way you intended, if I understand your intent correctly. Usually, `^<wrap>` indicates the position where the next line of a long (wrapped) paragraph should start, no matter why and how the paragraph happened. The paragraphs are usually separated with an empty line.

To clarify, do you what to have one or two paragraphs here? If two, please start each with space and leave an empty line between:

```
 This Boolean parameter controls the filtering of files or folders starting with a dot (“dotfiles”).

 It works in combination with the #Show hidden and system files and folders# option in the ~panel settings~@PanelSettings@ dialog.
```

If one, just break the lines in `.hlf.m4` as convenient, and do not add space at the beginning of continuation lines:

```
 This Boolean parameter controls the filtering of files or folders starting with a dot (“dotfiles”).
It works in combination with the #Show hidden and system files and folders# option in the ~panel settings~@PanelSettings@ dialog.
```

MKadaner: It is better to use exact phrase as it appears in the dialog (and in the help above): `#Show hidden and system files#`. The "files and folder" wording is more precise, and we should probably adopt it, but we should be consistent.

far/config.cpp
mihnita: > Is it a good idea to change the default, out of the box behavior? Even though I think I may use this feature in the future, as a Far user, I would be very unhappy if my .gitignore files disappear.

I hear you, and I thought about it.

What I am a bit worried about is that the feature will be undiscoverable if off by default, and only in `far:config`.

As they say
"If a feature is only in `far:config` and no one is around to see it, does it even exist?"
;-)

---

I don't think many people spend time in `far:config`.
I know I didn't.
I did it a bit now, and I discovered (for example) that I can change F8 in Editor/Viwer to include UTF-8 in the OEM/ANSI code page cycle.
Very nice, but undiscoverable.

And doing that is clunky.
I am in far:config, and is it a long list of flags, with no explanations.
I Shift+F1 on each, and some have explicit help, some take me to some generic help for a panel, and some do nothing.

---

How would you all feel if I add this back as a visible option in the Panel settings, under "Show hidden and system files"? (like in the second commit)
Off by default, but a bit more discoverable?

far/FarEng.hlf.m4
mihnita: I wanted one paragraph, but with a line break :-)
So not a `<p>`, but `<br>` (in html lingo).

I checked the current rendering with `^<wrap>`, and it does that.
But if that is not the intention of `^<wrap>`, and should not be used that way, I will remove it.

---

Sorry, at work, and I can't make changes now.
I'll do it tonight, at home.

mihnita: Ack, will do (tonight at home)

far/config.cpp
MKadaner: When I wear my developer's hat, I tend to agree with you. Moreover, because I am doing such-and-such feature for my own convenience, I also want to flip the default behavior, so that it is on for me out of the box. I am still not sure how I would take this behavior change as a user. Probably would be pissed off. And, since, as you aptly pointed out, the `far:config`-only option is undiscoverable, I would not even know how to fix the damn thing. 🙂

For the discoverability, as far as I have observed here, the things happen like this. Somebody comes with a question or suggestion, "I want to have this or that." And more often than not, somebody replies, "Please use this existing `far:config` (or another 'hidden') feature." Not very scalable but somehow works.

To be serious, un-discoverability of `far:config`-only options has bothered me for long time. Maybe we should add a directory of all articles covering such options to the Help? It probably can be composed automatically and linked from the main Help contents page.

@alabuzhev, what do you think? Yes, I am aware that this list can be seen on the `Help file index` page, as well as that the Help can be searched for `far:config` to get more or less same result. I would argue, however, that the discoverability of either of these ways is also less than great.

far/FarEng.hlf.m4
MKadaner: To me, a paragraph with a line break feels really strange, everywhere, not only here in the Help. With certain window width, we may see something like this:

```
 This Boolean parameter controls the filtering of files or folders starting with a dot
 (“dotfiles”).
 It works in combination with the #Show hidden and system files and folders#
 option in the panel settings dialog.
```

Looks weird. Did they mean one paragraph or two? Is it two but they forgot an empty line? Or is it one but they accidentally dropped LF in the middle?

This is my personal opinion.

mihnita: I don't feel strongly about this.
I'll make it one paragraph (tonight :-)

far/config.cpp
alabuzhev: It is indeed true that far:config is not exactly the first thing you notice, but that's the point.
Not every option has a UI.
Not every option should have a UI in the first place.
Some of them are rather obscure and of no interest to the general audience.
In fact, far:config is already immensely user-friendly, comparing to tech infos that predated it.

In this particular case I would strongly prefer _not_ having a visible option in Panel settings.
Simply because the whole idea of dotfiles is cursed (see my rant above) and we should not promote and advertise it.
I don't want an innocent soul to notice it in the UI, read about it and conclude that such a pattern is therefore a socially acceptable and widely supported way to hide files.

Regarding the default value, I don't have a strong opinion here.
We show hidden files by default, so it's irrelevant for out-of-the-box configurations.
And people who hide them likely don't have dotfiles anyway, or don't care.

far/farlang.templ.m4
HamRusTal: IMHO if it's UTF-8 and BOM does not break the build, we should leave it.

alabuzhev: Please keep the BOM.
Looks like I removed it in 71bbbcc55a8f26d72ae3a3b5f5611da44d9812ba by mistake.

far/config.cpp
mihnita: Done (changed to `false`)

far/FarEng.hlf.m4
mihnita: Done

mihnita: Done

far/aa
MKadaner, line 1: Oops!

mihnita (commented):
Implemented everything.
Defaulting to false, fixed help.

---

Let me know if you want me to squash the commits into one, for cleaner history.
Or you can squash?
I don't know the customs in this repo :-)

Thank you all,
Mihai

far/FarEng.hlf.m4
```diff
@@ -6874,6 +6874,22 @@ deleted, or renamed.
...
+ This Boolean parameter controls the filtering of files or folders
+starting with a dot (“dotfiles”). It works in combination with the
+#Show hidden and system files# option in the
+~panel settings~@PanelSettings@ dialog.
+
+ False - ^<wrap>Dotfiles will always be shown.
+ True  - Dotfiles will be hidden/shown together with the Hidden and System files.
+
+ Default value: True (hide/show the dotfiles in sync with the
+Hidden and System files).
```
MKadaner: Should be `False`?

mihnita: Darn :-)
Thanks, will do in a second!

far/aa
MKadaner: Thanks.

far/FarEng.hlf.m4
MKadaner: Thank you.

far/FarEng.hlf.m4
mihnita: Done

far/FarEng.hlf.m4
MKadaner: Thanks!

far/config.cpp
MKadaner: Thank you.

MKadaner:
> Let me know if you want me to squash the commits into one, for cleaner history.

The maintainers usually require squashing.

mihnita:
> The maintainers usually require squashing.

Done, all the feedback and squashed.

Sorry for all the "churn"

M.

far/FarEng.hlf.m4
MKadaner: Thanks.

MKadaner:
LGTM. Thank you.

> Sorry for all the "churn"

Never mind. I do it all the time; have never been reprimanded. 😀

mihnita:
> Never mind. I do it all the time; have never been reprimanded. 😀

If I am not reprimanded I might come back :-)

Mihai

alabuzhev (approved):

alabuzhev:
Thank you

mihnita:
Thank you very much, everybody!
M

Issue #1005, opened 2025-09-18
DM_LISTFINDSTRING finds incorrect items if the text contains path separators


rohitab:
### Far Manager version

3.0.6557.0 x64

### OS version

10.0.22631.5768

### Steps to reproduce

Run the following command:
```
lua:Keys("ShiftF4");dialog=actl.GetWindowInfo().Id;pos=dialog:send(far.Flags.DM_LISTFINDSTRING, 6, {Pattern="c*"});if pos then far.Show("Found", "Position: "..pos, "Text: "..dialog:send(far.Flags.DM_LISTGETITEM, 6, pos).Text) else far.Show("Not Found") end
```

This will open the _Open/create file_ dialog and search for the pattern `c*` in the _Code page_ dropdown list using `DM_LISTFINDSTRING`. If found, the position of the item and its text will be displayed in a message box.

### Expected behavior

The following message should be displayed.
```
  ╔═══════════════╗
  ║  1│ Not Found ║
  ╚═ 1 argument ══╝
```

Since none of the entries in the _Code page_ dropdown list start with the letter `c`, the message _Not Found_ is displayed.

### Actual behavior

The following message is be displayed. Note that the _Position_ will be different if you have configured _Favorites_.
```
  ╔═══════════════════════════════════════════╗
  ║  1│ Found                                 ║
  ║  2│ Position: 11                          ║
  ║  3│ Text: 37   │ IBM EBCDIC - U.S./Canada ║
  ╚═══════════════ 3 arguments ═══════════════╝
```

The item is found at position `11`. The item at that position has the text `37   │ IBM EBCDIC - U.S./Canada`. Since the text does _not_ start with the letter `c`, this item should not have been found.

### Additional Information
The issue appears to be with the following code from `VMenu::FindItem`:

https://github.com/FarGroup/FarManager/blob/96d48b7ffc9fef530857655384174e8f7a758dcd/far/vmenu.cpp#L3491-L3495

The above code calls `CmpName` with `skippath` set to `true`, which results in `PointToName` being used to strip out the path. This causes the `str` to change from `37   │ IBM EBCDIC - U.S./Canada` to `Canada` before comparison. Since the stripped out string starts with the letter `C`, the match is successful.

https://github.com/FarGroup/FarManager/blob/96d48b7ffc9fef530857655384174e8f7a758dcd/far/processname.cpp#L147-L157

### Patch
The following patch fixes this issue.

```patch
--- a/far/vmenu.cpp
[...]

PR #1007, opened 2025-09-18
DM_LISTFINDSTRING finds incorrect items if the text contains path separators

rohitab:
## Summary
If a menu contains items with path separators, then `DM_LISTFINDSTRING` can return incorrect items. For example, if a menu item contains the text `IBM/PC`, then calling `DM_LISTFINDSTRING` with the pattern `p*` will return the position of this item, even though it does not start with the letter `p`.

This PR fixes this issue by ensuring that paths are not skipped when comparing menu items in `VMenu::FindItem`.

 
## References
https://github.com/FarGroup/FarManager/issues/1005

## Checklist
- [x] I have discussed this with project maintainers: https://github.com/FarGroup/FarManager/issues/1005 <br/>

## Details
Modify `VMenu::FindItem` so that `CmpName` is called with `skippath` set to `false` instead of `true`. This only happens for the case when `Flags` do _not_ include `LIFIND_EXACTMATCH`.


alabuzhev:
Thank you

Issue #1011, opened 2025-09-20
Find all results menu does not show the complete line


rohitab:
### Far Manager version

3.0.6558.0 x64

### OS version

10.0.22631.5909

### Steps to reproduce

1. Open a file with very long lines in the Editor. For example, a minified JavaScript file.
2. Press `F7` and enter any word that is present in the file.
3. Press `Alt+A` to open the Find all results menu.
4. Press `Alt+End` to go to the end of the line.

### Expected behavior

The complete line is visible, including any highlighted words.
Here is a screenshot from Far version `6446` that shows the complete line including highlighted words.

<img width="1802" height="518" alt="Image" src="https://github.com/user-attachments/assets/f72fa30e-f2a7-4532-978e-97e28dc86d64" />

### Actual behavior

The complete line is _not_ visible.
Here is a screenshot from the current Far version `6558` that shows the issue.

<img width="1802" height="518" alt="Image" src="https://github.com/user-attachments/assets/c1c6b62a-627a-40ac-b408-48b1256e82c2" />

MKadaner:
Thank you for reporting, @rohitab.

Please do not come with menu items longer than 2GiChars. At least not in the next decade.

rohitab:
Fixed in version `3.0.6561.0 x64`.

Issue #1010, opened 2025-09-20
Annotations not aligned correctly in Find all results menu


rohitab:
### Far Manager version

3.0.6558.0 x64

### OS version

10.0.22631.5909

### Steps to reproduce

1. Open a file with very long lines in the Editor. For example, a minified JavaScript file.
2. Press `F7` and enter a word that is towards the middle/end of the long line.
3. Press `Alt+A` to open the Find all results menu.
4. Press `Ctrl+Clear` to align annotations.

### Expected behavior

Annotations are aligned.
Here is a screenshot from Far version `6446` that shows correctly aligned annotations.

<img width="1802" height="518" alt="Image" src="https://github.com/user-attachments/assets/8845dfe8-dddb-4699-9fe0-4ab28fc03746" />

### Actual behavior

Annotations are not aligned. In fact, the annotations are not visible at all.bar.
Here is a screenshot from the current Far version `6558` that shows the issue.

<img width="1802" height="518" alt="Image" src="https://github.com/user-attachments/assets/b4c9955d-473a-4d7e-951b-3a21046ee28c" />

MKadaner:
Thank you for reporting, @rohitab.

Please do not come with menu items longer than 2GiChars. At least not in the next decade.

rohitab:
Fixed in version `3.0.6561.0 x64`.

Issue #1009, opened 2025-09-20
Incorrect column number in Find all results


rohitab:
### Far Manager version

3.0.6558.0 x64

### OS version

10.0.22631.5909

### Steps to reproduce

1. Open a file with very long lines in the Editor. For example, a minified JavaScript file.
2. Press `F7` and enter a word that is towards the middle/end of the long line.
3. Press `Alt+A` to open the Find all results menu.
4. Select an entry and press `Ctrl+Enter` to navigate to that position in the Editor.

### Expected behavior

The column number shown in the Find all results menu matches the column number shown in the Editor status bar.
Here is a screenshot from Far version `6446` that shows the correct column number `2545605`.

<img width="1802" height="518" alt="Image" src="https://github.com/user-attachments/assets/7d717c3d-4a49-489f-94a0-c4f61daaf141" />

### Actual behavior

The column number shown in the Find all results menu does not match the column number shown in the Editor status bar.
Here is a screenshot from the current Far version `6558` that shows incorrect column numbers in the Find all results menu. Instead of column `2545605` it shows column `2254560`. The other column numbers are incorrect too.

<img width="1802" height="518" alt="Image" src="https://github.com/user-attachments/assets/85f801cb-468f-48c7-b132-ecca2d0d9973" />

MKadaner:
Thank you for reporting, @rohitab.

Please do not come with menu items longer than 2GiChars. At least not in the next decade.

rohitab:
Fixed in version `3.0.6561.0 x64`.

PR #1012, opened 2025-09-20
gh-1009: gh-1010: gh-1011: An item in Find All menu can be longer than max short.

MKadaner:
## Summary
An item in Find All menu can be longer than max short. Who could think?

 
## References
- #1009
- #1010
- #1011

## Checklist


far/editor.cpp
```diff
@@ -7123,6 +7123,22 @@ TEST_CASE("radix10_formatted_width")
 		{
 			return radix10_formatted_width(Value) == static_cast<short>(std::format("{}", Value).size());
 		}));
+	REQUIRE(std::ranges::all_of(
```
alabuzhev, line 7126: This [fails](https://github.com/FarGroup/FarManager/actions/runs/17882812760/job/50852661007#step:7:291) on x86.

MKadaner: Yes, I know. I'll fix it in half an hour.

BTW, do you want three separate commits like now or squash them together?

alabuzhev: I prefer commits to represent logically separate actions.
E.g. if there's a bugfix and a pre- or post-refactoring that makes sense in the context, then of course, no point in squashing them together (each should still be compilable).

On the other hand, there's no point in preserving separate commits that "just happen", e.g. some flow of thoughts or experiments or addressing comments etc.

MKadaner: Here, I have a bugfix, somewhat (un)related refactoring, and changelog + vbuild. I think it makes more sense to have them separate.

alabuzhev:
Thanks

MKadaner:
Thank you!

Issue #611, opened 2023-01-21
LuaMacro plugin: Enable win.GetFileTimes/win.SetFileTimes to get/set change time


rohitab:
### Description of the new feature or improvement

The Lua function `win.GetFileTimes` returns _Creation time_, _Last access time_, and _Last write time_ as shown in the following code. It does not return _Change time_.

https://github.com/FarGroup/FarManager/blob/2326d88085452dde8658d561f200ee3295a5a0ea/plugins/luamacro/luafar/win.c#L863-L870

Similarly, `win.SetFileTimes` sets the same time values as shown in the following code. It does not set _Change time_.

https://github.com/FarGroup/FarManager/blob/2326d88085452dde8658d561f200ee3295a5a0ea/plugins/luamacro/luafar/win.c#L917-L923

Could you please add support for getting and setting _Change time_ as well? This would allow Lua macros to set and restore all file times, similar to the _Preserve all timestamps_ feature in the _Copy_ and _Rename/Move_ dialogs.

### Proposed technical implementation details (optional)

Use `file::GetTime` and `file::SetTime` instead of `GetFileTime` and `SetFileTime` respectively.

https://github.com/FarGroup/FarManager/blob/2326d88085452dde8658d561f200ee3295a5a0ea/far/platform.fs.hpp#L258-L260

johnd0e:
@shmuz

shmuz:
@johnd0e, as you know I'm spending all my spare time nowadays on far2m, etc. and there's no `ChangeTime` there.

johnd0e:
@shmuz, yes I know.
But still hope that some day you will be able to find some time.
Or at least give us your opinion about proposed implementation.

shmuz:
1. It's not a proposed implementation here, it is just a suggestion to use the same code Far3 uses.
2. I respect you being a Far3 "patriot", but on the other hand you could help me a lot with far2m but you won't do that.

johnd0e:
> you could help me a lot with far2m but you won't do that.

I will be happy to help where I can.
At the moment I have no system where I could run far2m.

rohitab:
@johnd0e Thank you for trying to get this issue resolved.

> 1. It's not a proposed implementation here, it is just a suggestion to use the same code Far3 uses.

@shmuz The suggestion and the proposed implementation are the same, i.e. to use different functions, ones that get/set 4 file times instead of 3. I'd be happy to submit a patch if someone can review and commit it.

[...]

PR #1014, opened 2025-09-27
LuaMacro plugin: Enable win.GetFileTimes/win.SetFileTimes to get/set change time

rohitab:
## Summary
The Lua API's `win.GetFileTimes` and `win.SetFileTimes` currently allow retrieving and setting _Creation time_, _Last access time_, and _Last write time_. This PR adds support to get and set _Change time_ as well.

 
## References
https://github.com/FarGroup/FarManager/issues/611

## Checklist
- [x] I have discussed this with project maintainers: https://github.com/FarGroup/FarManager/issues/611<br/>

## Details
The functions `win.GetFileTimes` and `win.SetFileTimes` have been modified to use `NtQueryInformationFile` and `NtSetInformationFile` to get/set time. These are the same API's used by _Far_.

The API's work as before, except that the table has an additional field `ChangeTime` that can be set and retrieved. As before, `win.SetFileTimes` supports setting one, all or any combination of the available time values.

### DIFF
--- a/plugins/luamacro/luafar/lf_win.c
+++ b/plugins/luamacro/luafar/lf_win.c
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 #include <windows.h>
 #include <shellapi.h>
+#include <winternl.h>
 
 WARNING_PUSH()
 WARNING_DISABLE_MSC(4255)
@@ -858,10 +859,9 @@ int win_IsWinVersion(lua_State *L)
 	return 1;
 }
 
-static void PutFileTimeToTableEx(lua_State *L, const FILETIME *FT, const char *key)
+static void PutFileTimeToTableEx(lua_State *L, const LARGE_INTEGER *FT, const char *key)
 {
-	INT64 FileTime = FT->dwLowDateTime + 0x100000000LL * FT->dwHighDateTime;
-	bit64_pushuserdata(L, FileTime);
+	bit64_pushuserdata(L, FT->QuadPart);
 	lua_setfield(L, -2, key);
 }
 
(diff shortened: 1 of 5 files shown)


plugins/luamacro/luafar/lf_win.c
```diff
@@ -876,14 +876,23 @@ static int win_GetFileTimes(lua_State *L)
...
-				lua_createtable(L, 0, 3);
-				PutFileTimeToTableEx(L, &t_create, "CreationTime");
-				PutFileTimeToTableEx(L, &t_access, "LastAccessTime");
-				PutFileTimeToTableEx(L, &t_write,  "LastWriteTime");
-				res = 1;
+				IO_STATUS_BLOCK iob;
+				FILE_BASIC_INFO fbi;
+				const int FileBasicInformation = 4;
+				typedef NTSTATUS(NTAPI* NtQueryInformationFile)(HANDLE FileHandle, PIO_STATUS_BLOCK IoStatusBlock, PVOID FileInformation, ULONG Length, FILE_INFORMATION_CLASS FileInformationClass);
+				NtQueryInformationFile pNtQueryInformationFile = (NtQueryInformationFile)(INT_PTR)GetProcAddress(hNtDll, "NtQueryInformationFile");
```
alabuzhev, line 886: Minor:
PutFileTimeToTableEx function is supposedly called more than once, and calling GetProcAddress here every time adds redundant runtime overhead: it's not like ntdll can suddenly disappear.
It can be made static as [elsewhere](https://github.com/FarGroup/FarManager/blob/b5ef796f9f69eeb983b6823369ac5ce940cd111b/plugins/arclite/sysutils.cpp#L221-L222).

Same for ExtractFileTime.

rohitab: @alabuzhev thank you for reviewing the code.

> Minor: PutFileTimeToTableEx function is supposedly called more than once, and calling GetProcAddress here every time adds redundant runtime overhead: it's not like ntdll can suddenly disappear.

`PutFileTimeToTableEx` is indeed called `4` times, one for each time value. However, `GetProcAddress` is called only once for each call to `win_GetFileTimes` and `win_SetFileTimes`. But regardless, the issue you mentioned is still there.

You're right to call this out. I ran through a couple of options when I was modifying the code.

-  Use an initialization function, similar to what ProcList does in [`dynamic_bind`](https://github.com/FarGroup/FarManager/blob/b5ef796f9f69eeb983b6823369ac5ce940cd111b/plugins/proclist/Proclist.cpp#L92-L130).
- Initialize function pointers in [`DllMain`](https://github.com/FarGroup/FarManager/blob/b5ef796f9f69eeb983b6823369ac5ce940cd111b/plugins/luamacro/luafar/lf_service.c#L154-L170).
- Store the result of `GetProcAddress` in a static variable in each function.
- Link with `ntdll.lib`, either the standard one, or a custom-built lib with just these two exports.

Eventually, I decided to stick with the approach that was used in another function [`win_DeleteRegKey`](https://github.com/FarGroup/FarManager/blob/b5ef796f9f69eeb983b6823369ac5ce940cd111b/plugins/luamacro/luafar/lf_win.c#L214-L233) in the same source file [`luamacro/luafar/lf_win.c`](https://github.com/FarGroup/FarManager/blob/b5ef796f9f69eeb983b6823369ac5ce940cd111b/plugins/luamacro/luafar/lf_win.c#L214-L233)

> It can be made static as [elsewhere](https://github.com/FarGroup/FarManager/blob/b5ef796f9f69eeb983b6823369ac5ce940cd111b/plugins/arclite/sysutils.cpp#L221-L222).

Unfortunately, that won't work as-is, since LuaFAR code is in C, and function calls aren't allowed in constant expressions. But we could still store the result in a static variable afterwards to avoid reinitialization.

Personally, I would have preferred to use `ntdll.lib`, but I didn't want to make changes that were not approved and outside of the feature I was working on. However, I'm happy to make any changes that are needed. What would you suggest?

shmuz: > Minor: PutFileTimeToTableEx function is supposedly called more than once, and calling GetProcAddress here every time adds redundant runtime overhead <snip>
> 
> Same for ExtractFileTime.

I made a commit that seems to address this.

alabuzhev: > PutFileTimeToTableEx is indeed called 4 times, one for each time value. However, GetProcAddress is called only once for each call to win_GetFileTimes and win_SetFileTimes.

Apologies, I meant win_SetFileTimes and win_GetFileTimes of course - GitHub diff view collapsed the lines in a confusing way.
The point was that GetProcAddress is called every time the user wants to get or set something.

> Link with ntdll.lib

Oh, that's interesting. Is it part of WinSDK now and available by default and an officially blessed way to call Nt* funuctions? It used to be not the case. If so, and if it's also available in non-Microsoft SDKs (MSYS2, LLVM) we might want to switch to that approach.

> I made a commit that seems to address this.

Thanks, perhaps it would make sense to do that for win_DeleteRegKey too then?

shmuz: > perhaps it would make sense to do that for win_DeleteRegKey too then?

Done, thanks.

rohitab: > > Link with ntdll.lib
> 
> Oh, that's interesting. Is it part of WinSDK now and available by default and an officially blessed way to call Nt* funuctions?

Yes, it part of Windows SDK for x86, x64, arm and arm64. I'm not sure when Microsoft started including it, but since it's there now, I assume it's the official way to use API's, at least the ones that are guaranteed to be available on the supported platform.

> It used to be not the case. If so, and if it's also available in non-Microsoft SDKs (MSYS2, LLVM) we might want to switch to that approach.

I'm not sure about others, but MSYS2 comes with a [lib for NTDLL](https://packages.msys2.org/packages/mingw-w64-cross-mingw32-crt?variant=x86_64) and the build with gcc works using that library. For the ones that don't support it, a simple def file like this one will work.

```def
LIBRARY ntdll.dll
EXPORTS
    NtSetInformationFile
    NtQueryInformationFile
```
It does make the code much cleaner without all the `LoadLibrary`/`GetModuleHandle`/`GetProcAddress` calls.

PR #1021, opened 2025-10-02
New feature: panel resizing with mouse

yulian5:
## Summary
The PR introduces a new feature to Far Manager: panel resizing with mouse.

 
## Checklist
- [ X] I have followed the [contributing guidelines](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTING.md).

## Details
This feature introduces the ability to resize file panels in Far Manager using the mouse.

Users can now:

- Hover over the border between the left and right panels to see visual feedback indicating a resizable area (grip).
- Click and drag the border left or right to dynamically adjust the width of the panels.
- Double-click on the border to quickly restore the default 50/50 equal width split for the panels.

This enhancement provides a more intuitive and efficient way to manage panel layouts, improving user experience for those who prefer mouse-driven interactions.

It includes updates to:

FarColorW.pas and farcolor.hpp: Defines a new palette color COL_PANELDRAGBORDER for the visual feedback during resizing.

farlang.templ.m4: Adds new localization strings for the "Dragging border" text.

filepanels.cpp and filepanels.hpp: Implements the core logic for mouse event handling, border detection, resizing, and drawing the visual feedback. This includes new MouseState enum, IsMouseOverPanelBorder, DrawBorderFeedback, ClearBorderFeedback, and ResetAllMouseStates functions.

palette.cpp and setcolor.cpp: Initializes and integrates the new COL_PANELDRAGBORDER color into the color palette and settings.

*.hlf.m4 files (e.g., FarEng.hlf.m4, FarRus.hlf.m4): Updates help files for various languages to document the new mouse resizing functionality.


yulian5:
Do I need to fix any of SonarQubeCloud suggestions?
Thanks!

alabuzhev:
@yulian5, thanks for your PR.
I haven't looked at the changes yet, just some general comments:

https://github.com/user-attachments/assets/155be9de-5566-4c8f-b0ef-f1e2acbb38fd

Functional:
- It conflicts with the left panel scrollbar.
- It triggers too early, when the mouse is over the cells adjacent to the panel border:
  - It conflicts with the right panel sort indicator (another way to call Sort Menu & Disk Menu).
  - It conflicts with clicks on file names in both panels.

Visual:
- It assumes that both panels are of the same height, which is not always the case.
- Panels can refresh while hovering and partially or fully overwrite the indicator.

It's an interesting feature, but, as you can see, the place is pretty crowded already.
Which is one of the reasons we didn't implement something like this years ago.

yulian5:
@alabuzhev thank you very much for your testing and valuable inputs. Let me work on remaining issues. Thanks!

yulian5:
Hi @alabuzhev, I have the working code which addressed all mentioned issues:

far-panels-resizing2

The scrollbar is excluded from mouse resize detection and works as expected, but even with the scrollbar there is still enough grip area for effective resizing even in case of a single panel (what as I see ConEmu cannot do). For next PR I need to finish vertical resizing to complete the task where option for single panel resizing will be useful as well.

I need to clean up the code, add comments, check it for optimization, and then commit new changes. Do I need to modify changelog and vbuild.m4 again or it'll be done during the merge?

Thanks!

yegor-mialyk:
Do we really want to introduce a new color for this? It would be a breaking change for all custom themes.

HamRusTal:
> there is still enough grip area for effective resizing even in case of a single panel

On the screenshot, I see that the {left panel's cell below the scroll bar where the “status” area separator joins the right-hand-side frame} is **not highlighted** yellow whereas the corresponding right panel's cell on its left **is** (no scrollbar there though). Is it for a purpose or just a bug?

**BTW** Do we need to highlight **both** panels' frames? IMHO just the one being dragged is enough (at least while dragging). We might still highlight both frames _on hover_ though to indicate that both will be affected.

**Also**, it would be nice to handle the **`Esc`** key press (or RMB click) _while dragging_ to cancel the resize and restore the initial geometry (similar to Windows' window move/resize except for RMB click).

HamRusTal:
> Do we really want to introduce a new color for this? It would be a breaking change for all custom themes.

New color would be nice and flexible. But the code has to handle the missing color **gracefully** and use a sensible default replacement from the colors defined by the current theme.

alabuzhev:
> Do I need to modify changelog and vbuild.m4

Don't worry about these for now. There might be more changes requested which might take a few days and resolving conflicts every time someone merges into master doesn't make much sense.
Once we're ready to merge I'll let you know and those files can be updated then.

> use a sensible default replacement from the colors defined by the current theme

We do not have such a concept. Palette colors are independent and any missing ones will stay default aqua-on-blue-ish.
It might be useful to discuss and explore this more.
For example, background color in quite a few palette items (e.g. panel borders or column titles) is redundant and can be transparent and inherited from panel text or whatever would be previous in Z-order.
However, it might be too much for this particular PR.
Feel free to create an issue or a discussion.

yulian5:
@yegor-mialyk 
> Do we really want to introduce a new color for this? It would be a breaking change for all custom themes.

@alabuzhev 
> It might be useful to discuss and explore this more.

No, it doesn't break anything. All custom color themes work just fine. Yes, we need a new color. There is enum for it and menu entry. Here is sample, you can compile code and try for yourself.

color_border

I realized from the beginning that custom color themes obviously do not have a new color. But they can define it now. The fallback scheme is following: if color for dragging border is not defined in the theme then we use foreground color from dragging text and background color from normal text (panel background). And all color themes I checked have both colors defined. And in case they don't have them it will be yellow on blue that can be changed easily (but it should be something really broken):

custom

It looks natural and work on all themes included in Far by default. It still may be some problems here but I'm sure it's possible to fix.

yulian5:
@HamRusTal 

> New color would be nice and flexible. But the code has to handle the missing color gracefully and use a sensible default replacement from the colors defined by the current theme.

Yes, I explained that. Colors were my first concern in this feature!

yulian5:
@HamRusTal 

> On the screenshot, I see that the {left panel's cell below the scroll bar where the “status” area separator joins the right-hand-side frame} is not highlighted yellow whereas the corresponding right panel's cell on its left is (no scrollbar there though). Is it for a purpose or just a bug?

Yes, I notice that. The scrollbar at the bottom actually expands more than it depicted and this area is clickable and it scrolls the context down. This looks like the problem with scrollbar implementation. The top part of scrollbar looks fine. We may need to create new issue for that later.

yulian5:
@HamRusTal 

> BTW Do we need to highlight both panels' frames? IMHO just the one being dragged is enough (at least while dragging). We might still highlight both frames on hover though to indicate that both will be affected.

Actually I had the similar concern. I tried to make single lines while dragging but it looked awful. You can change dragging border color to normal color, it works very nice except you cannot see when it hovered (but if you know then it's fine). It's very possible to make separate colors for havering and dragging, not a problem at all. Do we need it? It maybe very confusing.

The single panel can be resized with mouse so to simplify the logic both panel borders are highlighted.

> Also, it would be nice to handle the Esc key press (or RMB click) while dragging to cancel the resize and restore the initial geometry (similar to Windows' window move/resize except for RMB click).

Yes, it's a very good idea, but that would be a completely separate task. Mouse and key handling in Far is quite complicated even without plugins and macros. Double click works fine for now.

HamRusTal:
> The scrollbar at the bottom actually expands more than it depicted and this area is clickable and it scrolls the context down. This looks like the problem with scrollbar implementation.

Actually, the entire panel frame below the scrollbar (and above it too) performs scrolling when clicked. It looks like a deliberate feature (with its own bugs though). But it does not justify that the resizing drag-highlight does not cover the cell I'm referring to. IMHO this should be fixed in the PR to get a more consistent look and feel.

> I tried to make single lines while dragging but it looked awful.

I am not suggesting to use single-line highlighted borders, I am rather suggesting to highlight only **one panel**'s (double-line) border. This will be consistent with resizing of a single visible panel too.

> Double click works fine for now.

Double-click does not address the scenario when there's some non-default panel width already and the user starts dragging the panel border (maybe by a misclick). Unfortunately, double-click will not restore the pre-drag non-default layout.

yulian5:
@HamRusTal 

> Actually, the entire panel frame below the scrollbar (and above it too) performs scrolling when clicked.

Yes, I already figured is out. I agreed with you and took one cell from the scrolling area:

scroll_area


It displayed as grip area but not working yet.

> I am rather suggesting to highlight only one panel's (double-line) border. This will be consistent with resizing of a single visible panel too.

But which one I have to highlight? If I highlight right border and the right panel is hidden? I need to change left/right highlighting depending which panel is hidden And another question: the grip area is actually 2 borders, displaying only have of grip area will be confusing. If it's so important maybe just make it configurable. 

> Double-click does not address the scenario when there's some non-default panel width already and the user starts dragging the panel border (maybe by a misclick). Unfortunately, double-click will not restore the pre-drag non-default layout.

I agree. Let me check that. If it's easy to implement I'll add it otherwise it will another PR.

BTW, I added height resizing with mouse:

https://github.com/user-attachments/assets/7687dabc-e249-4b23-befc-93dae8961306

yulian5:
@alabuzhev 
> Don't worry about these for now. ...............

Thank you very much for detailed explanation! I updated the code. It happens to be quite a huge project!

@HamRusTal 
I added Esc key feature: it's actually more useful for resizing that I excepted and I fixed multiple issues during its implementation.

far/filelist.cpp
```diff
@@ -722,7 +722,6 @@ class list_less
 
 		return false;
 	}
-
```
yegor-mialyk, line 725: Please do not delete an empty line between methods

yulian5:
Hi @alabuzhev 

I uploaded new code which fixed calculation of bottom border around text area for visual feedback during height resizing and fixed SonarQubeCloud issues.

However CodeFactor marked the code with red cross because of complexity GetInputRecordImpl() in keboard.cpp file. That's not my code and I don't think that I need to fix it now. I just added couple small functions to block all keys during resizing and allow only Esc key to revert resizing. Please help with that!

Thanks!

alabuzhev:
> complexity GetInputRecordImpl() in keboard.cpp file

It's fine to ignore.

far/filelist.cpp
yulian5: @yegor-mialyk  You're right, I'll fix that, some stupid formatting tool.

yulian5: There are still a lot mis-formatted places talking only about filelist.cpp.

yulian5:
I merged all my commits into one and updated changelog and vbuild.m4 to the latest. But GitHub doesn't merge it in PR automatically.

yegor-mialyk:
I think we need to revert all removed empty lines back in `fielist.cpp`

yulian5:
@yegor-mialyk
>  think we need to revert all removed empty lines back in fielist.cpp

I did that 2 days ago. It was my mistake. I tried to remove spaces that VS inserts without any reason. 

Do you see any removed single empty lines? But I converted some double empty lines to single empty lines between functions just to be consistent through the file. Only one double empty line is remaining before "namespace".

yegor-mialyk:
Oops, it's my fault, those removed empty lines are duplicates.

yulian5:
Hi @yegor-mialyk, I see the review is going to be long. I guess you're very busy. Do you need any help from me? I can briefly describe the algorithm and the main changes to the code. Please let me know. Thanks!

far/FarEng.hlf.m4
```diff
@@ -884,8 +887,8 @@ default button or does another similar thing).
  #Mouse#
 
  Clicking the #middle mouse button# in the ~panels~@PanelCmd@ has the same
-effect as pressing the #Enter# key with the same modifiers (#Ctrl#, #Alt#,
-#Shift#). If the ~command line~@CmdLineCmd@ is not empty, its contents will be executed.
+ effect as pressing the #Enter# key with the same modifiers (#Ctrl#, #Alt#,
```
alabuzhev: Please do not add leading spaces to random lines.
In hlf syntax a leading space means a new paragraph, it will break the formatting, hlf files will go out of sync and make the build checks unhappy.

Same for other places and files.

far/FarEng.hlf.m4
```diff
@@ -1160,6 +1163,30 @@ if the initial panel type differs it will be automatically set to the file panel
...
+ - ^<wrap>Click and drag the bottom border up or down to adjust the panel height.
+ - ^<wrap>The resizing behavior depends on the horizontal mouse position:
+   - ^<wrap>Over LEFT panel bottom border in Left half of panel → Resize LEFT panel only
+   - ^<wrap>Over LEFT panel bottom border in Right half of panel → Resize BOTH panels
+   - ^<wrap>Over RIGHT panel bottom border in Left half of panel → Resize BOTH panels
+   - ^<wrap>Over RIGHT panel bottom border in Right half of panel → Resize RIGHT panel only
+ - ^<wrap>Double-click on the bottom border to restore full panel height.
+ - ^<wrap>Press #Esc# during resizing to cancel and revert to the original panel heights.
+
+ #Note for ConEmu users:#
```
alabuzhev: This belongs in ConEmu's documentation, not here.

far/farlang.templ.m4
```diff
@@ -18791,6 +18791,21 @@ MSetColorPanelDragging
...
+"Граница перетаскивания"
+"Dragging border"
+"Tažená hranice"
+"Drag && Drop Border"
+"Vonszolt szegély"
+"Przeciągana krawędź"
+"Texto arrastrado"
+"Potiahnutý okraj"
+"Spostamento Bordo"
+"Межа перетягув."
```
alabuzhev: Why does it stop abruptly?
We should have enough space for the remaining 4 letters.

far/FarEng.hlf.m4
```diff
@@ -365,9 +365,12 @@ $ #Panel control commands#
  Hide/show right panel                                      #Ctrl+F2#
  Change panels height                             #Ctrl+Up,Ctrl+Down#
  Change current panel height          #Ctrl+Shift+Up,Ctrl+Shift+Down#
- Change panels width                           #Ctrl+Left,Ctrl+Right#
- (when the command line is empty)
+ Change panels height with mouse       #Drag bottom border of panels#
```
alabuzhev: Please keep it consistent with the rest (2 lines).

Same below.

far/FarEng.hlf.m4
```diff
@@ -365,9 +365,12 @@ $ #Panel control commands#
...
+ Change panels width with keyboard             #Ctrl+Left,Ctrl+Right#
+ (only when the command line is empty)
+ Change panels width with mouse          #Drag border between panels#
  Restore default panels width                          #Ctrl+Numpad5#
+                        #Mouse double click on border between panels#
```
alabuzhev: Formatting looks strange

far/filelist.cpp
```diff
@@ -121,7 +121,6 @@ THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 static_assert(static_cast<size_t>(panel_sort::BY_USER) == static_cast<size_t>(OPENPANELINFO_SORTMODES::SM_USER));
 
-
```
alabuzhev, line 124: Pls avoid changing formatting, adding or removing lines etc. if it's not directly related to your change.

It could make sense in isolation, or subjectively look better or whatever, but all these tiny changes here and there distract from the important part and only make the reviewing longer.

far/filepanels.cpp
```diff
@@ -1171,18 +1222,629 @@ bool FilePanels::ProcessMouse(const MOUSE_EVENT_RECORD *MouseEvent)
...
+	const auto& DragTextColor = colors::PaletteColorToFarColor(COL_PANELDRAGTEXT);
+	const auto& SelectedTextColor = colors::PaletteColorToFarColor(COL_PANELSELECTEDTEXT);
+	const auto& PanelTextColor = colors::PaletteColorToFarColor(COL_PANELTEXT);
+
+	// Check if the dragging border color is using the default palette value
+	// Default palette value for COL_PANELDRAGBORDER is F_YELLOW|B_BLUE
+	const auto DefaultDragBorderColor = F_YELLOW | B_BLUE;
+
+	// If the current color matches the default palette value, or if it seems uninitialized,
+	// apply the fallback logic (theme doesn't define this color or user hasn't customized it)
```
alabuzhev: Please don't overcomplicate it.
If the user hasn't customized it then that's exactly what the user wanted.
No code anywhere in Far should assume incomplete themes or anything like that.
The changes in this file are already too massive even without all this.

far/FarEng.hlf.m4
yulian5: Yes, you're right, let me remove it.

far/FarEng.hlf.m4
yulian5: @alabuzhev, It took a lot of copy/pasting and editing back and forth, let me fix the remaining issues.

far/FarEng.hlf.m4
yulian5: Hi Alex, thank you for the feedback!

I don't quite understand what you mean. But I've improved the appearance of the help section, and by the way, I forgot to add a description of how to restore the height using the mouse. Here's what it looks like now:

help

Is it OK? If so, I need to update the information in other languages ​​as well.

yulian5:
Highlighting borders on hover and drag (visual feedback) is quite convenient and helps users use the function without any instructions. But over time, it starts to get a little annoying. I'm thinking of adding an option to disable visual feedback for experienced users. What are your thoughts on this?

far/FarEng.hlf.m4
alabuzhev: Sorry, github line wraps made it confusing.
Pls ignore this particular comment.

far/filepanels.cpp
yulian5: @alabuzhev, if I remove that logic all custom color themes will have broken dragging border color. 

> If the user hasn't customized it then that's exactly what the user wanted.

They cannot customize it yet because they didn't know about that feature at all. But they can customize it now if needed.

The logic here is actually very simple: we have default dragging border color which is defined as F_YELLOW|B_BLUE in palette.cpp (it may be other combinations). If color theme has its own color (we compare with default) then we use it. Otherwise we use dragging text color for foreground color and normal text background for background color. The same in case if color is invalid for some reason. The users don't need to do anything, dragging border color just works.

> No code anywhere in Far should assume incomplete themes or anything like that.

Yes, that's most likely correct. That's why last year when you fixed transparent colors it broke all Far color themes and required changing and reapplying Far settings or reinstall the Far manager with deleting profiles. Because that change was not backward compatible for reasons you just mentioned. 

In opposite, it may be will be quite useful to add fallback logic in Far Manager, not only for colors but for other things as well.

> The changes in this file are already too massive even without all this.

Unfortunately it's the case. Originally I just wanted to introduce only width resizing with mouse which I used this year.  But then visual feedback, colors, different panel heights, height resizing, Esc key. Everything is needed and cannot be separated now. The help files are just nightmare.

far/FarEng.hlf.m4
yulian5: @alabuzhev It's OK. These help files are real nightmare, I spent half time on them.

Russian help doesn't fit a little, the black and white text without space.. All others languages look OK:

help rus

yulian5: > Pls ignore this particular comment.

Actually it was quite useful, I added text about height resizing with mouse I missed before.

far/filepanels.cpp
alabuzhev: >  we have default dragging border color which is defined as F_YELLOW|B_BLUE in palette.cpp (it may be other combinations). If color theme has its own color (we compare with default) then we use it. Otherwise we use dragging text color for foreground color and normal text background for background color.

And what if the user *wants* it to be yellow-on-blue in otherwise red-on-green theme?
Because it's his heart's desire or whatever?

> That's why last year when you fixed transparent colors it broke all Far color themes

That issue:
- Is unrelated to the potential issue that is being discussed here.
- Happened exactly because of the logic being too smart.
- Did not require reinstalling or deleting profiles etc.
  It was enough to go to corresponding color settings and uncheck "transparent".

> it may be will be quite useful to add fallback logic in Far Manager

Perhaps, but it should happen structurally and not for one particular color.

> Everything is needed and cannot be separated now. The help files are just nightmare.

Exactly. It's already hard for you as the author and way harder for anyone looking at it for the first time.
If you want this PR merged eventually, please try to make it as small as possible.
This means no good-to-have changes, no reformatting, and no non-trivial color handling.
We will update default color presets separately.

MKadaner:
I have a somewhat unrelated concern, and I apologize that I have not looked at the actual change. So, here it goes: if I have mouse support turned off in `Interface settings`, will I incur any performance penalty (besides a bit larger memory footprint) when this feature is merged?

yulian5:
> will I incur any performance penalty (besides a bit larger memory footprint) when this feature is merged?

Hi @MKadaner:

Based on code analysis:

**The answer is: NO, there will be NO performance penalty when mouse support is turned off.**
The reason is Mouse Event Processing is Conditional in Event-Driven Architecture with No Background Processing.

As for Memory Footprint: 
**This feature adds approximately < 100 bytes of memory per FilePanels instance.**

HamRusTal:
> < 100 bytes of memory per FilePanels instance

I assume you mean the additional data, yet the added instructions code will be larger I believe (still not an issue IMHO).

yulian5:
@MKadaner and @HamRusTal 

Nothing in this life comes for free. Here is the full estimates:

**x64 Release Build Footprint Estimate:**

**Code Segment (.text):**
    • Average x64 instruction density: ~3-4 bytes per line of optimized C++ code
    • 640 lines × 3.5 bytes = ~2,240 bytes (~2.2 KB)
**Data Segment (.data/.bss):**
    • Mouse state variables: ~50 bytes
    • Static constants: ~30 bytes
    • Total data: ~80 bytes
**Virtual Function Table entries:**
    • New virtual methods: ~40 bytes
**Total Memory Footprint: ~2.4 KB per process**

This is significantly more than initial "< 100 bytes" estimate, which only considered data variables. 
The actual code footprint is ~2.4 KB due to the substantial amount of mouse handling logic added.

far/filepanels.cpp
yulian5: Hi @alabuzhev,

I realized that we had a significant misunderstanding on this matter: you are talking about custom color themes, while I am talking about the color themes that come bundled with Far. We were trying to explain completely different things to each other.

> If the user hasn't customized it then that's exactly what the user wanted.

Absolutely agree. We cannot support user defined color themes. After introducing any feature with additional colors users may adjust colors to their needs.

> And what if the user wants it to be yellow-on-blue in otherwise red-on-green theme?

You're right, that's logic is broken now for default dragging color. Original "non-optimized" code was 3 times bigger.

The reason for this code was simple: obviously Far cannot be released with bundled **broken** color themes. And I didn't want to touch somebody's code for ethical and possible copyright reasons. I probably should ask in advance. Instead I created function that works as memory patch to compensate missing hovering/dragging color. It works for bundles and user defined color schemes as well. 

For user defined schemes that function doesn't make much sense: after the user changes dragging border color, the code uses that new color, so in this case my code works as a one-time function. 

As can I see the "extra\Addons\Colors" code is under the same license as other code so the bundled color scheme may have one new color. It will be much better that having an emulated color logic (for one-time use in some cases). 

Such kind of logic should be used mostly during the first initialization. As you said we may add something like this separately.

The same algorithm for hovering/dragging color (foreground color from `Panel.DragText` foreground and background color from `Panel.Text` background) works fine for all themes, except:

```
black_and_white.farconfig
Colors_from_admin_essp_ru.farconfig
FARColors242.farconfig
```
Let me figure out the consistent colors for those schemes and I'll submit the updated code.

All other issues are fixed, this one is the last (I hope). Also some language help files were not updated for panel height resizing info, I'm fixing them too.

alabuzhev: > obviously Far cannot be released with bundled broken color themes

It absolutely can. The out-of-box color themes were broken for 13 years (or even longer), see d478c218447af9dfc8c0ad1dfae8f57c8fd7c83f.
 
> I didn't want to touch somebody's code for ethical and possible copyright reasons.

It's absolutely fine.
People mentioned in the descriptions did not even create these files.
We did, based on some registry files shared by those people eons ago.
They are mentioned in the descriptions because it's the right thing to do, but there is no legal copyright for a bunch of config values.
These files were recreated, renamed, updated by project members multiple times throughout the history as we changed the config system, structure, representation, added new colors etc.
Which is, obviously, more ethical than letting them rot.

> the bundled color scheme may have one new color

Exactly. If you can come up with one for each theme, please feel free to do so.
Otherwise please just add a commented default entry, someone will update it later (see 4bce9ca9599f7e9ae70dd0c66857c39b142d6a5c for example).

> The same algorithm for hovering/dragging color (foreground color from Panel.DragText foreground and background color from Panel.Text background)

No extra algorithms please.
This new color has nothing to do with neither Panel.DragText nor with Panel.Text.
If you really want it to be smart and work well with customized colors:
1. Set the foreground to some bright color, e.g. yellow
2. Make the background transparent by default
3. Merge it with Panel.Border color, as if this new color lies on top of it in the Z-order.

This way the background will be inherited from Panel.Border, which should be reasonable for any color arrangement, and the yellow foreground should be contrast enough with anything. This should also help with your other concern:

> over time, it starts to get a little annoying. I'm thinking of adding an option to disable visual feedback for experienced users. What are your thoughts on this?

There is already an option: the experienced user can set this new color to the same value as Panel.Border (or make it fully transparent, if it's inherited as described above) and it will be gone.

yulian5: Hi @alabuzhev, thank you for detailed information! Very useful.

> No extra algorithms please.

Definitely not now.

> There is already an option: the experienced user can set this new color to the same value as Panel.Border

That's right, it will remove visual feedback. However it'll be the performance impact (quite small but anyway), even if Windows smart enough to cache the regions and do not draw the same color in the same place, Far knows nothing about that and it will continue drawing "highlighted" borders and redrawing the whole UI to remove them even if it's the same color. 

Some people very concern about performance impact (including me). But I'm not going to add anything else now. It looks like a little more optimizations are also available (for later again).

alabuzhev: > Far knows nothing about that and it will continue drawing "highlighted" borders and redrawing the whole UI to remove them even if it's the same color.

This is rather far from true.

yulian5: It's not clear what do you mean.

MKadaner:
> The actual code footprint is ~2.4 KB...

That is, of course, not a concern. Thank you for the confirmation about run-time performance being unaffected.

P.S. I prefer cats over mice.

yulian5:
@MKadaner, I myself use Far with a keyboard 95% of the time, but resizing the panels using the keyboard just drives me crazy.

Here is the full code performance analysis for **mouse processing** from claude-4-sonnet AI comparing 2 branches. It says that now it works faster than before with mouse, except resizing. But we didn't have resizing with mouse before. However I don't know how accurate it is.

**Input Caching** and **Early Returns** - optimizations from Copilot AI (however it skewed up a lot, I reverted multiple "optimizations")
**Conditional Calculations**  - optimization from me

**ProcessMouse Performance Analysis - FarManager.panel-mouse-resizing vs yulian5.master**

**Performance Impact Analysis**
Original Version (yulian5.master) - 55 Lines
Optimized Version (panel-mouse-resizing) - 363 Lines

**Performance Optimizations Added:**
    1. Input Caching (4 lines)
    2. Conditional Calculations - Expensive operations only when needed
    3. Early Returns - Avoid unnecessary processing

**New Mouse Processing Features Added:**
    • Width Resizing: 125 lines of panel width resize logic
    • Height Resizing: 120 lines of panel height resize logic
    • Border Detection: 47 lines for hover/click detection
    • Visual Feedback: 67 lines for border highlighting

**Performance Impact Assessment**

**Optimizations Gained:**
    • Memory Access: ~65% reduction in MouseEvent field access through caching
    • Calculation Overhead: ~40% reduction by deferring expensive calculations
    • Conditional Branching: ~30% improvement through early returns and state checks

**Processing Overhead Added:**
    • Code Size: 6.6x increase (55 → 363 lines)
    • Border Detection: +2-4ms per mouse event for geometric calculations
    • State Management: +1-2ms for mouse state tracking
    • Visual Feedback: +3-5ms when drawing border highlights

**Net Performance Impact:**

**For Normal Mouse Operations (95% of usage):**
    • Improvement: 15-25% faster due to input caching and early returns
    • Memory: 40% fewer memory accesses to MouseEvent fields

**For Border Interactions (5% of usage):**
    • Cost: 5-10ms additional processing for resize operations
    • Benefit: Significantly enhanced user experience with visual feedback

**Overall Assessment:**

The optimizations provide a net performance gain for typical usage while adding rich mouse interaction features. The caching and conditional logic improvements more than offset the additional border processing overhead for normal operations.

**Recommendation:** The performance optimizations are well-designed and provide both better performance for common operations and enhanced functionality for power users.

P.S. According to this analysis (if it's true) it make sense to add similar optimizations in other processing code in Far.
P.P.S. Dogs vs cats?

MKadaner:
> P.P.S. Dogs vs cats?

Sorry, it's too personal a question.

yulian5:
@alabuzhev Can you please update me on PR status? I completed all requests more than a month ago.

alabuzhev:
> Can you please update me on PR status? I completed all requests more than a month ago.

@yulian5, unlike Copilot, I'm but a human with some nasty addictions like sleeping and eating.
Those addictions are not only time-consuming, but also expensive.
Indulging in them requires resorting to even more time-consuming activities. e.g. working.

As you [already noticed](https://github.com/FarGroup/FarManager/issues/1025#issuecomment-3478910195) (thanks), I still manage to find some time for this project.
However, I would be lying if I said that getting through a generated talmud of extremely plausibly-looking, but not necessarily even remotely correct code is a pleasant pastime.

In other words, we get there when we get there.

Issue #1032, opened 2025-10-19
Crash if size of name column is less than size of mark character


rohitab:
### Far Manager version

3.0.6573.0 x64

### OS version

10.0.22631.6060

### Steps to reproduce

1. Press `Tab`, if required, to switch the Active Panel to the right.
2. Press `Ctrl+3` to switch the Active Panel to _Full_ mode.
3. Press `Ctrl+Clear` to make both panels the same width.
4. Go to _Options > Files highlighting and sort groups_ and press `Ins` to add a new entry to the top.
5. In the _Marking_ edit box, enter `123` and turn off the _Inherit_ checkbox.
6. Leave everything else as the default, then press `OK` to add the entry.
7. Press `Esc` to return back to the panels.
8. Note that each filename will display `123` before it.
9. Continuously press `Ctrl+Right` to reduce the size of the Active Panel.

### Expected behavior

Active panel width keeps reducing until the _Name_ column is `1` character wide.

### Actual behavior

Active panel width keeps reducing until the _Name_ column is `3` characters wide. Further reducing the width causes Far to crash with the following stack trace:

```
+0006055C KERNELBASE.dll!RaiseException+0x6C
+003663CB Far.exe!_CxxThrowException+0x97 (libvcruntime:throw.obj)
+0035DFD2 Far.exe!std::_Xlength_error(char const * __ptr64)+0x22 (libcpmt:xthrow.obj)
+00015CE0 Far.exe!std::_Xlen_string(void)+0x10 (cddrv.obj)
+0000A9FE Far.exe!std::basic_string<wchar_t,struct std::char_traits<wchar_t>,class std::allocator<wchar_t> >::reserve(unsigned __int64) __ptr64+0xFE (cddrv.obj)
+0017E406 Far.exe!FileList::ConvertName(class std::basic_string_view<wchar_t,struct std::char_traits<wchar_t> >,class std::basic_string<wchar_t,struct std::char_traits<wchar_t>,class std::allocator<wchar_t> > & __ptr64,unsigned __int64,unsigned __int64,int,unsigned long)const __ptr64+0x66 (filelist.obj)
+0017FDA0 Far.exe!FileList::ShowList(int,int) __ptr64+0xCF0 (filelist.obj)
+0017C32E Far.exe!FileList::ShowFileList(bool) __ptr64+0x13EE (filelist.obj)
```

### Analysis
[...]

PR #1033, opened 2025-10-19
Crash if size of name column is less than size of mark character

rohitab:
## Summary
This PR fixes a crash that occurs if the size of the name column is less than the size of the mark character used for file highlighting.

 
## References
https://github.com/FarGroup/FarManager/issues/1032

## Checklist
- [x] I have discussed this with project maintainers:  https://github.com/FarGroup/FarManager/issues/1032<br/>

## Details
The original code first checks if `Width>1`, but then reduces the width by the length of the mark character `Width -= static_cast<int>(MarkLength)`, which causes `Width` to become negative if the mark character is larger than the column width. This negative width is then used to reserve memory, causing an exception and crash.

This PR adds a check for `Width >= MarkLength` to ensure that the width doesn't become negative. Once the width reduces below the size of the mark character, the mark character is no longer displayed.


alabuzhev:
Thank you

PR #1038, opened 2025-11-02
Fixed copy counter for skipped folders

yulian5:
## Checklist

## Details

When copying files/folders in Skip mode you probably noticed something wrong with counter of copied files/folders:

skipped_dirs_count

In this real example the progress dialog shows 248 files copied with total size 0 bytes. There no any 0 length files there and it actually counts the skipped folders. The problem is in file copy.cpp, line 1971:

`return ConvertNameToFull(Src) == strDestPath? COPY_SKIPPED : COPY_SUCCESS;`

That code compares full source path with full destination path (which are cannot be equal in any OS in the same format, I mean no any OS allows to copy file with the same name in the same folder), so it always returns COPY_SUCCESS. 

The fix requires some other code changes (as one error was compensated for by another's) but the real problem was to find how to fix time update in case when files/folders at the beginning of copy process or even all of them are skipped (changes in copy_progress.cpp).


alabuzhev:
The change looks reasonable, but please explain how to reproduce the issue.
copy.cpp is rather... convoluted, so I'd prefer to better understand what is going on and make sure we're not breaking something else here.

> The problem is in file copy.cpp, line 1971:
> 
> `return ConvertNameToFull(Src) == strDestPath? COPY_SKIPPED : COPY_SUCCESS;`

Interestingly, this line is very, very, very old:
https://github.com/FarGroup/FarManager/blob/3f560a1d5f5b25f0795f92fef4d927fe839e2223/far/copy.cpp#L618

yulian5:
@alabuzhev Yes, it's probably original error. I do not have Far2 code to check. And far2l has different handling of copy files.

The easiest way to reproduce is:

- first copy folder with bunch of subfolders to somewhere (a few hundreds files at least)
- then copy them second time in copy SKIP mode
- it will show growing number of copied files with total size 0 even you know the destination already has all of them.

If you have a few files/folders you may put breakpoint on line 1379, copy.cpp to keep progress dialog open at the end, just keep in mind update interval 1s so it may not show it with a small total number of files.

far/copy.cpp
```diff
@@ -92,6 +92,7 @@ enum COPY_CODES
 {
 	COPY_FILTERED,
 	COPY_SKIPPED,
+	COPY_SKIPPED_DIRECTORY,
```
alabuzhev, line 95: There's a [warning](https://github.com/FarGroup/FarManager/actions/runs/19009311550/job/54621893730#step:7:169):
> enumerator 'COPY_SKIPPED_DIRECTORY' in switch of enum 'COPY_CODES' is not handled

Supposedly the case block should be added there as well?

far/copy.cpp
```diff
@@ -1968,7 +1977,7 @@ COPY_CODES ShellCopy::ShellCopyOneFile(
 					if (SetAttr!=DestAttr)
 						ShellSetAttr(strDestPath,SetAttr);
 
-					return ConvertNameToFull(Src) == strDestPath? COPY_SKIPPED : COPY_SUCCESS;
+					return strDestPath.ends_with(Src)? COPY_SKIPPED_DIRECTORY : COPY_SUCCESS;
```
alabuzhev: Do we actually need to compare paths here?
strDestPath is formed above in the same function:

```C++
auto strDestPath = strDest;
...
    path::append(strDestPath, Src.substr(KeepPathPos));
```

I experimented a little but could not trigger a path where strDestPath would not end with Src.
Perhaps we can just return COPY_SKIPPED_DIRECTORY unconditionally?

far/copy_progress.cpp
```diff
@@ -314,13 +322,6 @@ void copy_progress::add_total_bytes(unsigned long long const Value)
 
 void copy_progress::SetNames(const string_view Src, const string_view Dst)
 {
-	if (m_ShowTime && !m_Files.Copied)
```
alabuzhev, line 317: Supposedly you moved it to the ctor because when we skip everything m_Files.Copied stays 0 and reset happens every time?

If so, then it makes sense in general.
Unfortunately, I don't quite remember why I put this block here in SetNames and not in the ctor.
But since I did and went to great lengths to initialize it lazily there probably was a reason for that.

It could be that if we do it in the ctor then the clock starts ticking immediately, which might be undesirable if there are confirmation dialogs in between or something.

Maybe we could keep the block here just in case, but adjust the condition like
```C++
if (m_ShowTime && !m_CopyStartTime.time_since_epoch().count())
```
?

far/copy.cpp
yulian5: That case is inside conditional block (line 1713):

`if ((Flags&FCOPY_MOVE) && SameDisk && !IsDirectory)`

It'll never reach that case for directory. Does it make sense to add extra processing to remove the warning?

far/copy.cpp
yulian5: Interesting. Let me check that.

far/copy.cpp
alabuzhev: No, that conditional block ends at line 1755 and there's `default`, so no warning there.

The warning is below that, around lines 1780-1790. 
It was reachable in my experiments.
And without skip(0) there the total files counter didn't go all the way down to 0.

yulian5: Yes, you're right. There is one more switch statement. I'll fix that. Thanks!

far/copy_progress.cpp
yulian5: OK. I agree with that. Initialization in the constructor is more understandable and intuitive, but perhaps less precise in terms of timing in the case of an x86, a VM, or simply a very busy system, or any other scenario with limited resources where creating the UI might take a long time.

> which might be undesirable if there are confirmation dialogs in between

Just confirmation or warning dialogs are not a problem by themselves since we have `WaitUserTime` for that. But creation the copy progress dialog itself (and the ones mentioned above) could be a problem, it seems to me. In mentioned corner cases.

I'd like to use more type-safe condition that doesn't mixed different types and which is a little more clear also in `SetNames()` function:

`if (m_ShowTime && m_CopyStartTime == std::chrono::steady_clock::time_point{})`

far/copy.cpp
yulian5: @alabuzhev More researching that peace of code actually reveals that it's not an error in original code but more like design to catch any complicated cases where the source and destination are the same folders. It's probably not implemented exactly right (and certainly not right in case of skipped folders).

I think in this place code should distinguish an attempt to copy folder to itself and the case of skipped folder. However you're right it most likely will return COPY_SKIPPED_DIRECTORY in both cases. But at least in case of recursive copy we can display existing error dialog "Cannot copy folder onto itself".

There is probably one more issue, now with `IsSameName()` function. It can detect when user copy the same folder to it symlink. But copying symlink folder to itself is not detected and code executes silently till that return statement.

yulian5: It doesn't make sense to modify `IsSameName()` function do display message only when symlink matches the destination folder. It make sense to do that for all symlinks if FCOPY_COPYSYMLINKCONTENTS is not set. It's a separate task.

Other than that COPY_SKIPPED_DIRECTORY works fine even if check for equal folders fails and code executes to this point. They are just skipped.

yulian5:
Hi @alabuzhev. It looks fine for me now. All concerns are fixed and changes are minimal.

alabuzhev:
Thank you

PR #1043, opened 2025-11-08
Fix tmppanel losing files that differ in letter case

### LINKED ISSUE #1041: Temporary panel loses files that differ in letter case
If a file list to be opened in the temporary panel contains:
1. multiple file names (>2) in one directory, and
2. some of that files differ in letter case from the names of actual files in the file system

then the panel won't show the files from the point 2.

Within this bug there is a minor discrepancy. To reproduce both, you can create 3 files with names: `a.txt`, `b.txt`, `c.txt`. Also create a file list `list1.temp` with contents:
```
a.txt
B.txt
c.txt
```
And the second file list `list2.temp` with contents:
```
a.txt
B.txt
```

If you open `list1.temp` you'll see only `a.txt` and `c.txt` items (bug). If you open `list2.temp` you'll see both `a.txt` and `B.txt` (and this is ok).

The fix would be pretty trivial:
```diff
diff --git a/plugins/tmppanel/TmpClass.cpp b/plugins/tmppanel/TmpClass.cpp
index 88baeb5a9..8153430fc 100644
--- a/plugins/tmppanel/TmpClass.cpp
+++ b/plugins/tmppanel/TmpClass.cpp
@@ -380,7 +380,7 @@ void TmpPanel::UpdateItems(const bool ShowOwners, const bool ShowLinks)
        m_LastOwnersRead = ShowOwners;
        m_LastLinksRead = ShowLinks;

-       std::unordered_map<string_view, PluginPanelItem*> SameFolderItems;
+       unordered_string_map_icase<PluginPanelItem*> SameFolderItems;

        auto NameDataIterator = m_Panel->StringData.begin();
        for (auto CurItem = m_Panel->Items.begin(), end = m_Panel->Items.end(); CurItem != end; ++CurItem, ++NameDataIterator)
@@ -411,7 +411,7 @@ void TmpPanel::UpdateItems(const bool ShowOwners, const bool ShowLinks)
                // несколько файлов из одного и того же каталога. При этом
                // FindFirstFile() делается один раз на каталог, а не отдельно для
                // каждого файла.
-               if (SameFolderItemsNumber > 2)
+               if (SameFolderItemsNumber >= 2)
                {
                        WIN32_FIND_DATA FindData;
                        const auto FindFile = Dir + L"*"sv;
```

The only problem here is the usage of the `unordered_string_map_icase`. It is located in `far/string_utils.hpp` and cannot be used directly in the plugin without creating additional (probably unwanted?) dependencies.

ssvine:
## Summary
This PR fixes #1041

 
## Checklist

### DIFF
--- a/plugins/tmppanel/TmpClass.cpp
+++ b/plugins/tmppanel/TmpClass.cpp
@@ -380,7 +380,7 @@ void TmpPanel::UpdateItems(const bool ShowOwners, const bool ShowLinks)
 	m_LastOwnersRead = ShowOwners;
 	m_LastLinksRead = ShowLinks;
 
-	std::unordered_map<string_view, PluginPanelItem*> SameFolderItems;
+	unordered_string_map_icase<PluginPanelItem*> SameFolderItems;
 
 	auto NameDataIterator = m_Panel->StringData.begin();
 	for (auto CurItem = m_Panel->Items.begin(), end = m_Panel->Items.end(); CurItem != end; ++CurItem, ++NameDataIterator)
@@ -411,7 +411,7 @@ void TmpPanel::UpdateItems(const bool ShowOwners, const bool ShowLinks)
 		// несколько файлов из одного и того же каталога. При этом
 		// FindFirstFile() делается один раз на каталог, а не отдельно для
 		// каждого файла.
-		if (SameFolderItemsNumber > 2)
+		if (SameFolderItemsNumber >= 2)
 		{
 			WIN32_FIND_DATA FindData;
 			const auto FindFile = Dir + L"*"sv;
--- a/plugins/tmppanel/changelog
+++ b/plugins/tmppanel/changelog
@@ -1,3 +1,7 @@
+ssvine 2025-11-08 12:56:45+03:00 - build 122
+
+1. gh-1041: Temporary panel loses files that differ in letter case
+
 drkns 2024-11-24 11:48:49+00:00 - build 121
 
 1. Remove quadratic complexity from panel updates.
--- a/plugins/tmppanel/headers.hpp
+++ b/plugins/tmppanel/headers.hpp
@@ -10,6 +10,7 @@
 #include <string>
 #include <string_view>
 #include <vector>
+#include <unordered_map>
 
 #include <cstdlib>
 #include <cwctype>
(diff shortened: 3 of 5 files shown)


plugins/tmppanel/changelog
```diff
@@ -1,3 +1,7 @@
+ssvine 2025-11-08 12:56:45+03:00
```
alabuzhev: Add build number here pls

plugins/tmppanel/TmpClass.cpp
```diff
@@ -411,7 +411,7 @@ void TmpPanel::UpdateItems(const bool ShowOwners, const bool ShowLinks)
 		// несколько файлов из одного и того же каталога. При этом
 		// FindFirstFile() делается один раз на каталог, а не отдельно для
 		// каждого файла.
-		if (SameFolderItemsNumber > 2)
+		if (SameFolderItemsNumber >= 2)
```
alabuzhev, line 414: Why >=?
It seems unrelated to case sensitiveness.

plugins/tmppanel/changelog
ssvine: done

plugins/tmppanel/TmpClass.cpp
ssvine: Yes, it is not related directly, it's just a quick fix to a minor issue - if we use an optimization, let's use it for all suitable number of files (so at least 2, not 3)

plugins/tmppanel/headers.hpp
```diff
@@ -28,5 +29,36 @@ inline namespace literals
...
+	bool operator()(const string_view Str1, const string_view Str2) const
+	{
+		if (Str1 == Str2)
+			return true;
+		if (Str1.size() != Str2.size())
+			return false;
+		return std::equal(Str1.begin(), Str1.end(), Str2.begin(),
+			[](wchar_t c1, wchar_t c2)
+			{
+				FSF.LUpperBuf(&c1, 1);
```
alabuzhev, line 56: Can we add
```C++
if (c1 == c2)
    return true;
```
before converting to uppercase?
Should save us some work for strings like "File" vs "file".

plugins/tmppanel/TmpClass.cpp
alabuzhev: Fair enough

plugins/tmppanel/headers.hpp
ssvine: Yes, definitely. Done

alabuzhev:
Thanks

Issue #1044, opened 2025-11-15
New feature: show line numbers in internal Editor


yulian5:
### Description of the new feature or improvement

I'm actually already have code for this feature on **F9 key** in internal Editor:

<img width="711" height="390" alt="Image" src="https://github.com/user-attachments/assets/56ab6364-6d41-4bb3-b1f4-1a48d255d757" />

I'm using COL_EDITORTEXT for line numbers. Not ideal.
Any comments, suggestions?

HamRusTal:
IMHO if such a feature gets added, it should not consume the `F9` key. The option to show/hide line numbers should go to the Editor settings dialog.

shmuz:
> IMHO if such a feature gets added, it should not consume the F9 key.

Why? The users of this feature would want to toggle it ON/OFF instantly without invoking an extra dialog.

HamRusTal:
There's only a limited number of functional keys, and consuming one for this feature is not wise IMHO.
And for those who need to toggle the option instantly, FAR has a proper mechanism of macros.

yulian5:
I agree with @shmuz, this feature requires separate key.

> for those who need to toggle the option instantly, FAR has a proper mechanism of macros.

@HamRusTal Does an instant macro need a key also?

I implemented this feature [PR #1051](https://github.com/FarGroup/FarManager/pull/1051) using Ctrl+F3, the same way I did in Linux version. For those who work on different OS it's very important have the same keys.

It's also in Editor Settings and has a separate color if needed.

<img width="594" height="419" alt="Image" src="https://github.com/user-attachments/assets/85ec99e7-d099-4812-8f8c-8f0acd69fe19" />

yulian5:
@HamRusTal:

> There's only a limited number of functional keys, and consuming one for this feature is not wise IMHO.

But what's about using F4 and F10 for Quit? As I know Far Manager editor is the only one using F4 for exit.

yulian5:
Here is the example with LineNum/HideNum label. @HamRusTal please note how many keys are not assigned only with Ctrl modifier.

https://github.com/user-attachments/assets/e7d31262-4c88-4ccb-937e-9ad20b8cb9e1

shmuz:
@yulian5
I noticed there are 3 states of the 3-rd label: "LineNu", "HideNu" and empty. What does an empty mean?

yulian5:
@shmuz
[...]

PR #1051, opened 2025-11-28
Editor: add line numbers

yulian5:
## Summary
Adding line numbers to internal editor with hot-key and separate color

 
## References
[#1044](https://github.com/FarGroup/FarManager/issues/1044)

## Checklist

## Details
This is a necessary feature for working with source files, I just  surprised that nobody added it to Far Manager.


far/editor.cpp
```diff
@@ -243,6 +243,15 @@ void Editor::ShowEditor()
 	Color = colors::PaletteColorToFarColor(COL_EDITORTEXT);
 	SelColor = colors::PaletteColorToFarColor(COL_EDITORSELECTEDTEXT);
 
+	// Calculate line number column width if needed
+	int LineNumWidth = 0;
```
alabuzhev: Can we please reword it as 

```C++
const auto LineNumWidth = EdOpt.ShowLineNumbers?
    std::max(3, static_cast<int>(std::log10(Lines.size())) + 1) + 1 :
    0;
```
- const
- +1 for spacing is more explicit and can be changed in one place if needed.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
 	auto LeftPos = m_it_CurLine->GetLeftPos();
 	const Edit::ShowInfo info{ LeftPos, CurPos };
 	auto Y = m_Where.top;
+	int LineNum = m_it_TopScreen.Number() + 1;
```
alabuzhev: Please remove this and use `CurPtr.Number() + 1` directly when needed.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
...
+	for (auto CurPtr = m_it_TopScreen; CurPtr != Lines.end() && Y <= m_Where.bottom; ++CurPtr, ++Y, ++LineNum)
 	{
 		CurPtr->SetEditBeyondEnd(true);
-		CurPtr->SetPosition({ m_Where.left, Y, XX2, Y });
+		CurPtr->SetPosition({ m_Where.left + LineNumWidth, Y, XX2, Y });
 		CurPtr->SetLeftPos(LeftPos);
 		CurPtr->SetTabCurPos(CurPos);
 		CurPtr->SetEditBeyondEnd(EdOpt.CursorBeyondEOL);
 
+		// Draw line numbers if enabled
```
alabuzhev: Please remove the comment, it's rather obvious what is happening here.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
...
+		if (EdOpt.ShowLineNumbers && LineNumWidth > 0)
```
alabuzhev: Just ` && LineNumWidth)` will do, it can't be negative.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
...
+		{
+			const auto LineNumStr = std::format(L"{:>{}}", LineNum, LineNumWidth - 1);
```
alabuzhev: We use fmtlib, so `far::format` please.
While it's mostly the same thing now and we might migrate to std one day, there is no point in pulling in two implementations until then.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
...
+			const auto LineNumColor = colors::PaletteColorToFarColor(COL_EDITORLINENUMBERS);
```
alabuzhev: Move it out of the loop please and use const ref, as Sonar suggests.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
...
+			SetScreen({ m_Where.left, Y, m_Where.left + LineNumWidth - 1, Y }, L' ', LineNumColor);
```
alabuzhev: The area will be almost entirely overwritten by `Text` anyway.
Please change the format string to `L"{:>{}} "` (+trailing space for spacing) and remove this.

far/editor.cpp
```diff
@@ -318,15 +327,27 @@ void Editor::ShowEditor()
...
+			SetColor(LineNumColor);
+			GotoXY(m_Where.left, Y);
+			Text(LineNumStr);
```
alabuzhev: `Text` has an overload that allow to do everything in one go, e.g.
```C++
Text({ m_Where.left, Y }, LineNumColor, LineNumStr);
```

far/editor.hpp
```diff
@@ -97,6 +97,8 @@ class Editor final: public SimpleScreenObject
 	void SetCharCodeBase(int NewMode) { EdOpt.CharCodeBase = NewMode % 3; }
 	int GetCharCodeBase() const { return EdOpt.CharCodeBase; }
 	void SetShowScrollBar(bool NewMode) { EdOpt.ShowScrollBar = NewMode; }
+	void SetShowLineNumbers(bool NewMode) { EdOpt.ShowLineNumbers = NewMode; }
```
alabuzhev: Not used.

far/farlang.templ.m4
```diff
@@ -1,4 +1,4 @@
-﻿#
+#
```
alabuzhev: Keep the BOM please.

far/farlang.templ.m4
```diff
@@ -4527,7 +4527,7 @@ MEditAutoDetectCodePage
 
 MEditShareWrite
 "Разрешить редактирование открытых для записи &файлов"
-"Allow editing files ope&ned for writing"
+"Allow editing files opened for &writing"
```
alabuzhev: Reassigning hotkeys is not really the point of this PR, is it?

It can also break someone's muscle memory or macros or whatever, are you sure you want to change that?

far/fileedit.hpp
```diff
@@ -120,6 +120,7 @@ class FileEditor final: public window,public EditorContainer
 	int GetId() const { return m_editor->EditorID; }
 	FileEditor* GetById(int ID) { return GetId()==ID?this:nullptr; }
 	void AutoDeleteColors() const { m_editor->AutoDeleteColors(); }
+	void UpdateKeyBar() { InitKeyBar(); }
```
alabuzhev: As Sonar says, this shadows `window::UpdateKeyBar`.
Please rename it or just make `InitKeyBar` public.

far/palette.cpp
```diff
@@ -209,6 +209,7 @@ Init[]
 	{L"WarnDialog.DefaultButton.Selected"sv,           F_BLACK|B_LIGHTGRAY,    }, // COL_WARNDIALOGSELECTEDDEFAULTBUTTON,
 	{L"WarnDialog.DefaultButton.Highlight"sv,          F_YELLOW|B_RED,         }, // COL_WARNDIALOGHIGHLIGHTDEFAULTBUTTON,
 	{L"WarnDialog.DefaultButton.Highlight.Selected"sv, F_YELLOW|B_LIGHTGRAY,   }, // COL_WARNDIALOGHIGHLIGHTSELECTEDDEFAULTBUTTON,
+	{L"Editor.LineNumbers"sv,                          F_YELLOW|B_BLUE,        }, // COL_EDITORLINENUMBERS,
```
alabuzhev: Yellow on Blue is a beautiful color combination, I absolutely adore it.

However, in the default color scheme of this particular project it usually implies some form of highlight: see panel selection, column titles, or even your other PR about mouse dragging.

Are you sure that line numbers, which is basically some additional information in some service area, need to be notionally _highlighted_?
Shouldn't it be the other way around - something more subtle than the editor text, e.g. light or dark gray?

alabuzhev:
> This is a necessary feature for working with source files, I just surprised that nobody added it to Far Manager.

Perhaps not so necessary, if it took 29 years?
Never have I ever thought "what a pity there's no such a feature".
Maybe I just don't work with source files enough.

Anyway, why not. Thanks for you PR.

alabuzhev:
There is a [failed assertion](https://github.com/FarGroup/FarManager/actions/runs/19753387459/job/56600570199?pr=1051#step:7:203) in macro tests:

https://github.com/FarGroup/FarManager/blob/7ce37df7ce3cdd3ec9d5c121202daf3f02032ead/plugins/luamacro/macrotest.lua#L1816

Please update accordingly.
Alternatively, we can just assert that it's > 0, I don't think that checking for a specific value that tends to grow from time to time has some intrinsic value (unless @shmuz has a different opinion).

shmuz:
> Alternatively, we can just assert that it's > 0

Agreed.
Or we can assert that it's > 140.

far/farlang.templ.m4
rohitab: Also, the `w` hotkey is already used by `"&Warn when opening read-only files"`

far/palette.cpp
yulian5: That's not a problem. This color was requested by far2l users as default. 
Every man to his own taste. I use pink/blue now. Originally I use normal text color as you can see on the first screenshot in [#1044](https://github.com/FarGroup/FarManager/issues/1044) but it's difficult to distinguish from the other text.
It's easy to change in Color settings.
I'll change default to light or dark gray, will test first which one will be better.

It will be interesting to ask each of you (@alabuzhev, @rohitab, @HamRusTal) in 2-3 months what colors you use.

far/farlang.templ.m4
yulian5: As @alabuzhev mentioned a couple of times about user custom colors we cannot be responsible for users custom preferences. I guess that applies to macros too. I do not see that Far Manager is 100% backward compatible and anywhere promised that. It good to be backward compatible if possible. If possible.

@rohitab, you probably missed another thread where I published Editor settings [screenshot](https://github.com/FarGroup/FarManager/issues/1044#issuecomment-3589502236)

The UI standard for hotkeys is to have maximum possible association with the item name, for example, for **File** it usually **F**, for **Open** is **O**, and so on. Having **I** or **E** hotkey for **File** will be confusing and misleading the user and will make using the program(s) difficult. 

Far Manager usually follows that but still a lot of hotkeys look like randomly assigned taking next available symbol.

For line numbers natural hotkey looks like **L** or **N**, so I have to reassign some weird hotkeys to make it work and to be compatible with Linux far2l. Some other languages just have duplicated hotkeys for Editor settings.

editor-settings-2

I think, some keys should be reassigned across the Far, but that is a separate topic. 

If you will have much better ideas about Line Numbers hotkey, let's use them.

far/editor.cpp
```diff
@@ -319,14 +324,26 @@ void Editor::ShowEditor()
 	const Edit::ShowInfo info{ LeftPos, CurPos };
 	auto Y = m_Where.top;
 
+	const auto LineNumColor = LineNumWidth > 0 ? colors::PaletteColorToFarColor(COL_EDITORLINENUMBERS) : FarColor{};
```
alabuzhev: No need to have a condition here.
Just `const auto& LineNumColor = colors::PaletteColorToFarColor(COL_EDITORLINENUMBERS);` please.

far/editor.cpp
```diff
@@ -319,14 +324,26 @@ void Editor::ShowEditor()
...
-		CurPtr->SetPosition({ m_Where.left, Y, XX2, Y });
+		CurPtr->SetPosition({ TextLeft, Y, XX2, Y });
 		CurPtr->SetLeftPos(LeftPos);
 		CurPtr->SetTabCurPos(CurPos);
 		CurPtr->SetEditBeyondEnd(EdOpt.CursorBeyondEOL);
 
+		if (LineNumWidth > 0)
+		{
+			const auto LineNumStr = far::format(L"{:>{}}"sv, CurPtr.Number() + 1, LineNumWidth - 1);
+			SetScreen({ LineNumLeft, Y, LineNumRight, Y }, L' ', LineNumColor);
```
alabuzhev: As mentioned earlier, this call would be redundant if you just extend the template above to `L"{:>{}} "`.

far/editor.cpp
```diff
@@ -243,6 +243,11 @@ void Editor::ShowEditor()
 	Color = colors::PaletteColorToFarColor(COL_EDITORTEXT);
 	SelColor = colors::PaletteColorToFarColor(COL_EDITORSELECTEDTEXT);
 
+	// Calculate line number column width if needed
+	// Minimum width of 3 for "1", "10", "100", etc. plus 1 for spacing
+	const auto LineNumWidth = EdOpt.ShowLineNumbers ?
+		(Lines.empty() ? 0 : std::max(3, static_cast<int>(std::log10(static_cast<double>(Lines.size()))) + 1) + 1) : 0;
```
alabuzhev: The check is redundant.
If Lines is empty we won't even get here.
See the condition above.

far/farlang.templ.m4
alabuzhev: What I'm saying is that the point of this PR is introducing a new feature.
The smaller the PR and the fewer other features it affects, the faster it will get merged.

You can, for example, add `[ ] Show line numbers` without any hotkey at all, avoiding any conflicts with other hotkeys and any need to touch or discuss them now, and then raise a follow-up PR with hotkeys rearrangements only.

Or you can touch this highly subjective topic here.
The changes you suggest might be totally reasonable - indeed, maximum possible association with the item name is a logical goal, and personally I don't mind the new assignments at all, but other people might have different opinions (and because of the bikeshedding they absolutely will), so we might spend unholy amounts of time in this rabbit hole.

far/editor.cpp
```diff
@@ -319,14 +324,25 @@ void Editor::ShowEditor()
 	const Edit::ShowInfo info{ LeftPos, CurPos };
 	auto Y = m_Where.top;
 
+	const auto LineNumColor = colors::PaletteColorToFarColor(COL_EDITORLINENUMBERS);
+	const auto LineNumLeft = m_Where.left;
+	const auto LineNumRight = m_Where.left + LineNumWidth - 1;
```
alabuzhev: LineNumRight is unused now, please remove.

far/editor.cpp
```diff
@@ -319,14 +324,25 @@ void Editor::ShowEditor()
 	const Edit::ShowInfo info{ LeftPos, CurPos };
 	auto Y = m_Where.top;
 
+	const auto LineNumColor = colors::PaletteColorToFarColor(COL_EDITORLINENUMBERS);
```
alabuzhev: Please address the warning.

alabuzhev:
Thanks, I've added a couple more comments, pls have a look and squash the commits when ready, then we can merge.

yulian5:
@alabuzhev OK. Done.

alabuzhev:
Thanks

alabuzhev:
@yulian5, are you mentioned in [CONTRIBUTORS](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTORS.md) already?
If not, should I add you as yulian5 or as some other name?

PR #1059, opened 2025-12-17
Preserve timestamps for existing destination directories

### LINKED ISSUE #460: Directories timestamps are not preserved on copy/move when they already exist
### Far Manager version

3.0.5888.0

### OS version

10

Here's a simple scenario.

Say you're copying/moving a directory tree to a new location where some directories already exist.

In this case the "Preserve All Timestamps" option will **not** work for the **already existing directories** at the destination whose timestamps must be overwritten with the timestamps of the directories you're moving/copying.

```
     \file1     \file4
    a\file2 -> a\file5
     \file3     \file6

(source)    -> destination
```
The source directory `A` timestamps: `TIME_LWRITE TIME_CREATION TIME_LACCESS TIME_CHANGE`
After copying or moving the destination `A` directory timestamps will become `NOW NOW NOW NOW` and they must become `TIME_LWRITE TIME_CREATION TIME_LACCESS TIME_CHANGE`.

Please fix.

birdie-github:
## Summary
When copying or moving directories with "Preserve all timestamps" enabled, timestamps are restored only for newly created destination directories.

If the destination directory already exists, its timestamps are currently left unchanged, even when it corresponds to a source directory being copied. This breaks the expected behavior of the option.

Queue existing destination directories for timestamp restoration in the same way as newly created ones, so their timestamps are restored after copy completion.

## References
Fixes #460


birdie-github:
On the new Sonar Cloud's issue: the additional nesting already exists and this change adds no new structural complexity and keeps the behavior local and explicit.

alabuzhev:
Looks good.
Pls update changelog and vbuild.m4 too, squash and we can merge.

PR #1074, opened 2026-02-09
Fix msdocs links for chm

johnd0e:
.text property is not available in quirks mode

For example see the link in CPM_NATIVE article.


alabuzhev:
Thanks

johnd0e:
Funny but the mentioned link though fixed but still finds [nothing](https://learn.microsoft.com/en-us/search/?terms=File%20and%20Directory%20Names).

The MS article itself is [there](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file), but I have no idea how to reference it properly. Their search seems broken.

PR #1075, opened 2026-02-09
hh_compiler: add usage and diagnostic messages

johnd0e:
Every time (if I haven't edited the help files for a long time), I run into the issue of CHM compilation failing.
And each time, it takes some time before I figure out the cause.
The reason is trivial: the Microsoft HTML Help Workshop must be installed.
However, `hh_compiler.exe` simply exits silently without any error messages.
Given that it doesn't output any messages upon launch, it's initially unclear whether it's working at all.

Therefore, I've added the following messages:

**Usage**:
```
Usage: D:\xxx\FarManager\enc\tools\hh_compiler\hh_compiler.exe <CodePage> <InputHHP>
```

<details><summary>Error encountered</summary>

```
[*] Target CodePage: 1
[*] Loading hha.dll...
[!] LoadLibraryA("hha.dll") failed. Error: 126

--------------------------------------------------------------------------------
RECOMMENDATION:
Microsoft HTML Help Workshop should be installed.
--------------------------------------------------------------------------------

```
</details> 

<details><summary>Normal flow</summary>

```
[*] Target CodePage: 1251
[*] Loading hha.dll...
[*] Patching function: GetACP
[*] Patching function: MultiByteToWideChar
[*] Patching function: WideCharToMultiByte
[*] Starting compilation of 'pluginsr.hhp'...
Microsoft HTML Help Compiler 4.74.8702

Compiling d:\xxx\FarManager\enc\build\chm\ru\FarEncyclopedia.ru.chm

...

[*] CompileHHP finished successfully.

```
</details>

### DIFF
--- a/enc/tools/hh_compiler/hh_compiler.cpp
+++ b/enc/tools/hh_compiler/hh_compiler.cpp
@@ -1,5 +1,6 @@
 ﻿#include <cstdio>
 #include <windows.h>
+#include <shlwapi.h>
 
 int CodePage;
 
@@ -45,7 +46,10 @@ static bool write_memory(void const** const To, const void* const From)
 {
 	MEMORY_BASIC_INFORMATION Info;
 	if (!VirtualQuery(To, &Info, sizeof(Info)))
+	{
+		fprintf(stderr, "[!] VirtualQuery failed for address %p. Error: %lu\n", To, GetLastError());
 		return false;
+	}
 
 	DWORD Protection;
 
@@ -66,18 +70,30 @@ static bool write_memory(void const** const To, const void* const From)
 	}
 
 	if (!VirtualProtect(Info.BaseAddress, Info.RegionSize, Protection, &Protection))
+	{
+		fprintf(stderr, "[!] VirtualProtect (Unlock) failed for address %p. Error: %lu\n", Info.BaseAddress, GetLastError());
 		return false;
+	}
 
 	*To = From;
 
-	return VirtualProtect(Info.BaseAddress, Info.RegionSize, Info.Protect, &Protection);
+	if (!VirtualProtect(Info.BaseAddress, Info.RegionSize, Info.Protect, &Protection))
+	{
+		fprintf(stderr, "[!] VirtualProtect (Restore) failed for address %p. Error: %lu\n", Info.BaseAddress, GetLastError());
+		return false;
+	}
+
+	return true;
 }
 
 static bool patch(HMODULE Module)
 {
 	const auto Headers = reinterpret_cast<PIMAGE_NT_HEADERS>(PBYTE(Module) + PIMAGE_DOS_HEADER(Module)->e_lfanew);
 	if (Headers->OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_IMPORT].Size <= 0)
+	{
+		fprintf(stderr, "[!] Import directory not found or empty in module %p.\n", Module);
 		return false;
+	}
 
 	bool AnyPatched{};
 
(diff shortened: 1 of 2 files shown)


johnd0e:
The next thing I usually try is, instead of **installing** that legacy piece of software into my system,
I just **copy** `hha.dll` into the `hh_compiler\` directory.
Of course, that is not enough: "HHC6003: Error: The file Itircl.dll has not been registered correctly."
Unfortunately, this message is not helpful at all, and there is even no solution **in** the first **Google** search results.
That's why I added special handling for this case:

```
--------------------------------------------------------------------------------
RECOMMENDATION:
Make sure that `itcc.dll` exists and run the following command
with Administrator privileges:
regsvr32 "D:\xxx\FarManager\enc\tools\hh_compiler\itcc.dll"

ERROR:
The required file was NOT found at the expected location:
D:\xxx\FarManager\enc\tools\hh_compiler\itcc.dll
--------------------------------------------------------------------------------
```

(Source: [http://mvps.org/htmlhelpcenter/itircl.htm](https://web.archive.org/web/20000620213107/http://mvps.org/htmlhelpcenter/itircl.htm))

Note that this message appears only when `hha.dll` is already present either in system. or near `hh_compiler.exe`.

But yes, in general, there is no need to install the full **HTML Workshop** package; it's enough to have just **two** files of its content.

alabuzhev:
> RECOMMENDATION:
> Microsoft HTML Help Workshop should be installed.

> RECOMMENDATION:
> Make sure that `itcc.dll` exists and run the following command
> with Administrator privileges:
> rundll32 "D:\xxx\FarManager\enc\tools\hh_compiler\itcc.dll",DllRegisterServer

So what is the 'recommended' way?
Install it normally or mess with random dlls?

johnd0e:
Officially, I don't recommend anything other than the official installation. And even in this case, problems may arise, and the solution that I present was originally focused on a fully installed Workshop. 
My dll files are also not so random, but were once installed, but I prefer to mess with them for personal reasons (although I suspect that I'm not the only one).

alabuzhev:
> I prefer to mess with them for personal reasons (although I suspect that I'm not the only one)

That's perfectly fine, I'm not judging.
My point is that you want to add informational messages that somewhat contradict each other: one recommends installing the workshop, another recommends manually registering enc\tools\hh_compiler\itcc.dll, which _normally_ does not exist at that particular location.
It for sure exists in your clone (because you put it there), but for obvious reasons we need something more generic, e.g. without mentioning any paths.

> rundll32 "D:\xxx\FarManager\enc\tools\hh_compiler\itcc.dll",DllRegisterServer

`regsvr32 path\itcc.dll`, no?

johnd0e:
> one recommends installing the workshop

This is exactly what is encountered by anyone who does not have the workshop installed.

> another recommends manually registering enc\tools\hh_compiler\itcc.dll

This applies only in case of one very specific error. And the path is determined dynamically.

> which normally does not exist at that particular location

The location is wherever the loaded hha.dll is located. This is normally in c:\windows\system32 (syswow64).

> e.g. without mentioning any paths.

The exact path should still be specified even if the DLL is located in the current directory, because the command will be run with elevated privileges.

As for exact command, than I will check it again tomorrow, thank you for pointing this out.

johnd0e:
> regsvr32 path\itcc.dll, no?

fixed

enc/tools/hh_compiler/hh_compiler.cpp
```diff
@@ -1,5 +1,7 @@
 ﻿#include <cstdio>
 #include <windows.h>
+#include <shlwapi.h>
+#pragma comment(lib, "shlwapi.lib")
```
alabuzhev: Move the lib [here](https://github.com/FarGroup/FarManager/blob/1a4340d7d218edd01cd5bd09b2cfe011711e0125/enc/tools/hh_compiler/build.cmd#L1) pls

enc/tools/hh_compiler/hh_compiler.cpp
```diff
@@ -112,32 +144,119 @@ static int WINAPI unknown(int)
 	return 1;
 }
 
+// Custom logging function for intercepting error messages
+static int __cdecl LogCallback(const char* format, ...)
```
alabuzhev: __cdecl is default, no need to be explicit.

enc/tools/hh_compiler/hh_compiler.cpp
johnd0e: Done

enc/tools/hh_compiler/hh_compiler.cpp
johnd0e: Fixed

alabuzhev:
Looks good, squash pls

johnd0e:
> Looks good, squash pls

Done

alabuzhev:
Thanks

johnd0e:
@alabuzhev 
Shouldn't [hh_compiler.exe](https://github.com/FarGroup/FarManager/blob/master/enc/tools/hh_compiler/hh_compiler.exe) be rebuilded?

alabuzhev:
> Shouldn't [hh_compiler.exe](https://github.com/FarGroup/FarManager/blob/master/enc/tools/hh_compiler/hh_compiler.exe) be rebuilded?

Done

PR #1083, opened 2026-03-11
Read XML attributes via GetAttribute during import

yegor-mialyk:
## Summary

Improve `configdb` XML import to read attributes through the shared GetAttribute() helper instead of accessing TinyXML attributes directly.

 
## References

https://github.com/FarGroup/FarManager/pull/1081

## Checklist

## Details

Some XML attributes are written through `SetAttribute()`, which can base64-encode values when they contain non-text-safe data. A number of import paths were still reading attributes directly via `Element.Attribute()`, which bypassed that decoding logic.

That made import handling inconsistent: exported values could be encoded on write but some readers would still consume the raw encoded text.


alabuzhev:
> A number of import paths were still reading attributes directly via `Element.Attribute()`, which bypassed that decoding logic.

I only covered the places that could have base64-encoded attributes under sane circumstances.
I.e. the ones when the attributes represent some *user-defined strings*.
In places where attributes represent other data, e.g. numbers, UUIDs etc. it does not make much sense anyway.

For example, `const auto Uuid = GetAttribute(i, "guid");`: the value here is an ASCII string by definition, it would never be base64-encoded by Far. An encoded value could appear here only if the user crafted it manually for whatever reason.

Having said that, using GetAttribute everywhere is more consistent and future-proof, so why not.

alabuzhev:
Thanks

yegor-mialyk:
@alabuzhev It was not easy to find all the exact places where a correction was needed, so it was easier to convert everything.

A couple of examples where it did not work:
`Options -> Interface Settings -> Far window title addons`
`Options -> Command line settings -> Set command line prompt format`

But who needs to insert control characters into those settings? That is more of a rhetorical question.

alabuzhev:
You're right, it's safer this way.
I've optimized a few things in 6658 so that we don't pay for extra strings etc.

Issue #1086, opened 2026-03-27
Recent releases of FAR (6644 and newer) crash with ConEmu


mihnita:
### Far Manager version

3.0.6644, x64

### OS version

Win 11 25H2 Build 26200.8037

### Other software

The latest version of ConEmu, 230724.
Which is pretty old, from 2023.
It works with 3.0.6600, fails with 3.0.6644 and newer

### Steps to reproduce

* Install FAR (3.0.6644 or newer)
* Install ConEmu (https://github.com/ConEmu/ConEmu/releases/tag/v23.07.24)
* Open a file in editor (F4)
* Try navigating down (Down Arrow, or Page Down, of End). Try several times, (long press, auto-repeat)

### Expected behavior

Go to a different place in editor

### Actual behavior

The console (with FAR in it) just "disappears"

If there is a message it is gone very fast.

Enabling the ConEmu debugging does not capture anything suspicious in the logs.

mihnita:
See PR https://github.com/FarGroup/FarManager/pull/1087

mihnita:
The crash started with commit [2acff8e8e](https://github.com/FarGroup/FarManager/commit/2acff8e8eeaaf5c732dc482e15a89953c9462f37), the addition of new fields in EditorInfo: WindowArea and ClientArea.

It happens in [far/fileedit.cpp, line 2650](https://github.com/mihnita/FarManager/blob/master/far/fileedit.cpp#L2650): 

```cpp
	const auto result = m_editor->EditorControl(Command, Param1, Param2);
	if (result&&ECTL_GETINFO==Command)
	{
		const auto Info=static_cast<EditorInfo*>(Param2);
		if (m_bAddSignature)
			Info->Options|=EOPT_BOM;
		if (Global->Opt->EdOpt.ShowTitleBar)
			Info->Options|=EOPT_SHOWTITLEBAR;
		if (Global->Opt->EdOpt.ShowKeyBar)
			Info->Options|=EOPT_SHOWKEYBAR;
		if (CheckStructSize(Info, &EditorInfo::WindowArea))
			Info->WindowArea = GetPosition().as<RECT>();  // !!! HERE
	}
```

I commented the WindowArea update in the `if CheckStructSize` and I added a `printf` in the `if CheckStructSize` where I printed `Info->StructSize`, `sizeof(*Info)`, `CheckStructSize(Info)` and `CheckStructSize(Info, &EditorInfo::WindowArea)`

[...]

PR #1087, opened 2026-03-27
Fix for #1086, crash in ConEmu

mihnita:
## Description

Solves bug #1086, crash with ConEmu

 
## Checklist

## Details

The crash started with commit [2acff8e8e](https://github.com/FarGroup/FarManager/commit/2acff8e8eeaaf5c732dc482e15a89953c9462f37), the addition of new fields in EditorInfo: WindowArea and ClientArea.

It looks like the ConEmu plugin (latest version 230724, from 2023) calls `ECTL_GETINFO` with un-initialized EditorInfo. Which means that `StructSize` is not initialized, and it is often (a lot) bigger than `sizeof(StructSize)`.

* https://github.com/ConEmu/ConEmu/blob/master/src/ConEmuPlugin/ConEmuPlugin995.cpp#L357
* https://github.com/ConEmu/ConEmu/blob/master/src/ConEmuPlugin/ConEmuPlugin1900.cpp#L398
* https://github.com/ConEmu/ConEmu/blob/master/src/ConEmuPlugin/ConEmuPlugin2800.cpp#L444
* https://github.com/ConEmu/ConEmu/blob/master/src/ConEmuPlugin/ConEmuPluginA.cpp#L295

I means that a simple `CheckStructSize(Info)` will sometimes pass, because it tests for `>=`.
And 2000091480960 is >= 184 :-) (that is a real value seen on a running FAR + ConEmu)

The structure is invalid, so it is dangerous to update `WindowArea` (FAR crashes).

My fix checks the exact size of the structure.
Testing for `==` is not 100% safe, there is a tiny chance to just happen.
But that's one chance out of 2^32 or 2^64 (max-value of `size_t`, architecture dependent).

### DIFF
--- a/far/fileedit.cpp
+++ b/far/fileedit.cpp
@@ -2646,7 +2646,16 @@ intptr_t FileEditor::EditorControl(int Command, intptr_t Param1, void *Param2)
 			Info->Options|=EOPT_SHOWTITLEBAR;
 		if (Global->Opt->EdOpt.ShowKeyBar)
 			Info->Options|=EOPT_SHOWKEYBAR;
-		if (CheckStructSize(Info, &EditorInfo::WindowArea))
+		/* It looks like the ConEmu plugin (latest version 230724, from 2023) calls `ECTL_GETINFO`
+		   with un-initialized EditorInfo. Which means that `StructSize` is not initialized,
+		   and it is often (a lot) bigger than `sizeof(StructSize)`.
+		   I means that a simple `CheckStructSize(Info)` will pass, because it tests for `>=`.
+		   And 2000091480960 is >= 184 :-) (that is a real value seen on a running FAR + ConEmu)
+		   The structure is invalid, so it is dangerous to update `WindowArea` (FAR crashes).
+		   Testing for `==` is not 100% safe, there is a tiny chance to just happen.
+		   But that's one chance out of 2^32 or 2^64 (max-value of `size_t`, architecture dependent).
+		 */
+		if (Info && Info->StructSize == sizeof(*Info) && CheckStructSize(Info, &EditorInfo::WindowArea))
 			Info->WindowArea = GetPosition().as<RECT>();
 	}
 	return result;


alabuzhev:
As others have already explained, this is purely a conemu issue.

Plugins **must** always set StructSize to sizeof(that struct).
It's the contract.
It tells Far "I have this many data fields into which you can write".
It allows us to introduce new data fields in the future.
Plugins, compiled with new headers, will report the larger size and would be able to use those fields.
Old plugins will keep reporting the old size, compiled into their binaries, and Far will not attempt to write to non-existing fields.

As you have already discovered, conemu [does not initialize StructSize](https://github.com/ConEmu/ConEmu/blob/c8cebb921e69d26fde43ba0a2af79115d1a07977/src/ConEmuPlugin/ConEmuPlugin2800.cpp#L444).
[It does not even check if the call was successful](https://github.com/ConEmu/ConEmu/blob/c8cebb921e69d26fde43ba0a2af79115d1a07977/src/ConEmuPlugin/ConEmuPlugin2800.cpp#L445) before using the data.
Such a design is sloppy, careless and basically begging for problems.
Since it does not initialize StructSize, StructSize contains a random set of bits.
Sometimes those bits resolve to zero, sometimes to 2000091480960 or any other random number.
If that number happens to be large enough, which is usually the case, Far treats is as a promise that extra fields do exist.
Because it's the contract.
Far writes into those fields, but they do not exist, because the promise is a lie, so writing there just trashes the stack.
Ironically, conemu is compiled with stack checks, so eventually it notices that the stack is broken and terminates the process itself lol.

The only thing that needs fixing here is conemu.
"Oh but it worked in the past" is not an excuse.
Also, your PR does not make sense: you have added `Info->StructSize == sizeof(*Info)`, but that effectively prevents us from extending EditorInfo in the future, because doing so would break existing plugins.

That's not how it works. That's not how any of this works.

MKadaner:
@alabuzhev, I have never looked at the `CheckStructSize` implementation. This time my curiosity was stirred enough. I'd say that the implementation is absolutely not what I expected. I presumed that when Far checks `StructSize`, it would know exact size of each historical version of the struct in question and would somehow communicate to the caller which version we've got here. Then, the API handler would use this information to activate the code path which knows how to deal with this particular version. Some discipline and template magic would be required, but these are implementation details.

Granted, keeping track of all historical versions of each PlugAPI structure would be rather expensive. However, I believe that this is the price a developer who publishes extensible, backward-compatible API should pay.

Not saying that anything should be fixed immediately, even less retroactively. Just something to consider.

alabuzhev:
> keeping track of all historical versions of each PlugAPI structure would be rather expensive

@MKadaner, unfortunately, it would be prohibitively expensive.
Because it would require a time machine.

Let's imagine we implemented your approach, track all the known sizes and refuse to work with unknown ones.

Year 2026. Structure S has size N.
The corresponding API handler accepts only N.

Year 2027. We added something and S's size is now M.
The corresponding API handler accepts N and M.

Someone writes a plugin and compiles it using the latest headers (2027).
Or maybe just recompiles some old sources with them, because why not.
The plugin, naturally, reports S's size as M.

Do you see the problem?
History goes both ways. 
The plugin will be rejected by all pre-2027 versions of Far.
Because the API handler in them has no idea what the heck M is, it is only aware of N.
Because they were compiled and released into the wild before M happened.

Even if the plugin does not use any new features, it will not work with older versions.
Unless we somehow go back in time and retroactively patch that API handler.

I doubt we will be able to afford enough antimatter / mana / whatever it takes any time soon, if ever.

MKadaner:
> Do you see the problem?

Not sure I do. As far as I know, this approach works with Win32 API. Yes, MS publishes and maintains multiple versions of some data structures side-by-side in the headers. At ABI level, they are fed to the same API functions, which then decided how to behave based on the size field (effectively tag) at the beginning.

If a client (plugin) wants to work with older Far versions, it should use the old structure. New functionality will not be available. If they want the new features, they should compile with the new definition and become incompatible with old Far versions which do not support the new functionality anyway. If they want both (best effort on old, new features on new), they should ask Far version and use corresponding definition.

I do not see any problems besides certain maintainability overhead on _both sides_.

alabuzhev:
Ah, you mean maintaining all the versions not only within Far, but in public API as well?
Yes, this is possible of course, but it's way more work and way more mess and it's too late anyway.

Plugins authors are expected to just set the struct size properly, it's not that hard.

Issue #1000, opened 2025-09-13
Each item of Find All menu contains coordinates of the found text...


MKadaner:
... which is absolutely unnecessary.

With the existing implementation of "fixed columns" in menu, the content of the fixed columns is easy and much more efficient to supply on demand.

MKadaner:
### A remarkable milestone: our 1000th issue!

This isn't just a number—it's a tribute to the quiet brilliance and tireless dedication of our maintainers. You turn complexity into confidence and keep the foundation strong. This moment belongs to you! 🍺

---

Sorry, could not help noticing. Coauthored with my wife and GPT-5.

MKadaner:
Implemented in 6683, #1098.

PR #1098, opened 2026-04-27
gh-1000: Do not materialize text coordinates in Find All menu items.

MKadaner:
## Summary
Find All menu displays the line and position ("coordinates") of the found pattern in the fixed columns at the left side of the menu. The coordinates were formatted into the menu item text string, and to display them the corresponding part of the item text was printed in the fixed column.

After this change, Find All menu items' text is just the content of the corresponding editor line. To obtain the strings to display in the fixed columns, menu invokes the `fixed_column_provider` function which should be registered by the menu owner.

Macros should use `Menu.GetItemExtendedData` instead of extracting line and position of the found pattern from the item's text.

## References
- gh-1000

## Checklist
- [X] I have discussed this with project maintainers: gh-1000<br/>


MKadaner:
@alabuzhev, when you have time, could you please review? It's not small but reasonably straightforward.

alabuzhev:
Thanks

MKadaner:
Thank you!

PR #1110, opened 2026-06-08
Simplified menu item annotations.

MKadaner:
## Summary
Even though we can render multiple annotations on each `VMenu` item, we never set more than one. So why pretending? Let's be honest and say there can be no more than one. This assumption allows to simplify the code and opens opportunity for further improvements (stay tuned).

## Checklist
- [ ] I have discussed this with project maintainers: Not yet.<br/>

## Details
- Replaced the list of annotations in `menu_item_ex` with the single optional annotation.
- `markup_highlight` (former `markup_slice_boundaries`) returns `std::array<string_view, 3>`.
- Changed the types of `FixedColumnsArea` and `TextArea` of `menu_layout` from `small_segment` to `segment`.
- Everything else is mere consequences.


far/vmenu.cpp
```diff
@@ -3692,50 +3651,58 @@ TEST_CASE("find.nearest.selectable.item")
 	}
 }
 
-TEST_CASE("markup.slice.boundaries")
+TEST_CASE("markup.highlight")
 {
+	using namespace std::string_view_literals;
```
alabuzhev: No need to, it's used project-wide.

far/vmenu.cpp
```diff
@@ -3692,50 +3651,58 @@ TEST_CASE("find.nearest.selectable.item")
...
 
 	for (const auto& TestDataPoint : TestDataPoints)
 	{
-		Markup.clear();
-		markup_slice_boundaries(TestDataPoint.Segment, TestDataPoint.Slices, Markup);
-		REQUIRE(std::ranges::equal(TestDataPoint.Markup, Markup));
+		const auto Markup{ markup_highlight(TestDataPoint.Text, TestDataPoint.Highlight) };
+		REQUIRE(Markup.size() == TestDataPoint.Markup.size());
+
+		for (const auto [Actual, Expected] : std::views::zip(Markup, TestDataPoint.Markup))
```
alabuzhev: views::zip is [available](https://en.cppreference.com/cpp/compiler_support#C++23_library_features) in gcc 13, while we [still promise](https://github.com/FarGroup/FarManager/blob/b17fe77e42cee57d76455ee62140295fcdf05f72/far/headers.hpp#L41) to support 12.

I don't particularly mind bumping the requirements, but better separately / not in this PR.
[We have zip at home](https://github.com/FarGroup/FarManager/blob/b17fe77e42cee57d76455ee62140295fcdf05f72/far/common/view/zip.hpp#L245), could you pls try it instead? I'll replace it with std globally later.

alabuzhev (approved):

far/vmenu.cpp
MKadaner: Sure! The habit ingrained at work. Different code styles.

far/vmenu.cpp
MKadaner: Of course, it works. Thank you for the reminder.

alabuzhev:
Thank you

MKadaner:
Thank you!

PR #1114, opened 2026-07-04
Made `menu_item::Name` private and use cached `VisualLength` whenever possible.

MKadaner:
## Summary
With support of fullwidth characters, calculating the visual length of menu item name may be expensive.

Let's cache the visual length once it is calculated. The new field does not even increase `sizeof(menu_item)`.

## Checklist

## Details
1. Made `menu_item::Name` private.
2. Added accessors.
3. Added necessary constructors.
4. Replaced direct use of `menu_item::Name` with accessors calls.


far/vmenu.hpp
```diff
@@ -88,9 +88,31 @@ enum VMENU_FLAGS
...
+	menu_item(T&& Name, LISTITEMFLAGS Flags, DWORD AccelKey = {})
+		: Name{ std::forward<T>(Name) }, Flags{ Flags }, AccelKey{ AccelKey }
+	{}
+
+	const string& GetName() const noexcept
+	{
+		return Name;
+	}
+
+	string& MutableName() noexcept
```
alabuzhev: Most of the usages of this follow the `MutableName() = NewValue;` pattern.
Do we *really* need it?
Maybe something more conventional like `SetName(NewValue)` will do?

It's too easy to abuse it now by holding a mutable reference and modifying it randomly later.

far/vmenu.hpp
```diff
@@ -88,9 +88,31 @@ enum VMENU_FLAGS
...
+public:
 	LISTITEMFLAGS Flags{};
 	DWORD AccelKey{};
+	mutable int VisualLength{ InvalidVisualLength };
+	static constexpr auto InvalidVisualLength{ std::numeric_limits<decltype(VisualLength)>::min() };
+
+	menu_item() = default;
+
+	template<typename T> requires std::constructible_from<string, T>
+	menu_item(T&& Name, LISTITEMFLAGS Flags, DWORD AccelKey = {})
```
alabuzhev: Do we need a template here?
We're going to construct a `string` anyway, so might as well just accept it by value and move it into `Name`.

far/vmenu.hpp
MKadaner: Of course, `MutableName` is not exactly a good idea. I started from `SetName` but hit a couple of awkward places:

```C++
append(ListItem.MutableName(), L" \""sv, strFriendlyName, L"\""sv);
```

and especially

```C++
NewItem.MutableName().erase(NewItem.GetName().cbegin(), NewBegin);
```

If you are OK with the use of `SetName` here, I will sure like it.

far/vmenu.hpp
MKadaner: Leftover from when I experimented with overloaded constructors to reduce code churn at call sites.

P.S. C++ Core Guidelines suggest [passing “will-move-from” parameters by rvalue reference](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f18-for-will-move-from-parameters-pass-by-x-and-stdmove-the-parameter) (also, [overview of parameter passing techniques](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f15-prefer-simple-and-conventional-ways-of-passing-information)).

This topic is somewhat controversial around me. What do you think?

far/vmenu.hpp
alabuzhev: > `append(ListItem.MutableName(), L" \""sv, strFriendlyName, L"\""sv);`

`ListItem.SetName(concat(ListItem.GetName(), ...))`

> `NewItem.MutableName().erase(NewItem.GetName().cbegin(), NewBegin);`

`NewItem.SetName(NewItem.GetName().substr(NewBegin - NewItem.GetName().cbegin()))`

Doesn't look too bad.

far/vmenu.hpp
alabuzhev: > X&& binds to rvalues, which requires an explicit std::move at the call site if passing an lvalue

It is somewhat controversial indeed. What if we don't want to / can't move it?

MKadaner: Then we would have to explicitly instantiate the object, like `string{ MyUnmovableVar }`. So, everything is very explicit at the call site, which from my point of view is an advantage. May help avoiding perf issues when we accidentally make a copy. I've seen it couple of times.

far/vmenu.hpp
MKadaner: I meant

```C++
ListItem.SetName(far::format(L"{} \"{}\""sv, ListItem.GetName(), strFriendlyName));
```

Which is not really bad, considering.

MKadaner: Fixed them your way, though
- I think that `far::format` is more expressive (though probably more expensive), and it is already used to compose item name in few other places.
- I tried to avoid iterator arithmetic; it almost no better than pointer arithmetic.

alabuzhev:
Thank you

MKadaner:
Thank you!

PR #1079, opened 2026-03-01
Fixed drawing empty menu items.

MKadaner:
## Summary
The issue reported on Far bugtracker. Fallout of 6651.

## References
- https://bugs.farmanager.com/view.php?id=4095
- #1062

## Checklist
- [X] I have followed the [contributing guidelines](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTING.md).
- [X] I have discussed this with project maintainers: #1062<br/>
If not checked, I accept that this work might be rejected in favor of a different great big ineffable plan.<br/>

PR #1117, opened 2026-07-12
Clarified horizontal scrolling of empty menu items.

MKadaner:
## Summary
Now empty menu items participate in horizontal scrolling.

## References
- 6651 (#1062)
- 6652 (#1079)

## Checklist

## Details
In 6651 (#1062), horizontal scrolling of empty menu items was suppressed because of crashes and triggered assertions. The fix in 6652 (#1079) was a stopgap only. Empty menu items were still permanently attached to the left edge of the viewport.

While at the first glance scrolling empty menu items horizontally seems to be a no-op, it is meaningful in some scenarios. Consider for example a menu containing empty items in its initial state when all empty items logically positioned at the left edge of the viewport. If we scroll the items to the left, beyond the left edge, the left horizontal scroll indicator should be drawn even with empty items. However, it was not the case.

There was also a subtle issue when a forever fixed empty item would allow scrolling all other items indefinitely to the left or right. This issue could be observed if the user scrolls all menu items far to the right (for example) and then scrolls once to the left. The expectation is that the left character of at least one item should become visible immediately, no matter how many times the move-items-to-the-right command was executed.

This PR fixes all these issues by allowing empty menu items to natuarally participate in horizontal scrolling.


alabuzhev:
Thank you

MKadaner:
Thank you!

PR #1121, opened 2026-08-03
Clarified `DecorateItemsWithHotkeys`.

MKadaner:
## Summary
Now getting `DecorateItemsWithHotkeys` right... Hopefully.

The original implementation would misalign accelerator key of the item `"&1&2&3"`.

## References
N/A

## Checklist

### Fullwidth-aware `DecorateItemsWithHotkeys`
`DecorateItemsWithHotkeys` (former `AddHotkeys`) tried to compensate the visual string length accounted for by `fit_to_left` by comparing the results of `HiStrlen` and `visual_string_length`. If these functions produce different results, then there is an ampersand in the item name which will be removed during rendering, so we compensated for it adding an extra space before the accelerator string. Unfortunately, escaping ampersand removes all (unescaped) ampersands, so the string may occupy even less screen cells than the algorithm assumed.

The fix is simple: let's assume that `HiStrlen` returns honest visual length of the properly escaped string. Now, we need to append as many spaces as required to pad the visual length up to the maximum visual length.

### Refactoring
1. Renamed `menu_item` to `menu_item_data` and decoupled it from `menu_item_ex`.
   - The similarity between these classes is incidental: `menu_item_data` is a struct holding data required to construct menu items, while `menu_item_ex` is the actual non-trivial class representing menu items internally (well, mostly internally).
   - Having cached `VisualLength` in `menu_item_ex` enables future optimizations.
   - Removed `GetName` / `SetName` from `menu_item_data`. It is not needed there. I apologize for code churn.
2. Regularized `SetCheck`, `SetCustomCheck`, `SetSelect`, `SetDisable`, and `SetGrayed` helpers.
   - Now they are free functions, use a common helper, and serve both `menu_item_data` and `menu_item_ex`.


far/vmenu.hpp
```diff
@@ -86,22 +86,48 @@ enum VMENU_FLAGS
...
+	template<typename type>
+	concept menu_item_like = std::same_as<decltype(std::declval<type>().Flags), LISTITEMFLAGS>;
+
+	void set_flag(menu_item_like auto& Item, const bool Value, const LISTITEMFLAGS Flag) noexcept
+	{
+		if (Value) Item.Flags |= Flag; else Item.Flags &= ~Flag;
+	}
+}
+
+void set_select(detail::menu_item_like auto& Item, const bool Value) noexcept { detail::set_flag(Item, Value, LIF_SELECTED); }
```
alabuzhev: Why free functions?
Members usually have way better discoverability ("just press `.`"), and names like set_select, set_check etc. don't look like something menu-specific.

MKadaner: This way it was easier to define them for both, now unrelated, types. I wish we would have an equivalent of C# extension methods.

I am open to your suggestions, up to duplicating the whole bunch (while using the helpers). Moving `Flags` to a base class will break the order of definitions, so `menu_item_data` will require constructor to reorder the elements... It may work. Please let me know.

Meanwhile, I'll play with [explicit object member functions](https://en.cppreference.com/cpp/language/member_functions#Explicit_object_member_functions). I've never touched them because the feature seemed of rather limited use to me, but if combined with inheritance... Let me try.

alabuzhev: CRTP maybe?

MKadaner: I went with explicit object member functions. They feel more natural than CRTP, though conceptually similar. To be honest, I wanted to use this new to me feature, so that I remember how it works. Now `menu_item_data` requires constructors, but probably not a big deal.

On top of the functions manipulating flags, I renamed `GetName` / `SetName` to match the style.

Please see if you like it all this way.

Besides fixing `DecorateItemsWithHotkeys`, the refactoring is trivial though massive. If you want, I can split the change into several commits to make it easier to review and track later.

alabuzhev: Explicit object member functions feel a bit more wordy and would require bumping minimum compiler versions for gcc and clang (which is not a big deal though).

Yes, if possible, pls split to at least two (actual changes in the title in one and renamings in another).

MKadaner: Does bumping compiler versions require CI pipeline adjustments? If so, may I ask you to bump (at least) to GCC 14 and Clang 19, and probably MSVC to 19.43 if I read [the table](https://en.cppreference.com/cpp/compiler_support#cpp23:~:text=Explicit%20object%20member%20functions) correctly. If it is only adjusting [the requirements](https://github.com/FarGroup/FarManager/blob/91beb2d1842059333a074a0be6498c8cc3b6e190/far/headers.hpp#L40-L42), I will be happy to do it myself.

Hm-mm... I also noticed this [guard](https://github.com/FarGroup/FarManager/blob/91beb2d1842059333a074a0be6498c8cc3b6e190/far/common/shims_pre.hpp#L64-L65). Should it be fixed somehow?

I'll split the commit. If I finish before compiler version update, I'll replace explicit object member functions with CRTP.

---

P.S. I guess CRTP will be about the same size.

alabuzhev: > Does bumping compiler versions require CI pipeline adjustments?

No, CI uses lastest-ish, no need to update anything there.

> MSVC to 19.43

No need to. The table says it works in 19.32, except for modules support, and we won't care about modules in the foreseeable future.

But please update headers.hpp, for those who compile manually:
https://github.com/FarGroup/FarManager/blob/91beb2d1842059333a074a0be6498c8cc3b6e190/far/headers.hpp#L41-L42

> I'll replace explicit object member functions with CRTP

I'm not insisting on CRTP, feel free to keep explicit object member functions (this would also give me an excuse to modernize other parts of the codebase 😄)

> I guess CRTP will be about the same size

<table>
<tr>
<td>Explicit object member functions</td>
<td>CRTP</td>
</tr>
<tr>
<td>

```
namespace detail
{
	template<typename type>
	concept menu_item_like = std::same_as<decltype(std::declval<type>().Flags), LISTITEMFLAGS>;

	struct menu_item_base
	{
		void set_flag(this menu_item_like auto& Item, const bool Value, const LISTITEMFLAGS Flag) noexcept
		{
			if (Value) Item.Flags |= Flag; else Item.Flags &= ~Flag;
		}

		void set_select(this menu_item_like auto& Item, const bool Value) noexcept { Item.set_flag(Value, LIF_SELECTED); }
		void set_disable(this menu_item_like auto& Item, const bool Value) noexcept { Item.set_flag(Value, LIF_DISABLE); }
		void set_grayed(this menu_item_like auto& Item, const bool Value) noexcept { Item.set_flag(Value, LIF_GRAYED); }
		void set_check(this menu_item_like auto& Item, const bool Value) noexcept { Item.Flags &= ~0xFFFF; Item.set_flag(Value, LIF_CHECKED); }
		void set_check(this menu_item_like auto& Item, const wchar_t Char) noexcept { Item.Flags &= ~0xFFFF; Item.Flags |= LIF_CHECKED | Char; }
	};
}
```

</td>
<td>

```
namespace detail
{
	template<typename menu_item_type>
	struct menu_item_functions
	{
	private:
		void set_flag(const bool Value, const LISTITEMFLAGS Flag) noexcept { flags::change(static_cast<menu_item_type*>(this)->Flags, Flag, Value); }

	public:
		void set_select(const bool Value) noexcept { set_flag(Value, LIF_SELECTED); }
		void set_disable(const bool Value) noexcept { set_flag(Value, LIF_DISABLE); }
		void set_grayed(const bool Value) noexcept { set_flag(Value, LIF_GRAYED); }
		void set_check(const bool Value) noexcept { set_flag(false, 0xFFFF); set_flag(Value, LIF_CHECKED); }
		void set_check(const wchar_t Char) noexcept { set_flag(false, 0xFFFF); set_flag(true, LIF_CHECKED | Char); }
	};
}
```

</td>
</tr>
</table>

MKadaner: Done! The "Refactoring" commit is still large, but it was tricky to further split it.

alabuzhev:
Thank you

MKadaner:
Thank you!

PR #1131, opened 2026-09-21
configdb: fix newline rendering in database error messages

Zeroes1:
Summary
Fix multiline error messages in database fallback dialog to render properly.

Problem:
sshot-3

References
None.

Checklist
I have followed the [contributing guidelines](https://github.com/FarGroup/FarManager/blob/master/CONTRIBUTING.md).
Details
When a SQLite database fails to open and Far falls back to in-memory mode, the error dialog (ShowProblems) renders \n characters as visible glyphs instead of line breaks. This is because Message() displays each entry in m_Problems as a separate DI_TEXT element, and bare LF (0x0A) is not treated as a newline by the dialog renderer.

The fix splits embedded newlines in Report() messages so each line becomes a separate m_Problems entry, producing correct line breaks in the dialog.

After fix:
image


alabuzhev:
Thanks

Issue #986, opened 2025-08-07
Find dialog + External editor


qaqa339:
### Far Manager version

3.0.6524

### OS version

w11

### Steps to reproduce

Alt+F4 (run external editor) in Find File dialog

### Actual behavior

After returning to Far, dialog Find File disappears (arrow keys partially redisplay it - so dialog remains active, but hidden)
don't know which version start this issue (toooo many releases) but 6363 works OK

shmuz:
> don't know which version start this issue

Far 3.0.6482 is good
Far 3.0.6483 is bad

alabuzhev:
6525

PR #1132, opened 2026-09-23
findfile: redraw Find dialog after running an association

Zeroes1:
close https://bugs.farmanager.com/view.php?id=4116

F3 on a search result with a user association (e.g. notepad !.!) returned to the dialog without restoring it: ExecString refreshes the dialog first and then ShellUpdatePanels paints the panels over it, so the dialog stays invisible until something redraws it (e.g. mouse wheel over its area).

ProcessLocalFileTypes() returns true in this case and the whole refresh block in FindFiles::OpenFile() was skipped.

Force Dlg->Show() in that branch, same as gh-986 did for the external viewer/editor path.


alabuzhev:
Thanks.

(If you use LLMs, pls at least clean all that "one commit per line" nonsense after them)

Zeroes1:
I'll try to do better :)

